PR #331, opened 2017-10-08
Run button on code editor only

Effer:
Run button added in code editor buttons only.

image


AlRado:
It seems to me here a typo - "RUN [ctrl + f]", should be "RUN [ctrl + r]"

Effer:
Changes reviewed, thanks to point that out and fixed typos.

PR #335, opened 2017-10-13
Added the option to build using cmake.

ysblokje:
It compiles lua,lpeg,zlib,giflib and SDL2  from source.
Also it adds a compatibility header for dirent (and an empty one for unistd.h)

Please ignore the changes in the include statements. I did a manual merge and it does not change any of the sourcefiles. no idea how I can get github to change it mind.


ysblokje:
closing this , I don't trust github to do the right thing here.

PR #336, opened 2017-10-14
Cmakeify: Added the option to build using cmake take 2 

### LINKED ISSUE #320: Maybe move to CMake or something similar?
I think that we shall start moving to CMake or other automated build system, instead of hardcoding everything in the makefile. Because it is REALLY hard to make snap work with current makefile! Also, 
using provided libraries is a complete mess!

ysblokje:
It compiles lua,lpeg,zlib,giflib and SDL2 from source.
Also it adds a compatibility header for dirent (and an empty one for unistd.h)

It also changes a few inline functions to static inlines, this way the manual -O3 can be removed.

This should fix #320 at least for windows and linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,347 @@
+cmake_minimum_required(VERSION 3.6)
+project(TIC-80)
+set(CMAKE_C_STANDARD 99)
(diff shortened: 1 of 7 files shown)


ysblokje:
Closing, merging with aussieboke's efforts before we do another .
Sorry.

Issue #299, opened 2017-09-18
Live reload of .lua file(s) in developer mode


fmaida:
When I open a specific .p8 file in Pico-8 and I'm editing that same .p8 file in Atom, I can modify the file and when I save it, Pico-8 acknowledges that the file has been modified and reloads it on the fly (without showing again the boot sequence / startup animation).

I use Visual Studio Code and the TIC-80 plugin to edit TIC-80 files; I know that it is possible to enter a sort of "developer mode" and edit externally a .lua file by launching TIC-80 from the command line with the instruction:

`tic cart.tic -code game.lua`

But when I modify and save back the `game.lua`, I have to manually close and reopen TIC-80 in order to see applied changes I've made to the file.

Would it be possible to have TIC-80 wait for changes on a file and then reload? 
Basically, I'd like to:

1) Launch TIC with the command `tic cart.tic -code game.lua`
2) Edit `game.lua` with Visual Studio Code or any other code editor
3) Save `game.lua`
4) See TIC-80 that immediately reload the cart and runs it with the modifications I've just made

That would be really handy and a huge time saver IMHO

nesbox:
will do, thanks

fmaida:
Thank you very much

matheuslessarodrigues:
Could this be of help? https://emcrisostomo.github.io/fswatch/

matheuslessarodrigues:
Oh, maybe we could get a code reload when regaining focus. That could be simpler and we'd use the SDL event SDL_WINDOWEVENT_FOCUS_GAINED. I was trying to do that yesterday but I'm yet to figure out how to reload code.

matheuslessarodrigues:
@nesbox you think [this](https://github.com/nesbox/TIC-80/compare/tic_0.47.0...matheuslessarodrigues:tic_0.47.0?expand=1) would be a good solution?

What I did was to add this windows focus event to `studio.c` that will try to trigger code reload:
```c
case SDL_WINDOWEVENT_FOCUS_GAINED:
	{
		studio.console.codeLiveReload.reload(&studio.console,studio.code.data);
		if(studio.code.update)
			studio.code.update(&studio.code);
	}
break;
```

The code loading from file code was extracted from `cmdInjectCode()` and put into `loadFileIntoBuffer()`. This way it is now possible to be reused in `tryReloadCode` (everything inside `console.c`:
```c
static void tryReloadCode(Console* console, char* codeBuffer)
{
[...]

PR #340, opened 2017-10-16
Tic 0.47.0

matheuslessarodrigues:
Beside things described in issue299 ([link](https://github.com/nesbox/TIC-80/issues/299)), I've moved the `memset(buffer, 0, TIC_CODE_SIZE);` to inside the `if(contents)`.

Also, in order to implement the "-watch" option, I decided to do it as one single command "-code-watch".

So in order to invoke TIC injecting external code, one would still use
```
tic -code code.lua
```
And if you also want to watch for file modifications on window focus (sdl event)
```
tic -code-watch code.lua
```
I found this to be simpler as I didn't have to change this bit of code in `console.c`
```c
if (argc == 2) cmdLoadCart(console, argv[1]);
else if (argc == 3)
{
	cmdInjectCode(console, argv[1], argv[2]);
	cmdInjectSprites(console, argv[1], argv[2]);
	cmdInjectMap(console, argv[1], argv[2]);
}
else if (argc > 3)
{
	cmdLoadCart(console, argv[1]);

	for (s32 i = 2; i < argc; i += 2)
	{
		cmdInjectCode(console, argv[i], argv[i + 1]);
		cmdInjectSprites(console, argv[i], argv[i + 1]);
		cmdInjectMap(console, argv[i], argv[i + 1]);
	}
}
```
I belive that once we need to code more complex command line options, we'd be better with some command line arg parsing lib.

Is it cool? :)


src/console.c
```diff
@@ -2317,7 +2317,8 @@ static void tryReloadCode(Console* console, char* codeBuffer)
 
 static void cmdInjectCode(Console* console, const char* param, const char* name)
 {
-	if(strcmp(param, "-code") == 0)
+	bool watch = strcmp(param, "-code-watch") == 0;
+	if(strcmp(param, "-code") == 0 || watch)
```
incinirate: Micro-optimization: format as `if(watch || strcmp(param, "-code") == 0)` to prevent doing 2 `strcmp`s when only 1 is needed.

matheuslessarodrigues: Fair enough!

nesbox:
Pls don't forget to update the Wiki https://github.com/nesbox/TIC-80/wiki#command-line-arguments
Thanks

nesbox:
@matheuslessarodrigues 
found a bug here #346
pls fix

matheuslessarodrigues:
noted. should fix latter today!

PR #368, opened 2017-10-30
Arch latest flags

josefnpat:
First change relates to #367. Ubuntu won't compile if I remove the `-Llib/linux` from `LINUX64_LIBS`. I didn't make that change. 

Not entirely sure, but may fix #367 - can't tell until I test on my archbox.

Second change relates to makefile rename. Not sure if AUR maintainer needs to be informed.
Further reading for reason: https://www.gnu.org/software/make/manual/make.html#Makefile-Names

PR #372, opened 2017-11-01
Renaming makefile as per gnu recommendations

josefnpat:
As per discussion on PR #368


josefnpat:
targeted wrong branch

PR #414, opened 2017-11-23
Clip the sprite before drawing.

jahodfra:
I optimized the fast path. It provides quite large speed up even when the sprites are not clipped.

Please take a look.


nesbox:
yes, works faster, I got 30000 points in _benchmark.tic_
but, unfortunately, it's not stable on all the compilers, in VS2015 all is ok while with MinGW it gets stuck
it's weird, also, I like your idea, but it's too early to do so cardinal optimisations, let's optimise `drawTile` for all other modes with your first version, it should work with the same speed with flipped & rotated sprites too
and then we will try to optimise with this version

jahodfra:
I used gcc. I can try to compile it under mingw in linux. Does it stuck continuously?

I though about first optimizing fast version and then spreading the speed up to other sprite orientations. It's quite easy to make a mistake and cause segfault.

Then I though about merging fast path with scale > 1. (By computing coordinates through division).
Wrong branch prediction should be on the same scale as division (I am not sure if that will be faster).
What is the most common size of scale I should optimize for?

And finally I wanted to spread optimization to other orientations (by creating macro and parametrizing computations of "i" variable).

nesbox:
Yes, it gets stuck continuously on MinGW, I'll try to investigate why...

nesbox:
hmm, reproduces only with `-O3`, with `-O1` and `-O2` all is ok

nesbox:
something strange with the stack, if I add `static u8 mapping[TIC_PALETTE_SIZE];` all works fine on `-O3`

jahodfra:
Please do not merge now, I noticed some strange effect in panda.tic.

nesbox:
What effect did you notice in panda? Because it seems, I built prerelease with this effect :(

jahodfra:
Please see commit 9a9f29ee9ed748c48da9013a7871c86df558dca6. (When the tile was clipped on the left or top side the pixels of the tile were mangled).

jahodfra:
I tried to merge paths for scale > 1, but that just hurt performance.
Currently I don't see more room for improvement. I also tried to remove macros, but the performance was 10-15% slower on my desktop.

Please check the code if you are comfortable with it. I am not sure if I should not squash the commits (so the project could have cleaner history).

src/tic.c
```diff
@@ -200,55 +201,70 @@ static void drawRectBorder(tic_machine* machine, s32 x, s32 y, s32 width, s32 he
...
 
+#define DRAW_TILE_BODY(index_expr) ({\
+	y += sy; \
+	for(s32 py=sy; py < ey; py++, y++) \
+	{ \
+		s32 xx = x + sx; \
+		for(s32 px=sx; px < ex; px++, xx++) \
+		{ \
+			u8 color = mapping[tic_tool_peek4(buffer, index_expr)]; \
+			if(color != 255) tic_tool_poke4(machine->memory.ram.vram.screen.data, y * TIC80_WIDTH + xx, color); \
```
jahodfra: This should be index_expr. I forgot to publish the fix.

jahodfra:
I will squash commits and open the new pull-request.

PR #423, opened 2017-11-27
Speedup drawTile 2.

jahodfra:
Clip the tile in all orientations.
Speed up blitting for scale=1.

### DIFF
--- a/src/tic.c
+++ b/src/tic.c
@@ -258,6 +258,9 @@ static void drawRectBorder(tic_machine* machine, s32 x, s32 y, s32 width, s32 he
 	} \
 	})
 
+#define REVERT(X) (TIC_SPRITESIZE - 1 - (X))
+#define INDEX_XY(X, Y) ((Y) * TIC_SPRITESIZE + (X))
+
 static void drawTile(tic_machine* machine, const tic_tile* buffer, s32 x, s32 y, u8* colors, s32 count, s32 scale, tic_flip flip, tic_rotate rotate)
 {
 	static u8 mapping[TIC_PALETTE_SIZE];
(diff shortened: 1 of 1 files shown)


nesbox:
I'll review and merge it after .50 release

jahodfra:
Thanks. I am looking into api_blit function now. For testing this pull request you can use https://tic.computer/play?cart=70

jahodfra:
I implemented changes from overlap. I haven't tested the speed on Rpi3. It seems to me that it maybe make sense to have two versions of drawTile to avoid indirection when setting pixels. (I avoid doing that for now).

src/tic.c
```diff
@@ -154,7 +155,7 @@ static inline u8 mapColor(tic_mem* tic, u8 color)
 
 static void setPixelDma(tic_mem* tic, s32 x, s32 y, u8 color)
 {
-	tic_tool_poke4(tic->ram.vram.screen.data, y * TIC80_WIDTH + x, mapColor(tic, color));
+	tic_tool_poke4(tic->ram.vram.screen.data, y * TIC80_WIDTH + x, color);
```
nesbox: pls restore `mapColor` function here, because of mapping affects not only sprites, it affects all the drawing api like `pix`, `line`, `rect` and etc...

src/tic.c
```diff
@@ -197,7 +198,7 @@ static void setPixel(tic_machine* machine, s32 x, s32 y, u8 color)
 {
 	if(x < machine->state.clip.l || y < machine->state.clip.t || x >= machine->state.clip.r || y >= machine->state.clip.b) return;
 
-	machine->state.setpix(&machine->memory, x, y, color);
+	machine->state.setpix(&machine->memory, x, y, mapColor(&machine->memory, color));
```
jahodfra: I moved the mapColor here.

src/tic.c
jahodfra: I moved the mapping function up into setPixel, and line, rect, etc. call the setPixel therefore it should be correct.

src/tic.c
nesbox: ahh, ok, I didn't notice this, sorry :)
merging...

nesbox:
PS. could you pls add yourself to the _Credits_ https://github.com/nesbox/TIC-80/wiki#credits :)
or give me account you want to mention
thanks

jahodfra:
I don't use Twitter yet, so I added link to my github account.

PR #425, opened 2017-11-29
Optimize api_blit inner loop.

jahodfra:
Function takes half of the clock time.

pi3
before 0.741 ms per function call
after 0.335

pi2
before 2.145
after 1.070

### DIFF
--- a/src/tic.c
+++ b/src/tic.c
@@ -1665,14 +1665,40 @@ static void api_blit(tic_mem* tic, u32* out, tic_scanline scanline)
 	for(s32 r = 0, y = tic->ram.vram.vars.offset.y; r < TIC80_HEIGHT; r++, y++, rowPtr += TIC80_FULLWIDTH)
 	{
 		memset4(rowPtr, pal[tic->ram.vram.vars.border], Left);
-		memset4(rowPtr + Left, pal[tic->ram.vram.vars.bg], TIC80_WIDTH);
-
-		u32* colPtr = rowPtr + Left;
 
 		if(y >= 0 && y < TIC80_HEIGHT)
-			for(s32 c = 0, x = tic->ram.vram.vars.offset.x, index = y * TIC80_WIDTH + x; c < TIC80_WIDTH; c++, colPtr++, x++, index++)
-				if(x >= 0 && x < TIC80_WIDTH)
-					*colPtr = pal[tic_tool_peek4(tic->ram.vram.screen.data, index)];
+		{
+			u32* colPtr = rowPtr + Left;
+			s32 offset = tic->ram.vram.vars.offset.x;
+			s32 count = TIC80_WIDTH;
+			s32 index = y * TIC80_WIDTH;
+			if (offset > 0) {
+				memset4(rowPtr + Left, pal[tic->ram.vram.vars.bg], offset);
+				count -= offset;
+				colPtr += offset;
+			} else {
+				count += offset;
+				index -= offset;
+			}
+			// copy the first pixel if the line is not alligned to bytes.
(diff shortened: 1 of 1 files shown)


nesbox:
@jahodfra I'm sorry, but I had to rewrite `api_blit` function again to implement Overlap Mode #429 
pls review changes in blit function fce433d and remake your optimizations if not difficult
thanks

jahodfra:
Hi Vadim,
have you removed my optimizations because you find hard to understand them
or you do not consider them important?

If that is second then it probably doesn't make sense for me to reintroduce
them. For the first case it makes sense to document them better.

It will take me some time to understand how overlap works.

so 2. 12. 2017 v 17:29 odesílatel Vadim Grigoruk <notifications@github.com>
napsal:

nesbox:
Hi, 
I understand what you did, you replaced `poke4` with two assigning operations
so, the same can be done here https://github.com/nesbox/TIC-80/blob/fce433d2720f60dd21dc2294e55f99992a72ff69/src/tic.c#L1725
I'll try to restore your optimisations later...

jahodfra:
I think it is maybe better to redo your changes in api_blit over my
implementation.

There are couple of parts in optimization:
 * if the screen is shifter outside of line the line is filled by
background color in one operation.
 * otherwise we fill left part with background color (if the image is
shifted to the right).
 * if the line is shifter about odd number of pixels we draw the first pixel
 * we set two pixels at once (there is no out of bound checking inside the
loop; 2n, and because we start on even pixel each two following pixels are
in the same byte).
 * we paint the last pixel (bottom part of the last byte).
 * if the image is shifter to the left we paint the remains of the line
with background color.

Therefore we avoid setting of the same memory twice and we don't have
conditional check if we copy even or odd pixel.

I can do the change, but maybe it would be better if you can do it and I
will review the change (if you want).

Btw, I checked your implementation of overlap drawing. It is something
which will slow down the TIC even if you would not use that. Effectively it
prevents inlining of setPixel, getPixel (the most common used functions in
the whole TIC). I am not sure how to effectively solve that (or if that is
problematic).

It is possible to implement the overlap like the drawing of the second
frame with special blitting (basically having blitting with transparent
color). That would be more costly, but it would not cost anything if the
user will not use it.

Also there is possibility to make setPixel, getPixel macro and having the
same drawing code generated twice or doing something similar with templates
in C++.
Which bring me to the question what are performance targets for the TIC? If
we don't want to support e.g. Rpi3 and target just desktops we are fine
with being even 5 times slower. OTOH if we target some platforms there
should be some tests to prove that new releases don't degrade performance.

What do you think?

so 2. 12. 2017 v 22:04 odesílatel Vadim Grigoruk <notifications@github.com>
napsal:

jahodfra:
I checked you code in api_blit and I think I understand why and how you did
that.

I can reintroduce my changes. I am just not sure if your infinite wrapping
is good solution. Maybe it should be configurable. I suppose that shifting
of screen was used e.g. for collision and explosions. The whole screen
could just shake. With the wrapping the screen would shake only if there
will be black pixels at its sides.

so 2. 12. 2017 v 22:51 odesílatel František Jahoda <
frantisek.jahoda@gmail.com> napsal:

jahodfra:
OTOH it is easy to fill part of the screen with black and get the previous
behavior.

Still shouldn't the screen wrap even in y-direction? That can create effect
as on older analog TVs.

ne 3. 12. 2017 v 23:08 odesílatel František Jahoda <
frantisek.jahoda@gmail.com> napsal:

nesbox:
I'm not sure about wrap in Y direction too, I just made it the same as X direction, don't know where it might be needed

Issue #197, opened 2017-06-18
Impossible to exit infinite while loop


pixelherodev:
I created an infinite while loop for debugging reasons (printed out a variable, then did nothing). Nothing I did would exit, and I was forced to click "force exit" on the Not Responding window.

LeonLaci:
I also had this problem.

darkhog:
Yeah, ESC should break out of program immediately. If my thinking is correct, the way it currently works is that breaks only after reaching the end of TIC() routine which obviously can't happen with an infinite loop in place.

DjPale:
Have also encountered this - I think this is the most serious thing I have seen with TIC-80 (0.30.dev). It can potentially lead to data loss if you have unsaved changes and accidentally create an infinite loop.

nesbox:
hmm, don't know how to handle infinite loops :(
need some researching...

DjPale:
I guess it would need to be handled differently based on which interpreter is used. I'm pretty sure Pico-8 handles this as I tested it some time ago (didn't re-verify at this point).

matheuslessarodrigues:
Maybe something like: if the TIC() function doesn't return in X secs, break out of it. Kinda how Windows does to check for frozen processes. Obviously, this'd need multithreading, but it should not be ~that complex.

jahodfra:
For infinite loops in LUA it is possible to hookup a function on the interpreter which is called after each N operations.
This function can then check if user do not want to quit.

For infinite loops in C, it is possible to ask for that inside each unbounded loop.

Another options is to listen on signals and kill the whole process on the kill signal.

nesbox:
but what if a user doesn't call any function in a loop
``` lua
while true do end
```

nesbox:
as a solution, game could be run in another thread, don't like it

jahodfra:
https://www.lua.org/pil/23.2.html allows to add count hook into the
interpreter. I am not sure how that would affect performance.

I think signals are handled as interrupts so they stop the thread anywhere
(but that also means that you cannot do much with the running thread).

čt 7. 12. 2017 v 16:42 odesílatel Vadim Grigoruk <notifications@github.com>
napsal:

frenetic:
I think if a cartridge has a loop it should stay that way.
TIC-80, on the other hand, shouldnt freeze. Closing TIC-80 should work in normal fashion, without delays.

[...]

PR #463, opened 2017-12-12
Execution timeout error for Lua infinite loops

frenetic:
Adds a method that tries to detect if the Lua script is taking too long.
If so, kill the script and throw a nice error.
Bug: https://github.com/nesbox/TIC-80/issues/197

I only tested that on Ubuntu. I hope it works on all platforms. Sorry...

About performance: it seems to have little to no effect on performance.

Tested with default game:
`callLuaTick` takes an average of 0.000072 seconds to execute, with or without this fix.

Tested with FPS80:
`callLuaTick` takes an average of 0.014970 seconds to execute **WITHOUT** the fix.
`callLuaTick` takes an average of 0.014980 seconds to execute **WITH** the fix.

### DIFF
--- a/src/luaapi.c
+++ b/src/luaapi.c
@@ -25,11 +25,18 @@
 #include <lua.h>
 #include <lauxlib.h>
 #include <lualib.h>
+#include <time.h>
 
 #include "machine.h"
 #include "ext/moonscript.h"
 
 static const char TicMachine[] = "_TIC80";
+
+// infinite loop protection
+static const u32 LUA_LOC_STACK = 1E8; // 100.000.000
+static const u8 MAX_TICK_EXECUTION_TIME = 5;
+clock_t callLuaTick_start;
+
 
 s32 luaopen_lpeg(lua_State *L);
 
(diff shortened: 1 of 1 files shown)


jahodfra:
I think this is good direction. I am wondering about specific details:

Does this protect against infinite loops outside of the TICK function? (I suppose not, can we somehow ensure that)
Does this work for callbacks specified by lua code? e.g. map function allows users to specify callback.
I also think this is maybe too strict, wouldn't be better just to periodically check that the user hasn't requested the termination (e.g. by pressing some keys?)

adiron:
Great stuff. Infinite loops are very annoying to recover from especially if you haven't saved properly (can't say that has never happened to me).

I however don't think that automatically killing a cartridge is the correct course of action in terms of user experience. The key here in my opinion is to not block the user from killing the running cartridge in the event that it does enter a loop of some sort. It's best to keep control within the user's hands and not try to second guess them.

nesbox:
@frenetic I modified your PR and added ESC button handling, pls review it here 18aca45
Thank you

PR #465, opened 2017-12-13
Added Brainfuck language scripting support.

lolbot-iichan:
Added Brainfuck language scripting support.


nesbox:
To be honest, I don't think that Brainfuck is suitable script lang for gamedev and I wasn't going to add it, sorry.
Could it just live in your fork https://github.com/lolbot-iichan/TIC-80?
And we add link to your repo to the Wiki

Thank you

lolbot-iichan:
I don't see what TIC-80 is loosing by getting a useless lang support (except a silly name in changelog), so I'm kind of confused, but well, ok...

So, let's assume I've become a reallife maintainer of a Brainfuck-based Game Engine. Wow! But wait, are there any rules or restrictions? Should I disable Pro features in my fork or can I keep them? Can I rename the app file and change it's title? Can I distribute a free version on Android Market? Can I use TIC-80 mascot image? 

And one more thing, even if I manage to maintain a fork and release it's builds, TIC-80 is more than just an app, there is also a live community & website for sharing samples and games, which this fork will miss. I currently have coded about a dozen of sample cartridges similar to "<api> demo / <lang>" from "Demos" section for most of API functions and there is currently nowhere to publish them online so someone could toy with them. Do you have any recomendations?

P.S. I also had some plans for adding C language scripting support (with tcc) some time after I'm done with brainfuck. Would you reject that addition as well?

nesbox:
so, TIC has MIT license and there are no restrictions, except you have to mention original TIC in your version.

you can use your Github repo and wiki for BF demos I think.

I don't mind about TCC and won't reject it.

Sorry, if I offended you...

josefnpat:
This is very cool, and you should totally keep merging upstream from this repo. If there is really enough desire for a brainfuck language, i'm sure @nesbox would give it a second thought.

I agree with @nesbox in this case though; every language you add, you have to support when a new function is added down the line.

PR #471, opened 2017-12-17
Speedup api_circle

jahodfra:
2.4x speedup for Shapes test in Benchmark tic.


nesbox:
Unfortunately, I have to revert the changes because `circ` and `tri` functions don't work correctly

how circ(r=2) looks after your changes:
image

how it should look
image

how tri looks:
image

how it should look
image

Please fix

Thanks

PR #474, opened 2017-12-18
Speedup api_circle, api_tri.

jahodfra:
I fixed one off errors for circle and tri.
I also squashed everything into single commit so revert would be easier.


nesbox:
I made small fix here 6c6fa6c
it crashed if you draw a circle with y<0

Issue #192, opened 2017-06-16
Indent/Unindent no matter where you are on the line


darkhog:
The problem with intending in TIC-80 is that to intend/unintend a line, you must first get to the first column of it which depending on the user's skill with the keyboard may generate too many additional keystrokes and make it slower to type in general.

To that end, I propose adding following shortcuts to the editor:
- Ctrl+Tab will intend current line no matter where you are in it, as opposed to tab which would just insert a tab character at the current column.
- Ctrl+Shift+Tab will unintend current line by one tab. If it isn't intended already, it will do nothing.

frenetic:
@nesbox or @darkhog: this issue was fixed. Please, close it.

PR #485, opened 2017-12-28
Add support to crtl or shift to indent/unindent anywhere in the line

frenetic:
Adds support to indent or unindent, anywhere in the line, if user is pressing shift or crtl.

shift + tab will unindent.
crtl + tab will indent.

Fixes issue https://github.com/nesbox/TIC-80/issues/192

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -617,20 +617,31 @@ static void redo(Code* code)
 	update(code);
 }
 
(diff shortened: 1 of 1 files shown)


nesbox:
added small fix db261ff
because `history()` saves cursor position too

Issue #103, opened 2017-04-17
Mirror/rotate selection


AnastasiaDunbar:
It should be able to mirror/rotate only for the selected part, but will rotate the whole part instead.

nesbox:
will do

AnastasiaDunbar:
My question was a bit ambiguous for myself and doesn't even mention about the sprite editor, sorry about closing and reopening. Here's a GIF that explains it:
![gif](https://user-images.githubusercontent.com/11806048/32212644-d360d34e-be52-11e7-9a45-0bb73cdbd137.gif)

noxarivis:
i dont think its that important, its more a "nice to have"

its a tool to create retrogames at least

josefnpat:
How would we rotate a selection that is not square? We could rotate and then scale, but that sounds like a bad solution.

AnastasiaDunbar:
But you can't make an irregular selection, that's not necessary because you could just export sprites then open it in a graphics software and import it back but for something small like this might be useful in some cases.

frenetic:
Just added a pull request to solve flipping horizontally and vertically.
It also enables pressing the `trash` button to delete a selection.

The problem now resides on `rotate`.
We need to discuss proper behavior for the following scenario:
![rotate trouble](https://user-images.githubusercontent.com/122976/34542844-14c5a354-f0bd-11e7-870a-bdfe933ea009.gif)

AnastasiaDunbar:
I just discovered a glitch:
![glitch](https://user-images.githubusercontent.com/11806048/34545569-d584dde2-f128-11e7-9edd-31d437ceb1c7.gif)

There could be a way to check if the rotated rectangle is out of boundaries?

AnastasiaDunbar:
Here's a demo I made in TIC-80:
```javascript
// title:  rotation
// author: Anastasia Dunbar
// desc:   hold z and use arrow keys to resize selection, press x to rotate.
// script: js
var region={
	w:20,h:30
},
selection={
	x:0,y:0,w:1,h:1
};
function TIC(){
	if(btn(4)){
		if(btn(0)){selection.h=Math.max(selection.h-1,1);}
		if(btn(1)){selection.h=Math.min(selection.h+1,region.h-selection.y);}
		if(btn(2)){selection.w=Math.max(selection.w-1,1);}
		if(btn(3)){selection.w=Math.min(selection.w+1,region.w-selection.x);}
	}else{
[...]

PR #500, opened 2018-01-06
Rotate sprite selection

frenetic:
Adds selection rotation.
Solves issue https://github.com/nesbox/TIC-80/issues/103

solve rotation

It rotates to the right (as the previous rotation used to do).
It also keeps the selection at the front, so we can keep rotating and moving the selection around.

The only downside now is, to get rid of the selection tool, we have to click the selection tool to disable it.

### DIFF
--- a/src/sprite.c
+++ b/src/sprite.c
@@ -451,6 +451,66 @@ static void rightCanvas(Sprite* sprite)
 	pasteSelection(sprite);
 }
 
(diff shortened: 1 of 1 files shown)


nesbox:
Hey, I added you to the credits https://github.com/nesbox/TIC-80/wiki#credits
Pls fix the link if you want
Thank you for the help

PR #513, opened 2018-01-22
Use ctrl-pageup/pagedown to cycle thru modes.

technomancy:
While I was in the process of creating this patch I discovered the `Alt-1` thru `Alt-5` shortcuts, so maybe this one is redundant, but I thought it might be a nice shortcut to have. Feel free to close if you think it's unnecessary.


nesbox:
pls add info about the shortcuts to the Wiki (with version .70 tag), if not difficult, thanks

Issue #526, opened 2018-02-11
Print the whole stack trace for lua code.


jahodfra:
When the lua code fails it prints just the last line. It is not clear to me how to find out what called the line.

jahodfra:
there is bit of workaround which can be done in Lua.

```lua
local status, error = xpcall(func, debug.traceback)
if not status then trace(error) end
```

nesbox:
done in the PR #530

PR #528, opened 2018-02-15
Add debug library for lua.

jahodfra:
This allows to print tracebacks in lua and simplifies debugging.


frenetic:
@jahodfra does merging this fixes https://github.com/nesbox/TIC-80/issues/526?

jahodfra:
No, it just adds option to printout stack through lua wrapper. I think it would be better to print the stack trace by default.

PR #536, opened 2018-02-20
Textri using constant slope and fixed point math 

Nullious:
saved some space, fixes wibbling across texture


Nullious:
i was logged into wrong github account hence the mess ..

PR #602, opened 2018-05-22
Point iOS version to a fork that supports 0.60.3

lonelyclaud:


nesbox:
thanks for your time 👍

PR #609, opened 2018-05-26
Ignore files/folders generated by cmake Xcode generator on macOS (#320)

armadsen:
If you use cmake's Xcode generator, it creates some folders in build/macosx that weren't ignored. This PR adds those (3rd-party/, CMakeScripts/, and TIC-80.xcodeproj/, and build product folders) to the .gitignore file.


nesbox:
I checked and I see nothing near `build/macosx` folder
closing...

armadsen:
You're right. I invoked cmake in the build/macosx folder which is the only reason it put generated files there. Makes more sense to invoke it from outside the tic-80 folder altogether. Thanks!

PR #611, opened 2018-05-31
Add eval command

technomancy:
I have this working as a new command in console.c. I named it "eval"
so that it can be extended with other languages in the future. However
I think some of this code belongs in a different file. Maybe luaapi.c?
It feels wrong to be calling `lua_*` functions directly from here. Let
me know if you have a suggestion for how to improve this or if I'm
wrong and it's fine where it is.

Also, I haven't used the C API for Lua enough to know whether this is
correct use of the stack; I'm guessing that unless we call
`lua_settop(lua, 0)` the values from the code that's run (or the error
message string) would be left on the stack? Is `lua_settop` the right
way to clear out these values and ensure they get GCed?

The problem with moving the dostring code to another file is that I'm
not sure how you would return an error message from a function in
luaapi.c but still be careful not to call `lua_settop` until after the
string has been printed. Interested in hearing what you think about this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,10 @@ To make a retro styled game, the whole process of creation and execution takes p
 ![TIC-80](https://user-images.githubusercontent.com/1101448/29687467-3ddc432e-8925-11e7-8156-5cec3700cc04.gif)
 
 ### Features
-- Multiple progamming languages: Lua, Moonscript, Javascript and Wren
+- Multiple progamming languages: [Lua](https://www.lua.org),
+  [Moonscript](https://moonscript.org),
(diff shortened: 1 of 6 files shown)


technomancy:
Any thoughts on how I've implemented this PR? Happy to revise if I get some feedback.

technomancy:
Thanks for merging this!

I had an idea that it should probably call `trace` on the return value automatically in Fennel, because Fennel has no statements, only expressions. So I might make a follow-up pull request for that.

I would also like the same thing in Lua, except it is problematic because Lua has statements as well as values. It's a bit annoying but doable to tell the difference when writing Lua code, but doing it from C I don't know if there's a good way to do that.

Issue #661, opened 2018-08-12
Caps Lock not working


VinylMelody:
VERSION: 0.70.1

caps lock suddenly broke
have to use shift when entering capital letters
it doesn't appear at version .60.3 :/

nesbox:
fixed with #655

PR #690, opened 2018-08-19
Attempt to fix keyboard layout issues (WIP)

### LINKED ISSUE #655: Keyboard mappings broken!
The TIC-80 thinks my nordic keyboard is American(or british).
**Really annoying**. I don't have this problem in 0.60

db0z:
This approach uses SDL_TextInputEvent a bit similarly to 0.60.3; The 1st byte out of 16 currently unused is now used to store one of the characters that were input in the current frame or 0 if there weren't any. getKeyboardText() simply returns it. It's readable with peek(65420) and can be also made available via an API function, for convenience.

Note: I didn't know how to convince the compiler that tic80_input should be 13 bytes, so it's 16 instead, with 12 bytes remaining.

Also, this breaks the android touch keyboard. Probably. I don't know how to build for android. It might be fixable by simply using the code I removed from getKeyboardText() to set text_char, but only when the touch KB is in use.

but fixes #655, #661

### DIFF
--- a/include/tic80.h
+++ b/include/tic80.h
@@ -121,7 +121,7 @@ typedef struct
 	tic80_gamepads gamepads;
 	tic80_mouse mouse;
 	tic80_keyboard keyboard;
-
+	s8 text_char;
 } tic80_input;
 
 TIC80_API tic80* tic80_create(s32 samplerate);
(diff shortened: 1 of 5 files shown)


nesbox:
fixed here 607ebd6, sorry
and thank you for the effort ;)

PR #695, opened 2018-08-23
Fix bass-tones by smoothing waveforms

shaggyfox:
Hey, i noticed a very noticeable 'overtone' on low notes. Here I've a small patch, that smooths the waveform so that there are no unwanted spikes that produce those overtones.


nesbox:
Could you pls attach an example cart with the _overtone_ demo to see the difference?
Thanks

nesbox:
I compared the sound output with the changes you made and with Famitracker output, it shouldn't sound like this.

PR #721, opened 2018-09-07
Octave shortcut collides with cartridge save shortcut

JasonNumberThirteen:
I noticed when everytime I want to save my project while being in SFX editor using Ctrl+S, it automatically changes sound octave to C# because it's attached to the "S" key which causes collision. IMHO, it might be nicely to make some detection which changes octave only when user doesn't hold the Ctrl key.


nesbox:
ok, I created an issue https://github.com/nesbox/TIC-80/issues/722
thanks

PR #724, opened 2018-09-09
Sound and music exporter

JasonNumberThirteen:
I have an idea to implement to "TIC-80" some kind of WAV and MP3 exporter (for sound and music respectively) to make normal audio file readable by OS. I understand that you have a lot of more important things to do, but this as a future feature it would seems very comfortable.


nesbox:
moved to #725
pls use issues to create task instead pull request

StinkerB06:
Can you please have the channel on/off buttons effective on music output to make different files of different channels?

PR #738, opened 2018-09-23
Wren outline lets you navigate by class.

jminor:
This PR adds support for the outline mode when coding in Wren.

It navigates from class to class. It would be better to navigate method to method, but parsing that is a bit harder. This at least gives you something useful than an empty outline :)

The code is mostly copied from jsapi.c's `getJsOutline()`


nesbox:
yes, please rebease it to the master.
thanks

nesbox:
merged here https://github.com/nesbox/TIC-80/pull/744#event-1863687933

PR #742, opened 2018-09-24
Add install cmake step to build with VS2017 instruction

SagePtr:


nesbox:
VS2017 installer has CMake inside

PR #759, opened 2018-10-14
Add squirrel language support to TIC-80

GrahamClemo:
Squirrel is a scripting language with a C-like syntax, lists, tables, classes, generators, coroutines etc.
See http://squirrel-lang.org/ for more details.
This PR adds Squirrel language support to TIC-80, it is dependent on this PR (https://github.com/nesbox/3rd-party/pull/5) which adds Squirrel to the 3rd-party libraries repo.


GrahamClemo:
I thought I did that already! Let me check if it's got lost in git somewhere. Thanks.

PR #771, opened 2018-11-04
Can set parameters and specific cartridge through arguments option.

taketin:
I wanted to set an optional parameter (for example `-uiscale`) and specific cartridge when initialize time on `wasm`.
Then I wrote the following code in Javascript.
```javascript
<script type="text/javascript">
  var Module = {
    canvas: document.getElementById('canvas'),
    arguments: ["tetris.tic", "-uiscale", "1.5"]
  };
</script>
```		

However, it couldn't work as my expected. An optional parameter was worked but didn't boot the cartridge.

I investigate the cause then found it. This p/r is my proposal.

My idea is that if `argc[1]` has `.tic` to suffix, we judge to it was given a cartridge.

Please take a look if you have time.
Regards.

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -1528,15 +1528,21 @@ static void onEmscriptenWgetError(const char* error) {}
 
 static void emsStart(s32 argc, char **argv, const char* folder)
 {
-	if(argc == 2)
+	if (argc >= 2)
 	{
-		startVars.argc = argc;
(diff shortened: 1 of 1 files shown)


nesbox:
sorry for the delay with the review

taketin:
Thanks for merged it!

PR #779, opened 2018-12-01
In iOS 11 error: 'system' is unavailable: not available on iOS 

FreeBlues:
Solution:

in loslib.c change the

```
int stat = system(cmd);
```

to

```
int stat = nftw(cmd, unlink_cb, 64, FTW_DEPTH | FTW_PHYS);
```

add 

```
#include <ftw.h>
```

add new function:

```
int unlink_cb(const char *fpath, const struct stat *sb, int typeflag, struct FTW     *ftwbuf)
{
    int rv = remove(fpath);
    
    if (rv)
        perror(fpath);
    
    return rv;
}
```


nesbox:
can't merge the PR, a lot of conflicts, sorry

PR #819, opened 2019-03-14
Workaround for freeze on fullscreen under macOS

kawa-yoiko:
Hi all ^ ^

On macOS 10.13.5, the TIC-80 window freezes when fullscreen mode is toggled. Reproducible with both official release and binary built from source.

Only the SDL backend has the problem, and I've traced down the issue to be an upstream issue reported in [SDL bugzilla](https://bugzilla.libsdl.org/show_bug.cgi?id=4369). In short, the Cocoa implementation of `SDL_PumpEvents()` has a weird behaviour — it may report "no events" during one call and find an event in the next (called in a row without any delay). This would result in some events not processed and thus freezing the SDL window.

I suppose it's better to research a bit more and report it to SDL, but I have come up with a workaround for now. It's very simple, but it's inconvenient for me to confirm it doesn't break on other platforms, so please help with that. Thanks in advance!

### DIFF
--- a/src/system/sdlgpu.c
+++ b/src/system/sdlgpu.c
@@ -776,6 +776,7 @@ static void pollEvent()
 
 	SDL_Event event;
 
+	SDL_PumpEvents();
 	while(SDL_PollEvent(&event))
 	{
 		switch(event.type)


kawa-yoiko:
FYI, the bug can be triggered by changing [the event loop](https://github.com/nesbox/TIC-80/pull/819/commits/a7c7b658f933d1df8c3a16b04256d48c7f19be52#diff-1d81140f7c676a6f02efee699ae6a123R780) as follows:

```cpp
for (int x = 0; x < 3; ++x) {
    while (SDL_PollEvent(&event)) {
    // ...
    case ...:
        printf("round #%d\n", x + 1);
    }
}
```

Ideally the output should all be "round 1", but actually there are occasionally "round 2" lines.

nesbox:
@kawa-yoiko could you please reproduce it on the latest build, because all is ok on my MacBook?
Thank you.

kawa-yoiko:
Apologies for the late response!

I'm on macOS 10.13.6 and the problem persists with the latest revision 225b67a. The SDL issue mentioned in the original comment has been recently marked as resolved, and after manually upgrading SDL to 2.0.12 it gets way better but still occasionally happens (window does not refresh until a keypress or focus loss). With this workaround patch applied everything goes well. I can provide a screencast if needed; or if it's deemed a minor issue it might as well be fine to leave it as is.

nesbox:
I updated SDL to 2.0.12 here d0d961e
and merged your PR here 2fe6166
Thank you.

kawa-yoiko:
Thanks for the effort 🎉 🚀

I've done a fresh rebuild and the freeze is gone here now.

PR #821, opened 2019-03-17
Fix (and simplify) blurry Sokol rendering

kawa-yoiko:
I'm by no means experienced with Sokol, but I somehow managed to get a fix.

The first commit already fixes the blurriness, while the second one e968db3 aims at getting rid of the unnecessary upscaling process taken from [the examples](https://github.com/floooh/chips-test/blob/4984c9df664e996dcfedcf5faea1ac13ad99cd0c/examples/common/gfx.h), but I cannot conveniently test the output on platforms other than macOS/Metal. In particular, I'm not sure whether the image will flip vertically with different backends (with left-handed and right-handed cooordinate systems), so this should not be merged until anyone helps confirm the correctness on other devices — I'd appreciate it!

Apologies for flooding with PRs, and thanks for your effort and time (^ ^;)


nesbox:
Hi, thank you for the PR, I'll review and approve it soon...

kawa-yoiko:
Oooh my carelessness… Thanks >_<

On a sided note, you have made sure it works correctly on other platforms such as Windows and Linux, right?

nesbox:
I tested it on MacOS and Windows, works correctly.
Thank you :)

PR #847, opened 2019-06-01
cmake: Allow disabling Sokol/SDL

RobLoach:
This change allows disabling building the Sokol and SDL projects conditionally.

```
cmake . -DBUILD_SOKOL=OFF
```


RobLoach:
Merged into the libretro branch.

PR #856, opened 2019-06-23
cmake: Fix directory resolution

RobLoach:
This prefixes any relative path with a CMAKE_SOURCE_DIR, so that path resolution works when building from a different directory.


nesbox:
Also, pls don't forget to add yourself to the Credits section here https://github.com/nesbox/TIC-80/wiki#credits

PR #877, opened 2019-07-17
Silence buffer overflow compiler warning

RobLoach:
The compiler was complaining about a potential buffer overlow error, so I switched it to use snprintf instead, and then silenced the warning.


nesbox:
same on Windows https://ci.appveyor.com/project/nesbox/tic-80/builds/26034408/job/wcir1t9isao0riy8#L2005

nesbox:
only Clang is successfully built

RobLoach:
Thanks for the review... I'll remove that compile option, and leave in the snprintf().

PR #904, opened 2019-08-14
libretro: Update Core Options API

RobLoach:
This allows for translations for the core options.


RobLoach:
Thanks for a heads up. I'll keep watch on the builds to see if things are still good :+1:

Issue #902, opened 2019-08-08
Palette loading bug


StinkerB06:
First of all, the *.tic file format allows for chunk sizes smaller than the memory associated with them. This is a very simple compression method for carts, in case you have zero-filled bytes left at the end of the memory region. The remaining bytes that are not stored in the file are therefore zero when the chunk is loaded.

But this isn't fully true for the 48-byte palette memory though. When TIC-80 saves a cart, it applies this compression method to the palette chunk in the file (along with the sprites, map, sound, etc. chunks). The loading process is the slightly different part for palettes.

Instead of filling the remaining palette memory with zeroes (or black colors/intensities in this case), it instead uses the DB16 palette's byte values at offsets _n_...47 (where _n_ is both the size of the palette chunk in the cartridge file, and the offset of the first palette memory byte to not be present in the chunk). This may ruin a game's palette.

**Repro Steps:**
1. Create a new cart with `new` command.
2. Paste this palette string into the palette editor (saved chunk is 8 bytes, 40 bytes are filled):
```
00000000000000ff00000000000000000000000000000000000000000000000000000000000000000000000000000000
```
3. Save the cart with whatever filename.
4. Re-load the cart you just saved.

**Expected Result:** The palette should be two blacks, followed by a green and thirteen more blacks.
**Preserved Result:** The palette instead becomes two blacks, followed by a mint green, and then DB16 colors 3...15.

A workaround is to make color 15 have the darkest level of blue, i.e. RGB `#xxxx01`.

RobLoach:
Worked for me... After I saved and quit, and re-ran, the palette data was saved.

![Screenshot at 2019-08-08 03-34-53](https://user-images.githubusercontent.com/25086/62683908-81c1cd80-b98d-11e9-937b-018b14adc9bc.png)

StinkerB06:
I get something like this instead:
![TIC-80 Palette Bug](https://user-images.githubusercontent.com/30059831/62744329-00a11f80-b9fb-11e9-9108-f48d26c0f345.gif)

RobLoach:
What version are you running?

StinkerB06:
I'm running 0.80.0-dev Pro, which I compiled using the MinGW build
instructions. The bug also persists in 0.70.6.

kawa-yoiko:
I have a fix in #915. The problem does not occur with carts saved in plain text format, which I guess is the reason @RobLoach did not reproduce it ^ ^

StinkerB06:
[...]

PR #915, opened 2019-08-29
Fix palette loading #902

kawa-yoiko:
From the semantics in `console.c`, it appears that the `palette` argument in `api_load()` denotes whether the palette should be loaded from the cart, i.e. if it is false, the palette will be overwritten with the one in the configuration.

https://github.com/nesbox/TIC-80/blob/634f3e7d7d78fdc0a0d7f3c2ccf2ce0f5356fa13/src/console.c#L377-L383

Also, I removed a redundant line of code. I have tested this with different combinations of cart loads, section loads and cart saves (both binary and text).


nesbox:
As I see in the comments https://github.com/nesbox/TIC-80/issues/902#issuecomment-526449452 the PR doesn't fix the issue with a palette. Is it true? @kawa-yoiko

kawa-yoiko:
We are unsure why RobLoach did not meet the problem at all, but I reproduced it and fixed it with this change (macOS with Clang). It would always be nice if more people could try this.

StinkerB06:
Can he record a video of him trying to reproduce the bug?

nesbox:
fixed here 15eb3a4
closing PR...

Issue #907, opened 2019-08-16
Disabling BUILD_DEMO_CARTS breaks the build


RobLoach:
After some testing on https://github.com/nesbox/TIC-80/commit/4eef02123e44de6e80392f4e69fd2ad4b5fbc556#diff-75218e1e64b8201a55c2ac2203e8970cL293 , I've found disabling building the demo carts can break some builds....

```
git clone https://github.com/nesbox/TIC-80.git
cd TIC-80
git submodule update --init --recursive
cd build
cmake .. -DBUILD_SDL=0 -DBUILD_SOKOL=0 -DBUILD_LIBRETRO=1 -DBUILD_DEMO_CARTS=0
make
```

```
[ 83%] Building C object CMakeFiles/tic80lib.dir/src/console.c.o
/home/rob/Documents/TIC-80/src/console.c: In function ‘getDemoCart’:
/home/rob/Documents/TIC-80/src/console.c:518:14: fatal error: ../bin/assets/luademo.tic.dat: No such file or directory
     #include "../bin/assets/luademo.tic.dat"
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
CMakeFiles/tic80lib.dir/build.make:75: recipe for target 'CMakeFiles/tic80lib.dir/src/console.c.o' failed
make[2]: *** [CMakeFiles/tic80lib.dir/src/console.c.o] Error 1
CMakeFiles/Makefile2:78: recipe for target 'CMakeFiles/tic80lib.dir/all' failed
make[1]: *** [CMakeFiles/tic80lib.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
```

This may affect Android and Emscripten builds, since they have the demo carts disabled by default.

nesbox:
You have to call `bash build_demos.sh` command before build if you disabled `DBUILD_DEMO_CARTS ` flag.

msx80:
Are we sure that the way demo carts are handled is good? There's a number of issues with it. For example, building the bin2txt utility and then using it breaks any crosscompilation attempt. Also the files get placed in an unintuitive folder where they're probably not "cleaned" (i think this is why it often works when trying special builds, becouse they were produced in previous builds).

the build_demos.sh script solve some of these issues but it's outside of the build scripts and is not mentioned in documentation.

[...]

Issue #912, opened 2019-08-21
Integrating Raspberry Baremetal build


msx80:
Hi there, to continue discussion on #907 , i made some tests and updated the build of my baremetal port to the latest TIC80 sources.

In the end, modifications on the codebase is limited to the following:

- fs.c: this is the biggest change, most of the primitive functions in here need a different implementation.
- tic80_types.h: need changes to avoid a type conflict.
- CMakeList.txt: a block for Baremetal is added at the beginning as a preamble, to set all variables etc.

Everything else is new files that could go in any folder.
The strategy i used is to build tic80lib as usual (i haven't added a new target) with BAREMETAL switch, which is basically just crosscompiling it to ARM with some custom stuff for RPI. Then i made a separate program (the kernel), based on circle examples code, that link to the tic80lib.a. This is a separate Makefile that include stuff from circle and circle-stdlib, not sure it is possible to include it in the main CMake build file.

I think we could put the kernel on src/system/baremetalrpi and the Makefile on build/baremetalrpi, or some similar combination. In build/baremetalrpi i can put a readme with all steps to checkout circle and build everything.

If you're ok i can prepare a pull request (in a couple of weeks, i'm leaving for vacation soon :D ). Obviously everything can be changed later if we find a better fit.

I was able to build TIC80 with all languages except squirrel (for some reason it breaks crosscompilation, i think it has some dangling arm switch inside somewhere).

nesbox:
Sounds awesome, as for `fs.c` I think we could make some kind of abstraction layer for the file system operations, will do after your PR...
Thank you.

nesbox:
will setup CI scripts soon
thanks

msx80:
Hi there, you may want to wait for my next pr (with some more hardware support) before setting up ci, as the building instructions have changed a little

nesbox:
Hi,
I tried to setup CI for the bare-metal build but with no success, seems we are a bit outdated with the Circle repo now.
@msx80 could you pls look at?
Thank you.

msx80:
Hi there, i'll take a look! With RPi4 and RPI400 it starts to get interesting.
But i'll take some time as i'm a bit busy lately..

nesbox:
Hi, thank you 👍

msx80:
I gave a try, i've updated some stuff and can compile some of the libs, but zip is giving me errors.

`In file included from /home/nicola/TIC-80/vendor/zip/src/zip.c:47:
[...]

PR #916, opened 2019-08-30
Add Demo Carts

RobLoach:
This adds the compiled demo carts directly into the repository.

I understand this is not the optimal solution, but was still failing on ARM devices. Seemingly the only way we could get TIC-80 to compile on irregular ARM devices.

Barebones may be an option, but this is easiest and most reliable for now.

## References
- Fixes https://github.com/nesbox/TIC-80/issues/907
- Related to https://github.com/nesbox/TIC-80/issues/912


RobLoach:
Appveyor failing is unrelated. Issue over at https://github.com/nesbox/TIC-80/issues/917

RobLoach:
Thank you for merging this. I know as developers, we all want perfection, and this is not perfect. Does help building for embedded platforms though, so thank you.

PR #918, opened 2019-08-31
Baremetal Raspberry PI build support

msx80:
Hi there, the changes implement #912, make it possible to build TIC-80 for Raspberry PI in baremetal mode. There's a new folder under build, baremetalpi, with all instructions. Actual sources are under src/system/baremetalpi

The build should work for both RPI 2 and 3.


msx80:
wops the fix of surf accessing the disk (#919) was meant to stay on my repository, accidentaly ended up here, i'm still learning the magic of git :P Well it should work anyway

build/baremetalpi/Makefile
```diff
@@ -0,0 +1,47 @@
+#
+# Makefile
```
RobLoach: Would you like help porting this to CMake?

build/baremetalpi/boot/Makefile
```diff
@@ -0,0 +1,14 @@
...
+#
+# These files must be copied along with the generated kernel.img
+#     onto a SD(HC) card with FAT file system.
+#
+
+firmware:
+	wget -q -O LICENCE.broadcom https://github.com/raspberrypi/firmware/blob/master/boot/LICENCE.broadcom?raw=true
+	wget -q -O bootcode.bin https://github.com/raspberrypi/firmware/blob/master/boot/bootcode.bin?raw=true
+	wget -q -O fixup.dat https://github.com/raspberrypi/firmware/blob/master/boot/fixup.dat?raw=true
+	wget -q -O start.elf https://github.com/raspberrypi/firmware/blob/master/boot/start.elf?raw=true
```
RobLoach: Could likely use a git submodule for these? Unless the firmware is crazy large.

msx80: it's 2 or 3 MB IIRC. This is the same Makefile they have on circle, I haven't inquired but i think there's some legal reason not to include them in other projects. Or maybe it's just to get the latest version? Anyway, i think this works ok, it's a one time task anyway :)

build/baremetalpi/Makefile
msx80: Well, that would be awesome but there's some issues: that Makefile includes a lot of stuff from circle .mk files. So first of all, it won't be easy, and second, if we "copy" over some of that stuff, we diverge and stay behind in case circle adjust something. I think we'll better stay like this, at least for now (i'm still making some adjustment).

Btw have you tried the port maybe? :)

RobLoach: Ah, I see...
```
include $(CIRCLEHOME)/Rules.mk
```

PR #963, opened 2019-11-27
Fennel 0.3.0 syntax

technomancy:
This updates the syntax highlighting for new keywords in Fennel.

It also disables "strict globals" mode to make it more forgiving like Lua.

One question: it appears that the file `demos/fenneldemo.tic` is supposed to be loaded when you run `new fennel`. However, the actual contents that get loaded are slightly different; there is a stray `>` character that is not there in the `demos/` file. In addition the number passed to `cls` is different. Do you have any idea which file needs to get updated to fix this?

### DIFF
--- a/src/luaapi.c
+++ b/src/luaapi.c
@@ -1632,7 +1632,8 @@ const tic_script_config* getMoonScriptConfig()
 #define FENNEL_CODE(...) #__VA_ARGS__
 
 static const char* execute_fennel_src = FENNEL_CODE(
-  local ok, msg = pcall(require('fennel').eval, ..., {filename="game", correlate=true})
+  local opts = {filename="game", correlate=true, allowedGlobals=false}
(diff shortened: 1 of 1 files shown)


technomancy:
Thank you!

Do you know what could be causing the stray > character which is not present in the source?

nesbox:
It's weird because I don't see stray char on my side, both carts `demos/fenneldemo.tic` and `new fennel` are the same. Seems you have changed `default_fennel.tic` in the `.local` folder, do you?
Thanks.

nesbox:
Another thing, I see strange garbage in the map section of the `fenneldemo.tic`
image

technomancy:
Thanks; the problem on my machine was a stray file in `~/.local` that was overriding the built-in default cart. Once I deleted it I saw what you were seeing. I'll open a new pull request to clear out the map garbage.

PR #967, opened 2019-12-02
Misc code editor additions/fixes

wadetb:
- Ctrl+backspace, Ctrl+delete delete adjacent words.
- Ctrl+x without selection cuts the entire line. Previously
it was implemented but buggy; it copied the line but didn't have a
selection to replace.
- Outline match supports non-consecutive forward matches, for
instance "dph" will match "drawPlayerHit". Matching characters are
highlighted in the outline view.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -472,6 +472,37 @@ static void backspaceChar(Code* code)
 	}
 }
 
+static void deleteWord(Code* code)
+{
+	const char* end = code->src + strlen(code->src);
+	char* pos = code->cursor.position;
+
+	if(pos < end)
+	{
+		if(isalnum_(*pos)) while(pos < end && isalnum_(*pos)) pos++;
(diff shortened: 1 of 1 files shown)


wadetb:
This is just some minor editor stuff that I found useful, feel free to take it or leave it.

wadetb:
code-editor-1
code-editor-2

wadetb:
Sure- I took a quick look and found that it's two separate issues with the undo system. 

If there is no change (e.g. the cursor doesn't move) there will be no list entry created by `history_add`. You can repro the same behavior with just the delete key. 

Also, undo state is recorded _after_ the action, so the first edit made will always undo the cursor back to the top of the file.

I'll spend some time later and try to fix both issues without causing other problems.

nesbox:
Ahh, it looks like a more general issue and doesn't depend on the PR, I'll merge it and create a separate issue then, it will be better.
Thanks :)

PR #968, opened 2019-12-02
Map view fixes when wrapping around the edge

wadetb:
- When trying to select across the edge of the map, the selection
rectangle no longer jumps and stretches across the entire map.

The mouse coordinates were being normalized prior to expanding the
selection extent, so for example a 1,1 start point and a -1,-1 end point
would evaluate to 1,1..W-1,H-1, turning a 2x2 selection into the entire
map!.

It's actually fine for the selection rectangle to go outside the valid
range because the copy/paste functions all normalize the individual
gets/sets.

- Don't allow scrolling into negative offsets in the world view. This
breaks the grid drawing code temporarily when you switch back to the
map.

- When the map rectangle wraps around in the world view, draw all the
wrapped quadrants. This makes it clearer how the map wraps around and
avoids "losing" the viewport when you scroll slightly left or up.

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -648,8 +648,6 @@ static void processMouseSelectMode(Map* map)
 				mx /= TIC_SPRITESIZE;
 				my /= TIC_SPRITESIZE;
 
-				normalizeMapRect(&mx, &my);
-
 				if(map->select.drag)
 				{
 					s32 rl = MIN(mx, map->select.start.x);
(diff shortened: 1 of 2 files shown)


wadetb:
Here are some quick GIFs of the bugs:
map1
map2
map3

wadetb:
And the fixes:
newmap1
newmap2

PR #971, opened 2019-12-08
Fix MSVC build issues

wadetb:
1. Disable shared SDL build, fixes various _fltused linker errors when building the DLL.

2. Make Sokol build WIN32 in Debug (never CONSOLE) because sokol_app defines WinMain, not main. Fixes linking tic80-sokol under VS2019.

It would be possible to retain the Debug-is-CONSOLE by defining SOKOL_WIN32_FORCE_MAIN as a PUBLIC compiler flag for the sokol library, but this would require changes to player-sokol.

Also, some minor formatting changes which helped me understand the file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,59 +24,57 @@ option(BUILD_PLAYER "Build standalone players" ${BUILD_PLAYER_DEFAULT})
 
 if (BAREMETALPI)
 
-set(BUILD_SDL off)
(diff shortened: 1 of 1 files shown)


nesbox:
please, don't forget to add yourself to the credits section https://github.com/nesbox/TIC-80/wiki#credits :)

wadetb:
Added myself, thanks for the invitation!

PR #975, opened 2019-12-11
Touch keyboard tweaks and fix

wadetb:
Hi, this is two separate changes to the touch keyboard. They can be separated and taken independently if you want. 

The first is some usability stuff I encountered - making the keyboard appear at startup and after resume, and squelching the keypress that makes it appear.

The second is an issue I saw people talking about in Discord and is a pretty straightforward bug.


StinkerB06:
Also, can you please change it, so that using `input:  keyboard` metadata will cause the keyboard to show up instead of the gamepad?

wadetb:
Sure, I'll look into that when I have some more time.

Issue #969, opened 2019-12-03
The Code Editor doesn't save cursor position history when using DEL button


nesbox:
more info here https://github.com/nesbox/TIC-80/pull/967#issuecomment-560600406

wadetb:
I have this fixed locally, but it needs some cleanup.

nesbox:
fixed here 4fa1d5988145a334a3fe0fc9a032ad551e797e7d

PR #985, opened 2019-12-29
Code cursor undo change

wadetb:
Fixes the cursor getting out of sync when undoing and redoing in the
code window.

Under the hood this is quite complicated, but I hope it produces the
desired effect and also it may enable some other undo improvements in
the future.

The principle is that when undoing a code change, you want the cursor to
end up where it was before the code change, and when redoing a code
change, you want the cursor to end up where it was _after_ the change.

To accomplish this, cursor undo is stored with a TAG (before or after).
When the code is modified by any means, both a BEFORE and an AFTER
cursor tag are added. When undoing, the cursor history is rewound to the
next BEFORE tag, and when redoing, it is forwarded to the next AFTER
tag.

One issue that is handled is the case of an edit after undoing. Normally,
edits creates a BABA... alternation in the cursor undo history.
If the user makes some changes, undoes a change, and then makes more
changes, you can get a BABBA where the BB is the point of the undone
change. To resolve this, when the tag being added to the history matches
the latest tag in the history, the latest tag is replaced so there won't
be two of the same tag in the sequence. I'm on the fence about whether
this behavior belongs in the history system or in the code, but there
are no other users of the tags yet so I'll leave it as written. In the
future the tags system might be used to implement "word" undo ala
VSCode, or stroke undo (instead of per-pixel) in the sprite editor. But
I'm not sure, these features might better benefit from an overhaul of
the undo system, I'd need to research other possibilities.

As a simplification, code changes are now flagged in a bool, which
triggers all the syntax highlighting, undo, etc. This may have fixed a
few places where these were applied redundantly.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -39,11 +39,8 @@ struct OutlineItem
 #define OUTLINE_SIZE ((TIC80_HEIGHT - TOOLBAR_SIZE*2)/TIC_FONT_HEIGHT)
 #define OUTLINE_ITEMS_SIZE (OUTLINE_SIZE * sizeof(OutlineItem))
 
-static void history(Code* code)
-{
-	if(history_add(code->history))
-		history_add(code->cursorHistory);
-}
+#define CURSOR_BEFORE_EDIT 1
+#define CURSOR_AFTER_EDIT 2
 
 static void drawStatus(Code* code)
 {
@@ -440,10 +437,7 @@ static bool replaceSelection(Code* code)
 		code->cursor.position = start;
 		code->cursor.selection = NULL;
 
-		history(code);
-
-		parseSyntaxColor(code);
-
+		code->changed = true;
 		return true;
 	}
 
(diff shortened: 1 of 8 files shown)


wadetb:
Note, the code changes aren't as extreme as the comment might suggest, but I hard a hard time justifying parts of it, thus the extensive commit msg :)

If you know of other apps with a decent undo system that handle things a better way, lmk and I'll investigate.

nesbox:
@wadetb sorry for the delay, I'll review it soon :)

nesbox:
I found the simplest solution, I'm going to store cursor position near bookmarks data, somewhere here https://github.com/nesbox/TIC-80/blob/0995ce5e640b6f0ec808c294123a7622f169cbf1/src/studio/editors/code.h#L59
This will allow us to handle cursor history without any problems.
Will be fixed with #969

Issue #983, opened 2019-12-22
(0.80.0-dev) Some carts that poke to sound registers are broken


StinkerB06:
After the effect-column update, I've noticed that some carts [like this one](https://tic.computer/play?cart=975) don't produce any software-generated sound anymore.

The reason for this is because of a change in the stereo registers at `0x0FF98` to `0x0FF9B`. The 4-bit values for the left and right balance levels used to be _**inverted**_ prior to the update (0 becomes 15, 1 becomes 14, etc.). As a result, poking `0x00` (the default value) to either of the four stereo registers causes the level to be 15 (the maximum volume) on both sides.

Now after the update, the left/right level registers aren't inverted anymore. This means writing `0x00` will _directly_ set both sides to 0 (mute), and now you'll need to write `0xFF` to set both sides to max. **Wait, what? Pre-0.80.0-dev carts don't write here!**

The [Fallspire](https://tic.computer/play?cart=806) demo isn't affected by this change, because that calls `sfx()` on all four channels, therefore setting all four stereo registers to `0xFF`.

nesbox:
Stereo registers aren't released yet, it's only .80 feature.
I think there is something else that caused the issue if your cart works on .70

StinkerB06:
1. I know that they are only on 0.80.0-dev.
2. Nah, I'm pretty sure the stereo registers are being defaulted to `0x00` instead of `0xFF`, when a channel is not playing SFX or part of music (that play when the `sfx()` or `music()` function is called). The `NEW TYPE OF MUSIC` cart linked above mixes drum sounds on one channel (which are played using `sfx()`), and some other weird sound on another channel (by ONLY poking to the `0x0FF9C` to `0x0FFE3` RAM region.)

StinkerB06:
@nesbox?

nesbox:
I see the problem, stereo registers value have to be `0xF-volume`

StinkerB06:
Yup. You did this exact same thing when implementing the volume column into the music tracker back in version 0.27.0.

StinkerB06:
> I see the problem, stereo registers value have to be `0xF-volume`

Will you make a fix commit or so, please?

StinkerB06:
@RobLoach But @nesbox isn't responding! Please, fix it sooner rather than later. I know you might be busy but this is going downhill for almost a month.  I think you got too obsessed with the `Mxy` command in the tracker and messed up the stereo registers implementation.

RobLoach:
[...]

PR #993, opened 2020-01-18
cmake: Fix build when SDL is disabled

RobLoach:
When SDL is disabled, don't attempt to build the SDL player.


StinkerB06:
Can you please fix #983?

RobLoach:
Looks like nesbox sees what the issue is there. Please don't derail Pull Request conversations. Keep the sound issue conversation in that thread :wink:

PR #1021, opened 2020-03-17
Add 3ds compilation

phcoder:


phcoder:
> could you pls add
> 
> ```c
> #ifdef DISABLE_NETWORKING
> void* netGetRequest(Net* net, const char* path, s32* size)
> {
>     return NULL;
> }
> #else
> ...
> #endif
> ```
> 
> to the `net.c` and remove `net-dummy.c`
> thanks for the PR :)

Done

merge_static.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
RobLoach: Is the root directory the best place for merge_static.sh? Perhaps in `src` instead?

phcoder: I don't care too much either way

PR #1024, opened 2020-03-22
Fix resulting name for libretro emscripten variant

phcoder:


RobLoach:
The Travis failure is unrelated to this change.

PR #1028, opened 2020-03-22
Split up 3rd-party submodule into individual vendor submodules

RobLoach:
Using the 3rd-party submodule has caused some problems when updating dependencies. While the idea is sound, it makes synchronising vendor code difficult. You have to update the [`3rd-party`](https://github.com/nesbox/3rd-party.git) repository, and then update the TIC-80 submodule to reference the fork.

This change splits the 3rd-party submodule into individual submodules in the `vendor` directory. The same code is checked out, so there isn't any change in how TIC-80 functions. All this does is makes updating third-party vendor code easier since you can then just update it in one place.


RobLoach:
> also, I think, we have to mention all these submodules in the readme

Agreed. There is docs to run `git clone --recursive`, but have a specific call out is a great idea.

RobLoach:
Also, thanks! I can completely understand its use and purpose. Just added a little bit of complexity. Hopefully taking this approach we don't lose out on some of the `3rd-party` repo benefits.

PR #1037, opened 2020-03-26
Don't add -s USE_SDL=2 with libretro-emscripten

phcoder:
Apparently it has some weird conflict

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,11 +148,6 @@ if(MSVC)
 else()
 
     set(CMAKE_C_STANDARD 99)
-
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -148,11 +148,6 @@ if(MSVC)
 else()
 
     set(CMAKE_C_STANDARD 99)
-
-    if(EMSCRIPTEN)
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -s USE_SDL=2")
```
RobLoach: Thanks! This definitely cleans up the entries.

Issue #841, opened 2019-05-21
Persistent real time clock


lf94:
In many retro consoles, they provided clocks inside cartridges to keep time for various reasons. The best example of this is in Pokemon Gold and Silver for tracking the time of day. Providing a `prtc()` or `clock()` function would be great!

RobLoach:
`os.time`?
https://www.lua.org/pil/22.1.html

lf94:
Would that do the trick? On the surface I want to say yes.

RobLoach:
Try it, if it works, then great!

StinkerB06:
The `os` library was removed from TIC-80.

lf94:
> Try it, if it works, then great!

I opened this issue not for myself, but for others :) I have no intention on using TIC-80 at this time.

StinkerB06:
`print(os)` just says `nil`, and trying to use any of its functions (like `print(os.time())`) crashes the cart.

RobLoach:
[`time()`](https://github.com/nesbox/TIC-80/wiki/time) is being used already. We could introduce a `date()` function that returns an array.

``` lua
local dateTable = date()
local year = dateTable[0]
local month = dateTable[1]
local day = dateTable[2]
local hour = dateTable[3]
local minute = dateTable[4]
local seconds = dateTable[5]
local milliseconds = dateTable[6]
```

RobLoach:
Hmmm, I think having `date()` return the UNIX timestamp may be the best, most portable, option.

PR #1045, opened 2020-04-09
Add timestamp() function

RobLoach:
This adds a `timestamp()` function to allow grabbing the [Unix timestamp](https://en.wikipedia.org/wiki/Unix_time) in seconds. This would allow making persistent games by storing the timestamp in `pmem()`, and then calculate elapsed time between game runs.

``` lua
currentTime = timestamp()
```

Fixes https://github.com/nesbox/TIC-80/issues/841

# Wiki entry

## timestamp
timetamp    -> The current unix timestamp in seconds.

## Output:
* **seconds** : the number of _seconds_ that have passed since January 1st, 1970. 

## Description:
This function returns the number of _seconds_ elapsed since January 1st, 1970. Useful for creating persistent games which evolve over time between plays.

## Example

Example

``` lua
-- title: timestamp demo

elapsed = -1

function TIC()
  cls(15)

  -- Display the current time stamp
  current = timestamp()
  print('Timestamp: ' .. current, 10, 10, 1)

  -- Calculate how long ago they last played
  last = pmem(0)
  if last <= 0 then
    last = current
  end
  if elapsed == -1 then
    elapsed = current - last
  end
  pmem(0, current)

  -- Display the elapsed time away
  print('Time away: ' .. elapsed, 10, 24, 1)
end
```

### DIFF
--- a/src/jsapi.c
+++ b/src/jsapi.c
@@ -591,6 +591,15 @@ static duk_ret_t duk_time(duk_context* duk)
 	return 1;
 }
 
+static duk_ret_t duk_date(duk_context* duk)
+{
+	tic_mem* memory = (tic_mem*)getDukMachine(duk);
+
+	duk_push_number(duk, memory->api.date(memory));
+
+	return 1;
+}
+
 static duk_ret_t duk_exit(duk_context* duk)
 {
 	tic_machine* machine = getDukMachine(duk);
@@ -809,6 +818,7 @@ static const struct{duk_c_function func; s32 params;} ApiFunc[] =
 	{duk_trace, 2},
 	{duk_pmem, 2},
 	{duk_time, 0},
+	{duk_date, 0},
 	{duk_exit, 0},
 	{duk_font, 8},
 	{duk_mouse, 0},
--- a/src/jsapi.c
+++ b/src/jsapi.c
@@ -591,11 +591,11 @@ static duk_ret_t duk_time(duk_context* duk)
 	return 1;
 }
 
-static duk_ret_t duk_date(duk_context* duk)
+static duk_ret_t duk_timestamp(duk_context* duk)
 {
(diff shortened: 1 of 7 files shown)


RobLoach:
Thanks! Updated at https://github.com/nesbox/TIC-80/wiki/timestamp

StinkerB06:
Also add a function to convert Unix time values to year/month/week/day/hour/minute/second and vice versa.

RobLoach:
May be easiest to use a Luafunction for that for now. There are so many use cases that could mess it up, timezones, leap years. Offloading outside of the tic80 API for now may be best. Here's a quick script I found after some googling....
``` lua
local function get_date_from_unix(unix_time)
    local day_count, year, days, month = function(yr) return (yr % 4 == 0 and (yr % 100 ~= 0 or yr % 400 == 0)) and 366 or 365 end, 1970, math.ceil(unix_time/86400)

    while days >= day_count(year) do
        days = days - day_count(year) year = year + 1
    end
    local tab_overflow = function(seed, table) for i = 1, #table do if seed - table[i] <= 0 then return i, seed end seed = seed - table[i] end end
    month, days = tab_overflow(days, {31,(day_count(year) == 366 and 29 or 28),31,30,31,30,31,31,30,31,30,31})
    local hours, minutes, seconds = math.floor(unix_time / 3600 % 24), math.floor(unix_time / 60 % 60), math.floor(unix_time % 60)
    local period = hours > 12 and "pm" or "am"
    hours = hours > 12 and hours - 12 or hours == 0 and 12 or hours
    return string.format("%d/%d/%04d %02d:%02d:%02d %s", days, month, year, hours, minutes, seconds, period)
end
```

nesbox:
I renamed `timestamp()` to `tstamp()` here cf49079
Just to make it shorter

PR #1048, opened 2020-04-15
Add libretro Makefile

gouchi:
- We will maintain it if needed.


RobLoach:
Thanks for the addition! Most of the properties are now handled in the CMakeList file:
https://github.com/nesbox/TIC-80/blob/master/CMakeLists.txt#L440-L483

But it's good to have this when cmake isn't available. Thanks!

gouchi:
- I will try thank you.

PR #1057, opened 2020-04-20
Add big-endian support

phcoder:


include/tic80.h
```diff
@@ -78,10 +96,17 @@ typedef union
 {
 	struct
 	{
+#if RETRO_IS_BIG_ENDIAN
+		tic80_gamepad fourth;
+		tic80_gamepad third;
+		tic80_gamepad second;
+		tic80_gamepad first;
+#else
```
RobLoach, line 104: This looks like a lot of maintenance overhead. Is there a way to achieve big endian support without having to flip all the struct data?

nesbox: It scares me too :)
Really, is there an easier way?

phcoder: The problem stems from the abuse of union. The other way is to kill either part of the union which means either:
1) kill "u32 data" and have functions that get it from the components.
2) kill the struct and have functions to manipulate bits

twinaphex:
Hi there, could the merge conflict be fixed and then this merged?

This could really help us out on big endian game consoles and even PowerPC Macs.

nesbox:
Won't be merged sorry, a lot of overhead for no reason.

PR #1064, opened 2020-04-29
Add sustain to notes and effects on frame change

### LINKED ISSUE #382: Music editor: sustaining notes through frames
Hello, dev/s.

Using and testing the music editor I have noticed that empty patterns don't work for keeping notes sustained in a channel, as they obligatorily cut off when switching frames. Is this intentional? I have had to use cloned instruments with plain volume envelopes to simulate sustaining which in fact consists of note retriggering in disguise.

I'm suggesting being able to sustain a note from a frame to another, using an empty pattern (different than zero).

Seeing how the patterns are limited (60 patterns when 1 song out of 8 alone can take up to 20 or more, especially when using more to mimic sustains), and I would assume that so are the instrument slots, I assume that you're going for a policy of making the best use out of a reduced quantity of resources; **this is only a suggestion** that I'm bringing up because I think it would help save space and allow for more possibilities in the music editing department.

On that note, I also wonder if it would be much trouble to add the feature to place volume changes on notes already playing without needing to retrigger them.

I hope this helps improve the program if you're willing to include it. This is also the very first fantasy computer I've ever used and except for that one detail, I'm quite pleased with it. Keep up the good work!

Regards,
-kyusawamura/mihaelkyeah

stefandevai:
This fixes #382.

Channels were being reseted each new frame, that caused the notes/effects to stop. Maybe they were being reseted for another specific purpose that I'm unaware of? Now channels are only reseted if the `music_state` is `tic_music_stop`.

EDIT:
Added a cartridge that makes it easier to test the fixed issue:
[test-sustain.tic.zip](https://github.com/nesbox/TIC-80/files/4549896/test-sustain.tic.zip)


StinkerB06:
But there's at least a music track or so made on TIC-80 using the note-cutting upon frame transitions.

stefandevai:
> But there's at least a music track or so made on TIC-80 using the note-cutting upon frame transitions.

Yeah, that's [Hyrum's Law](https://www.hyrumslaw.com/).

In my opinion, I find it more consistent when the note gets sustained, so that each frame is just the continuation of a track, not a break point. What do you think is better? :)

nesbox:
We have a lot of tracks previously created without the note sustaining and we have to keep compatibility with them.
So, as a workaround, we could add a `sustain` flag to the RAM (somewhere here `0x13FFC` https://github.com/nesbox/TIC-80/wiki/RAM#music-position) to control this behavior.

stefandevai:
> So, as a workaround, we could add a `sustain` flag to the RAM (somewhere here `0x13FFC` https://github.com/nesbox/TIC-80/wiki/RAM#music-position) to control this behavior.

Sounds good to me, I'll try to implement that :)

nesbox:
Great, also we need to add a `sustain` param to the `music(track, frame, row, loop, sustain=false)` API, as well as UI button in the Music Editor to enable the mode.

stefandevai:
Maybe we should add a settings popup or second screen with this mode, follow mode, and future modes we may add? In this way, the tracker's UI remains clean and only with the most used tools. I don't think people will be constantly toggling follow mode or sustain mode. Plus, a cluttered UI usually turns people off. What do you think?

nesbox:
We can add a popup or second screen with the options, but the easiest way is to add a button to the toolbar to enable sustain mode, I think. There is a lot of space :)
image

stefandevai:
That's right, right now there aren't many config options to justify a window. Well, if you add more features in the future you can consider the config window :)

src/squirrelapi.c
```diff
@@ -733,11 +734,17 @@ static SQInteger squirrel_music(HSQUIRRELVM vm)
                     SQBool b = SQFalse;
                     sq_getbool(vm, 5, &b);
                     loop = (b != SQFalse);
+                    if(top >= 6)
+                    {
+                        SQBool b = SQFalse;
+                        sq_getbool(vm, 5, &b);
```
nesbox, line 740: pls use `sq_getbool(vm, 6, &b);` here

src/tic.h
```diff
@@ -299,6 +299,7 @@ typedef struct
     struct
     {
         u8 music_loop:1;
+        u8 music_sustain:1;
         u8 music_state:2; // enum tic_music_state
         u8 unknown:5;
     } flag;
```
nesbox, line 305: Please, move `u8 music_sustain:1;` after `u8 music_state:2;` because it breaks compatibility.
``` c
    struct
    {
        u8 music_loop:1;
        u8 music_state:2; // enum tic_music_state
        u8 music_sustain:1;
        u8 unknown:4;
    } flag;
```

stefandevai:
> awesome, could you pls fix two points I mentioned in the comments?
> thank you

Right! Thanks :) I also want to make some tests with all the APIs to ensure nothing went wrong...

src/tic.h
nesbox: Also, pls use `u8 unknown:4;` because you don't fit 1 byte :)

nesbox: I added a static assertion here 9dace07 to check `tic_sound_state` size

stefandevai: Okay, done :)

stefandevai:
How do you deal with the API reference in the wiki? Should we update it now or after a release?

nesbox:
Yes, I forgot it, we have to update the wiki with `0.80` badge

stefandevai:
Ok. I'll do that then :)

PR #1072, opened 2020-05-02
Release drag even if mouse is out of canvas

### LINKED ISSUE #1071: Sprite editor select tool jumbles the selection content when moved
Compiled from 92ced62f on ubuntu

Selecting beyond the frame of the sprite editor causes the selected content to get corrupted when moved around with the arrow keys.

### Steps to reproduce : 
- open the sprite editor and center on a drawn part of the spritesheet
- use the select tool and drag the selection outside of the canvas (leave some space to be able to move), release the mouse button but keep the mouse out of canvas
- move the selection around with an arrow key -> content gets corrupted

bug_select

### What's happening :
Keeping the mouse out of canvas prevents the release of the drag and the call to `copySelection` (set a breakpoint there to see it's not reached in that case). Therefore, when the selection is moved, it considers the pending selected area, but not the corresponding data.

ddelemeny:
Fixes #1071

### DIFF
--- a/src/sprite.c
+++ b/src/sprite.c
@@ -265,11 +265,10 @@ static void processSelectCanvasMouse(Sprite* sprite, s32 x, s32 y)
                 sprite->select.rect = (tic_rect){sprite->select.start.x, sprite->select.start.y, 1, 1};
             }
         }
-        else if(sprite->select.drag)
-        {
-            copySelection(sprite);
-            sprite->select.drag = false;
-        }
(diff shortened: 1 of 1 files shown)


nesbox:
I tested your pull request, but unfortunately, it doesn't fix the problem, I see the corruption of a sprite area, so, I made a quick fix here 38d4ed7.
Thank you for pointing me this bug.

PR #1076, opened 2020-05-05
Highlight matching delimiters in code editor

### LINKED ISSUE #1075: Paren matching in code editor
When I'm using TIC without an external editor, it's great, but one difficulty I have is with matching parentheses up. In other editors sometimes when you put the cursor on a paren, it colors the matching paren on the other side in a way that helps you know which correspond to which.

Do you think this would be a good addition in the TIC code editor? I could give it a shot if so.

technomancy:
When you're in the code editor, this will make it so having your cursor over a `(` character will highlight the matching `)` character. (or similarly with `[`/`]` and `{`/`}`). This makes it much easier to avoid syntax errors.

The highlighting uses `rectb` around the matching delimiter with the same color as the cursor, so it is visually similar to the cursor but more subtle.

I could also add a config setting to allow you to disable this feature, but I haven't done so yet; do you think this is a good idea?

Fixes #1075.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -151,8 +151,8 @@ static void drawCode(Code* code, bool withCursor)
 
         if(code->matchedDelim == pointer)
         {
-            matchedDelim.x = x, matchedDelim.y = y,
-                matchedDelim.symbol = symbol, matchedDelim.color = *colorPointer;
+            matchedDelim.x = x, matchedDelim.y = y, matchedDelim.symbol = symbol,
+                matchedDelim.color = theme->colors[*syntaxPointer];
         }
 
         if(symbol == '\n')
@@ -219,9 +219,13 @@ static void removeInvalidChars(char* code)
     for(s = d = code; (*d = *s); d += (*s++ != '\r'));
 }
 
-char* findMatchedDelim(const char* start, char* current)
+char* findMatchedDelim(Code* code, char* current)
 {
-    // TODO: return immediately if inside a string or comment
+    const char* start = code->src;
+    // delimiters inside comments and strings don't get to be matched!
+    if(code->syntax[current - start] == SyntaxTypeComment ||
+       code->syntax[current - start] == SyntaxTypeString) return 0;
+
     char initial = *current;
     char seeking = 0;
     s8 dir = (initial == '(' || initial == '[' || initial == '{') ? 1 : -1;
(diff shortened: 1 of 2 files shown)


src/code.c
```diff
@@ -200,11 +219,45 @@ static void removeInvalidChars(char* code)
     for(s = d = code; (*d = *s); d += (*s++ != '\r'));
 }
 
+char* findMatchedDelim(Code* code, char* current)
```
nesbox, line 222: pls use `const` qualifier for `char*`

src/code.c
```diff
@@ -200,11 +219,45 @@ static void removeInvalidChars(char* code)
...
+    while(*current && (start < current))
+    {
+        current += dir;
+        // skip over anything inside a comment or string
+        if(code->syntax[current - start] == SyntaxTypeComment ||
+           code->syntax[current - start] == SyntaxTypeString) continue;
+        if(*current == seeking) return current;
+        if(*current == initial) current = findMatchedDelim(code, current);
+    }
+    return 0;
```
nesbox, line 252: pls return `NULL` instead `0`

src/code.c
```diff
@@ -200,11 +219,45 @@ static void removeInvalidChars(char* code)
...
+    s8 dir = (initial == '(' || initial == '[' || initial == '{') ? 1 : -1;
+    switch (initial)
+    {
+    case '(': seeking = ')'; break;
+    case ')': seeking = '('; break;
+    case '[': seeking = ']'; break;
+    case ']': seeking = '['; break;
+    case '{': seeking = '}'; break;
+    case '}': seeking = '{'; break;
+    default: return 0;
```
nesbox, line 240: pls return NULL instead 0

src/code.h
```diff
@@ -98,6 +98,7 @@ struct Code
         s32 index;
     } outline;
 
+    char* matchedDelim;
```
nesbox, line 101: pls use `const` qualifier for `char*`

nesbox:
All works fine, please add config option to disable the matching.
Also, I made some cosmetic changes in your code bfb2538.
Thank you for the PR :)

technomancy:
Thanks! I'll follow up with a config change next.

If anyone is curious to see the results without building, a recording is here: https://p.hagelb.org/match-delim.gif As you can see, it helps even more when coding Fennel.

PR #1077, opened 2020-05-06
Don't match delimiters if config has MATCH_DELIMITERS=false.

technomancy:
This should allow configurability of #1076 as discussed earlier.


nesbox:
I moved `matchDelimiters` option to the `theme.code` section 225b67a

PR #1078, opened 2020-05-06
Accept and run commands from argv.

technomancy:
I was thinking it might be nice to be able to export to HTML from my makefile. I have two commits here to make that happen; the first one makes it so the CLI accepts a `-command foo` flag which gets run as a console command, and then exits. As far as I can tell this works. The second commit attempts to make it so that the console command `export html foo.zip` will use "foo.zip" as its output filename instead of popping up a system save dialog box. This part does not work yet.

Anyway, if you think this is a good idea I can continue to work on this and get the problems ironed out, but if you think this complicates things too much that is fine too; I wanted to ask your opinion before I spent much more time on it.

Thanks!

### DIFF
--- a/src/console.c
+++ b/src/console.c
@@ -3306,15 +3306,27 @@ static bool checkUIScale(Console* console, const char* param, const char* value)
     return done;
 }
 
-static bool checkCommand(Console* console, const char* command)
+static bool checkCommand(Console* console, const char* argument)
 {
     for(s32 i = 0; i < COUNT_OF(AvailableConsoleCommands); i++)
     {
+        char command[16];
+        for(s32 i = 0; i < sizeof(command); i++)
+        {
+            if(argument[i] == ' ' || argument[i] == '\t')
+            {
+                command[i] = 0;
+                break;
+            }
+            command[i] = argument[i];
+        }
+        command[15] = 0;
+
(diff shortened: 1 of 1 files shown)


nesbox:
Wow, I like it. We could develop your idea further and add such behavior for all the export commands which used UI file save dialog, so you could export music from a cart, for example, without opening TIC's console. Please, continue.
Thanks.

RobLoach:
This is great! I would love to see the external command line interface improve a lot more. Keep it up :+1: 

Quick question.... Could we do something like this?
```
tic80 -code mycode.lua -sprites mysprites.gif -cover cover.gif export cart.tic
```

technomancy:
@RobLoach do you mean specifically that it should interpret any arguments after the already-recognized commands as if they were entered in the console? I think that makes sense if @nesbox agrees.

Running commands is pretty easy to implement; changing the `export` command specifically to accept a filename is much more difficult. Because of that I will probably split this up into 2 separate patches.

technomancy:
I think I have this working now just for the part where it accepts args from argv. Allowing the `export` command to take a filename can be a separate PR once we get this figured out, I think.

technomancy:
@nesbox I think this one is ready to go; thanks.

nesbox:
Ok, thank you :)

technomancy:
Sure thing; thanks!

PR #1091, opened 2020-05-18
WIP: Upgrade to Fennel version 0.4

technomancy:
This brings in the latest version of Fennel, which is 0.4.1-dev. The actual release of 0.4.1 is still a few days away, but the remaining issues do not affect TIC-80 at all. I would like to update this PR later so that it includes the final version of 0.4.1 anyway, but I wanted to open this as WIP so that it doesn't get missed for the 0.8.0 release of TIC.

One big improvement in this version is the reporting when there is a compiler error:

nice-error


nesbox:
Hmm, I see the Amethyst Waters game which is written on Fennel doesn't work now
https://stefandevai.itch.io/amethyst-waters
Could you please check it?

technomancy:
Yes, definitely. Sorry for the miscommunication here; "WIP" means "work in progress"; I wanted to post this for discussion even though it still needed more testing. =)

Thanks for catching this bug! Will get a fix in soon.

nesbox:
Ok I reverted it yet here 8f332b9
Sorry for my misstep

technomancy:
No problem; I'll have something ready by the end of the week. Thanks.

PR #1095, opened 2020-05-21
Enable chromakey array on map and textri

### LINKED ISSUE #996: Array of color keys in map/font/textri?
The `spr()` function's `colorkey` parameter can accept a table array of numbers (instead of just a single number), so that we can specify multiple simultaneous transparent color indexes in the same sprite. However, such functionality doesn't exist with the other three functions as mentioned in the title.

ddelemeny:
Partially fixes #996

### DIFF
--- a/src/jsapi.c
+++ b/src/jsapi.c
@@ -709,7 +709,6 @@ static duk_ret_t duk_textri(duk_context* duk)
         pt[i] = (float)duk_to_number(duk, i);
     tic_mem* tic = (tic_mem*)getDukMachine(duk);
     bool use_map = duk_opt_boolean(duk, 12, false);
-    u8 chroma = duk_opt_int(duk, 13, 0xff);
 
     static u8 colors[TIC_PALETTE_SIZE];
     s32 count = 0;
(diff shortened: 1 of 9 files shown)


RobLoach:
Looks like the other languages are not yet implemented. Seems like a neat idea, but I'm not quite sure how handy it will be. Most sprite sheets almost have one chromakey. Will this have an impact on performance?

ddelemeny:
@RobLoach Thanks for your answer. This should have negligible impact on performance as it leverages
- for `map`, the fact that `drawTile` uses a cached LUT (`mapping`) to handle both palette remapping and multiple chromakey 
- for `textri`, the same thing implemented here : https://github.com/nesbox/TIC-80/commit/22360509cf50209b71858c22ad2b593a9449906c#diff-eb4f22e725ac78ffcd0bab3f665147bbR1063. (Prior to this commit, the behavior was to call `mapColor` on every pixel through the use of `setPixel`, which was arguably not optimized)

The only difference in perf should come from parameter handling for each language bindings, which should be small compared to actually drawing pixels.

As for the utility of it. I'm looking for consistency in the API, and I believe that if it wasn't that useful, `spr` wouldn't have it in the first place.

Which languages are not implemented ?

ddelemeny:
Thanks, I've updated the wiki accordingly.

PR #1101, opened 2020-05-27
Allow "export html foo.zip" to provide name non-interactively.

technomancy:
I have it set right now to not overwrite any existing file with that name.

Currently it saves into the TIC data directory rather than the current
directory; if you pass it an absolute path it fails; I believe the
path should be interpreted as relative to the current directory, but
I'm not sure how to accomplish this.

### DIFF
--- a/src/console.c
+++ b/src/console.c
@@ -1983,7 +1983,7 @@ static void onConsoleExportNativeCommand(Console* console, const char* cartName)
 
 #include "zip.h"
 
-static void onConsoleExportHtmlCommand(Console* console);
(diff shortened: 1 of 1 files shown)


technomancy:
Thanks for merging this. Do you have any thoughts on my question? Right now this only saves into the TIC data folder in `~/.local/share/com.nesbox/TIC-80`, but I think it would be better if it saved in the current directory instead.

nesbox:
@technomancy yes, saving to a working folder would be more predictable.
I think I'll redesign and clean up command line parameters handling in the next version and will try to make TIC work more like a compiler.

PR #1107, opened 2020-05-30
Fix bug in bpp demo

ddelemeny:
It took me a while to realize that the following gif had a problem :
ezgif com-gif-maker

Spoiler : the spritesheet portion displayed isn't always the good one

PR : Pass the right segment value to rFontSheet


nesbox:
thanks

PR #1108, opened 2020-05-31
Oops. flipmask for font should be nonzero

ddelemeny:
Fix a bug introduced on the change of spec for blit_segment.

Note to self : don't skip testing on other carts.


ddelemeny:
Last commit restores pre-refactoring spacing sizes for print, in order to preserve compatibility.

Thanks @StinkerB06 for the feedback!

nesbox:
I thought, maybe we could create some kind of unit testing for all the api functions to not break compatibility during implementing new features.

ddelemeny:
I was thinking about this. We could have some reference test carts run automatically.

PR #1121, opened 2020-06-19
Add pixel color format field to TIC-80 structs and blitting.

asiekierka:
Allows for natively outputting to the buffer in a format different than
the default SDL/Sokol one. This also means that the libretro port now
avoids an additional conversion step.

More or less necessary for the 3DS port.

### DIFF
--- a/src/studio.c
+++ b/src/studio.c
@@ -1417,7 +1417,7 @@ static void setCoverImage()
     {
         enum {Pitch = TIC80_FULLWIDTH*sizeof(u32)};
 
-        tic_core_blit(tic);
+        tic_core_blit(tic, TIC_PIXEL_COLOR_RGBA8888);
 
         u32* buffer = malloc(TIC80_WIDTH * TIC80_HEIGHT * sizeof(u32));
 
@@ -1661,7 +1661,7 @@ static void drawRecordLabel(u32* frame, s32 sx, s32 sy, const u32* color)
     }
 }
 
-static void drawDesyncLabel(u32* frame)
+static void drawDesyncLabel(u32* frame, tic_pixel_color_format fmt)
 {
     if(getConfig()->showSync && impl.missedFrame)
     {
@@ -1676,7 +1676,7 @@ static void drawDesyncLabel(u32* frame)
         
         enum{sx = TIC80_WIDTH-24, sy = 8, Cols = sizeof DesyncLabel[0]*BITS_IN_BYTE, Rows = COUNT_OF(DesyncLabel)};
 
-        const u32* pal = tic_tool_palette_blit(&impl.config->cart.bank0.palette);
+        const u32* pal = tic_tool_palette_blit(&impl.config->cart.bank0.palette, fmt);
         const u32* color = &pal[tic_color_6];
 
         for(s32 y = 0; y < Rows; y++)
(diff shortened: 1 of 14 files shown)


src/studio.c
```diff
@@ -1945,12 +1950,18 @@ static void studioTick()
             memcpy(tic->ram.font.data, impl.systemFont.data, sizeof(tic_font));
         }
 
-        data
-            ? tic_core_blit_ex(tic, scanline, overline, data)
-            : tic_core_blit(tic);
+        if(isRecordFrame())
```
asiekierka, line 1953: Should we be handling the edge case where the frame record format is equal to the blitting format? On modern computers I don't think it makes much of a difference, but...

src/tic.h
```diff
@@ -611,3 +611,12 @@ typedef enum
...
 } tic_cursor;
+
+typedef enum {
+    TIC_PIXEL_COLOR_ARGB8888 = (1 << 8) | 32,
+    TIC_PIXEL_COLOR_ABGR8888 = (2 << 8) | 32,
+    TIC_PIXEL_COLOR_RGBA8888 = (3 << 8) | 32,
+    TIC_PIXEL_COLOR_BGRA8888 = (4 << 8) | 32
+} tic_pixel_color_format;
+
+#define TIC_PIXEL_COLOR_DEPTH(fmt) ((fmt) & 0xFF)
```
asiekierka, line 622: This is a macro for future expansion, if we ever want to support non-32-bit depths (such as RGB888, or RGB565).

RobLoach:
Happy to see this cleaned up! Thanks a lot.

src/system/baremetalpi/kernel.cpp
```diff
@@ -181,25 +181,14 @@ static System systemInterface =
...
 {
 	u32 pitch = screen->GetPitch();
 	u32* buf = screen->GetBuffer();
 	for (int y = 0; y<136;y++)
-	for (int x = 0; x<240;x++)
 	{
-		u32 p = ts[(y+4)*(8+240+8)+(x+8)];
-		buf[pitch*y + x] = rgbaToBgra(p);
+		u32 *line = ts + ((y+TIC80_OFFSET_TOP)*(TIC80_FULLWIDTH) + TIC80_OFFSET_LEFT);
+		memcpy(buf + (pitch * y), line, 240 * 4);
```
nesbox, line 191: pls use `TIC80_WIDTH` `TIC80_HEIGHT` instead `240` `136` integers here

asiekierka: Done. Are there any build instructions for baremetalpi? Not sure how to check if it still compiles...

src/studio.c
nesbox: I think we don't have to handle the edge case, it looks more clear atm

asiekierka: OK

src/system/baremetalpi/kernel.cpp
nesbox: Here are the build instructions https://github.com/msx80/BareMetalTic80#how-to-build.
Also, we are going to add the baremetal build to the CI server and it will be checked automatically.

nesbox:
@asiekierka 
Pls don't forget to add yourself to the contributors list here https://github.com/nesbox/TIC-80/wiki#credits
Thanks

PR #1123, opened 2020-06-27
Change keybindings in the code editor to resemble bash a bit more.

Kozova1:
What Changed:
CTRL-A is now the same as HOME
CTRL-E is now the same as END
CTRL-SHIFT-A has the old behavior of CTRL-A, where the whole code is selected.

These changes were made because bash and emacs both use CTRL-A to go to the start of the line, and CTRL-E to go to the end of it. I moved CTRL-A to CTRL-SHIFT-A to still keep _some_ backwards compatibility

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -1239,7 +1239,7 @@ static void processKeyboard(Code* code)
         if(keyWasPressed(tic_key_left))             leftWord(code);
         else if(keyWasPressed(tic_key_right))       rightWord(code);
         else if(keyWasPressed(tic_key_tab))         doTab(code, shift, ctrl);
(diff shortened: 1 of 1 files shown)


Kozova1:
by the way I just verified and it builds just fine and it works just as intended in the editor.

Kozova1:
Also, if this gets merged, don't forget to update the shortcuts section in the wiki.

nesbox:
Hi, thank you for the PR, I think all these keybindings are a matter of habit, as for me `CTRL+A` to select all the text is more convenient because I'm a Windows user. We could add keyboard bindings data to the config to let all set up their own.
I'll create an issue.

nesbox:
will be here #1125

PR #1161, opened 2020-09-05
Update to Fennel version 0.6.0.

technomancy:
Rename modified compiler file to loadfennel to make it clearer that
it's not the exact same thing as the upstream module. (It has a shim
in it to populate package.loaded because it will not be loaded
normally thru require.)

Remove mention of fennelfriend module; it is now integrated with the
main compiler.

Add some keywords to syntax highlighting.

This version of Fennel contains several bug fixes from the old one (0.4.1) you can see here: https://fennel-lang.org/changelog


nesbox (approved):
Thank you!

technomancy:
@nesbox I have found a somewhat problematic bug in Fennel 0.6.0, so I would like to get 0.6.1 released and incorporated into TIC before the next release. I saw that the beta for TIC 0.8.0 got published not too long ago. Do you have a rough estimate for when 0.8.0 will be released? I want to make sure I get my fix in before that happens. Just wondering if it's likely to be a few days, a few weeks, or over a month. Thanks.

nesbox:
I'd like to release it in 1-2 weeks.

technomancy:
Thanks, will try to get a PR with a fix by the end of the week.

PR #1165, opened 2020-09-07
fixes #1163 (0.80 beta) time() always returns 0 when built with sokol

### LINKED ISSUE #1163: (0.80 beta) time() always returns 0 when built with sokol
I remember this has also happened before 0.80 beta, but the function `time` always returns 0 when using `tic80-sokol`.

The easiest way to see this is when running the `benchmark.tic` cart. When running one of the benchmarks they will never finish, because they aren't detecting any time has taken place. This seems like a fairly important bug to me.

The main reason I use the sokol build is because it performs better and the audio isn't choppy on my computer. It looks like the audio issues for the main build are being looked into for the 0.90 version, though.

drako0812:
This allows the function `time` to return valid values by implementing `getPerformanceCounter` and `getPerformanceFrequency`:
https://github.com/drako0812/TIC-80/blob/42f685cc3d0b06036a756017644f48c68462ebd1/src/system/sokol.c#L79-L87

I don't like how I had to return a raw value for `getPerformanceFrequency`, but `sokol_time` doesn't provide a convenient way to query this information.

It seems to work just fine, but I'm not 100% sure I put the call to `stm_setup` in the right spots.

Currently `stm_setup` is called:
here
https://github.com/drako0812/TIC-80/blob/42f685cc3d0b06036a756017644f48c68462ebd1/src/system/sokol.c#L167
and here
https://github.com/drako0812/TIC-80/blob/42f685cc3d0b06036a756017644f48c68462ebd1/src/player/sokol.c#L45

Also the modifications to `.gitignore` could be reverted, those were just to prevent staging various files creating by the way I went about building TIC-80.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -156,3 +156,11 @@ build/windows/tic80.rc
 *.ninja
 tic80.sublime-workspace
 build/*.tic
+build/*.vcxproj
+build/*.vcxproj.filters
+build/*.vcxproj.user
+build/CMakeDoxyfile.in
+build/*.sln
+build/.local/
+build/*.dir/
+build/x64/
--- a/src/player/sokol.c
+++ b/src/player/sokol.c
@@ -41,6 +41,8 @@ static void app_init(void)
     }
 
     sokol_gfx_init(TIC80_FULLWIDTH, TIC80_FULLHEIGHT, 1, 1, false, false);
+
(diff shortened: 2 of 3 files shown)


nesbox (approved):
Looks good.
Thank you.

PR #1174, opened 2020-09-13
update editor upon processing backspace

### LINKED ISSUE #1167: Columns do not update on backspace
If you backspace, the column value does not change. Only if you add something to the current line, it updates.

msarilar:
update editor upon backspace and delete

should fix #1167


nesbox (approved):
Looks good, thank you.

PR #1175, opened 2020-09-14
Update Wren to 0.3.0

RobLoach:
This change updates to [Wren 0.3.0](https://github.com/wren-lang/wren/releases/tag/0.3.0). Not many changes, but good to target a tagged release rather than the commit hash we had before.


nesbox (approved):
thank you

PR #1176, opened 2020-09-14
SDL: Render in the correct aspect ratio

RobLoach:
This change makes two changes...

1. Starts the window at 3x the original size so that it doesn't appear so tiny
2. Makes it so that when the user resizes it, it ends up rendering in the correct aspect ratio


src/player/sdl.c
```diff
@@ -57,17 +57,17 @@ int main(int argc, char **argv)
 			SDL_Init(SDL_INIT_VIDEO | SDL_INIT_AUDIO);
 
 			{
-				SDL_Window* window = SDL_CreateWindow("TIC-80 SDL demo", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, TIC80_FULLWIDTH, TIC80_FULLHEIGHT, SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE);
+				SDL_Window* window = SDL_CreateWindow("TIC-80 SDL demo", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, TIC80_FULLWIDTH * 3, TIC80_FULLHEIGHT * 3, SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE);
```
nesbox, line 60: Could you pls use a constant or compiler definition instead `3` here?
Thank you :)

nesbox (approved):
thank you 👍

PR #1217, opened 2020-10-24
n3ds: fix crash in async net_httpc access, close #1200

### LINKED ISSUE #1200: [3DS v0.80.1344] Crash on Startup
Hello again, I finally got the 3DS version of TIC-80 built, but I have a bug to report.

On an o3DS, TIC-80 will crash upon startup, right around when the console is meant to appear. A crash dump can be found here
[crash_dump_00000000.zip](https://github.com/nesbox/TIC-80/files/5313749/crash_dump_00000000.zip)

EDIT: This happens on both the CIA and 3DSX versions.

asiekierka:


Kikasuru:
Wait, why is .CIA building support getting removed?

asiekierka:
There are some reasons:

* .3DSX is the de-facto standard; while there used to be certain things you could not easily accomplish with a .3DSX that you could with a .CIA in the past, that is no longer the case; 
* The software typically used to install .CIAs is also commonly used for activites I strongly do not wish to see associated with the homebrew community at large,
* There are rumors of existing .CIA building tools relying on information gathered from the official SDK, which would, again, be something I do not wish to be associated with the way homebrew is usually made.

nesbox (approved):
Thank you!

PR #1218, opened 2020-10-24
libretro: Add Mouse Hide Delay option

RobLoach:
This adds a Mouse Cursor Hide Delay option, to configure the number of seconds to wait before the mouse is hidden.


nesbox (approved):
Thank you!

PR #1221, opened 2020-10-26
Update Lua to 5.3.6.

asiekierka:
The release on GitHub is not tagged as 5.3.6, but the code (as diffed) is identical.

As Lua 5.3 is unlikely to see further maintenance, might be a good idea to consider adding a separate "lua54" language for Lua 5.4.x...


nesbox (approved):
Thank you!

PR #1222, opened 2020-10-26
n3ds: add prescale linear filter, wide mode support; close #1219

### LINKED ISSUE #1219: [3DS] Sharp linear (prescale) filter
Hello! I'm hoping it's possible to add a prescale 2x or 3x screen filtering option to the 3ds port. Right now there seems to be: pixel perfect, nearest neighbor, and linear scaling. 

Something similar to this would be ideal:
https://retropie.org.uk/forum/topic/12387/a-simple-sharp-pixel-shader-config-for-retropie-with-integer-prescale-and-no-pixelwobble

As far as I understand it the original screen is integer scaled 2x or 3x with nearest neighbor scaling. Then at the larger size a linear or bilinear filter is applied. Then the screen is scaled back down to the desired screen resolution (using nearest neighbor).

This would greatly improve readability at full screen. @asiekierka is this a possibility? Thank you for all your work on the 3ds port, it really is amazing!

asiekierka:


nesbox (approved):
Thank you!

nesbox:
Would you like to add a 3DS build to Github Actions here https://github.com/nesbox/TIC-80/actions?
And we will have a build on every commit to the repo.
What do you think? :)

nesbox:
You just need to add build instructions here https://github.com/nesbox/TIC-80/blob/master/.github/workflows/build.yml

nesbox:
Added 3DS build to Github Actions 6d0f880

asiekierka:
Thank you! Sorry for not getting around to it.

nesbox:
Thank you for the PR :)
Could you pls add yourself to the Credits https://github.com/nesbox/TIC-80#credits

PR #1237, opened 2020-11-10
WIP: Adding Python support via micropython.

jminor:
This PR is an attempt to add support for Python in TIC-80 by embedding micropython.


jminor:
This PR is stalled due to this upstream issue: https://github.com/micropython/micropython/pull/5964

If anyone is able to help with that, or find an alternate way to embed Python, I'd appreciate the help :)

ItzzCode:
This seems pretty interesting, ill check this out if it works out.

joshgoebel:
Python would be a great addition for sure.

ItzzCode:
Yes

jminor:
The upstream issue in micropython is resolved now, and I have made significant progress on this PR. I now have a Python demo cart mostly working!

There are two main challenges ahead. One is simply wrapping the rest of the TIC API in micropython. I have been porting each of the `lua_XXX` functions from `lua.c` into `python_XXX` functions in `micropython.c` For example here in the implementation of `python_pix`: https://github.com/jminor/TIC-80/blob/micropython/src/api/micropython.c#L184 
If anyone else wants to lend a hand at porting the rest of the API I would much appreciate it.

The other challenge is to get the cmake system to build micropython properly. To do it manually just do this:
```bash
% cd TIC-80/vendor/micropython/mpy-cross
% make
% cd ../examples/embedding
% make
```
That will build `libmicropython.a` which then will link into TIC-80.

I attempted to put some `add_custom_command` or `execute_process` into `CMakeLists.txt` to do this automatically. It works fine now on my machine (macOS 11 Big Sur), but does not work in GitHub Actions. I will definitely need some help getting that part working.

jminor:
Sadly, I don't think I will have the time to come back to this PR in the future. If someone else wants to take it on, or try a different approach (maybe via CPython instead of micropython?) feel free to try.

AlecTroemel:
just saw [this project called pocketpy](https://github.com/blueloveTH/pocketpy), I wonder if it could be another option for whoever trys at this again

jminor:
Oooh, that looks much easier to integrate!

PR #1241, opened 2020-11-14
README.md: Linux Ubuntu build instructions -> missing sound

sdesalas:
Sound missing in Ubuntu when following build instructions in README.md


sdesalas:
Ah, this was for 18.04 not 14.04

PR #1256, opened 2020-11-27
bare metal rpi build updated to latest TIC-80 and Circle

msx80:
See issue #912
I patched up the build as best as i could, it works now. There's still some issue on the filesystem access (biggest thing is that "cd" command doesn't work) but it boots, carts can be loaded, played and saved.

I've updated the build procedure [here](https://github.com/nesbox/TIC-80/blob/master/build/baremetalpi/README.md).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,55 @@ set(VERSION_MINOR 90)
 set(VERSION_REVISION 0)
 set(VERSION_STATUS "-dev")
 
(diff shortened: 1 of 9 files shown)


nesbox (approved):
Thank you :)

nesbox:
Could you please add yourself to the Credits https://github.com/nesbox/TIC-80#credits?
Thank you :)

PR #1274, opened 2020-12-14
gitignore: Ignore compile_commands.json

RobLoach:
After building, compile_commands.json ends up showing in git status. This change ignores that file.


nesbox (approved):
Thanks

PR #1278, opened 2020-12-19
Allow multiple screenshots and videos to be saved

### LINKED ISSUE #1275: Please add back ability to name gifs
In the latest TIC-80 version the app saves gifs under name video.gif. Which you can't change and the gif will be overwritten without warning. I suggest to add at least numbering like `video0.gif` `video1.gif` `video2.gif` etc. is the file `videoN.gif` exists

RobLoach:
This saves `video1.gif` or `screen1.gif`, and increments the index when one is already saved.

Fixes #1275


nesbox (approved):
Thanks a lot :)

PR #1318, opened 2021-01-28
Make RAM and VRAM a bit more explicit

joshgoebel:
Changes in the interest of clarity and being a bit more explicit.  This follows from the work I already did on the Wiki.

- "FONT" => "SYSTEM FONT"
- K => KB
- Adds VRAM to the RAM map (seems important and that VRAM exists inside of RAM)
- Adds "reserved" and "free" comments for clarity
- SIZE => BYTES
- "BORDER" => "BORDER COLOR"


nesbox (approved):
OK, let's try this
Thanks.

Issue #559, opened 2018-03-01
mouse() returns x,y outside 0,0~239,135 when mouse pointer is on screen border


lolbot-iichan:
TEST CASE 1 (bf):
1. run http://lolbot-iichan.github.io/TIC-80/bin/html/bf_mouse2.html
2. move mouse pointer from game screen center to the left
EXPECTED RESULT: x goes down to 0 and stops
ACTUAL RESULT: x goes 5,4,3,2,1,0,255,254,253,252,251,250,249,248

TEST CASE 2 (lua):
1. build and run web version of TIC-80
2. type "new lua"
3. insert "trace(mouse())" line to main TIC function
4. move mouse pointer from game screen center to the left
5. press ESC and read traces
EXPECTED RESULT: x goes down to 0 and stops
ACTUAL RESULT: x goes 5,4,3,2,1,0,255,254,253,252,251,250,249,248

Compare with https://tic.computer/play?cart=24

AnastasiaDunbar:
When cursor goes further into the border's top left corner, `mouse()`'s first two elements will return 255.
To get around this you have to do this:
```js
var m=mouse();
if(m[0]===255){m[0]=0;}
if(m[1]===255){m[1]=0;}
```

AnastasiaDunbar:
![a](https://user-images.githubusercontent.com/11806048/44622921-d9714100-a8c2-11e8-8aec-22958c16e971.gif)

nesbox:
Fixed here 8fab61c
We save in ram mouse pos of full `255x144` screen with border, while `mouse()` API returns pos of `240x136` screen.
![mouse](https://user-images.githubusercontent.com/1101448/97574543-6f9dd600-19fc-11eb-8b67-ca25cd49c54c.gif)

nalquas:
It seems like TIC-80 doesn't adjust for changes in border size due to window resizing, so this issue can still (kind of) appear when the window has been resized (or put in fullscreen).

### Description
- When the TIC-80 window is in its original size, moving the mouse towards the left edge of the window results in `mouse()` returning an x value of ...,3,2,1,0 (which is correct behaviour).
- When the TIC-80 window is scaled (e.g. by extending the window to the right or by going into fullscreen), moving the mouse towards the left edge of the window results in `mouse()` returning x values of ...,3,2,1,0,255,254,... (which is incorrect behaviour)

[...]

PR #1322, opened 2021-01-30
Clip mouse values in SDL and Sokol

nalquas:
As I noted in issue https://github.com/nesbox/TIC-80/issues/559#issuecomment-770119463, mouse values wrapped around when moving the mouse towards the edges of the screen, especially in fullscreen.

The problem is that the mouse's x/y values are stored as 8bit after being written in RAM.

One possible fix is to clip the values _before_ writing them into RAM, which is what I did here. The mouse's x values are now clipped to a range of 0 to TIC80_FULLWIDTH-1 (and 0 to TIC80_FULLHEIGHT-1 for y) before written into ram.

This is basically identical to the clipping/checking already done in n3ds and in libretro:
https://github.com/nesbox/TIC-80/blob/acddaaad3ae5f5f403db9f1a69872697fb49efb5/src/system/n3ds/main.c#L497-L508
https://github.com/nesbox/TIC-80/blob/acddaaad3ae5f5f403db9f1a69872697fb49efb5/src/system/libretro/tic80_libretro.c#L480-L492

What this means for the mouse api is that `mouse()` now returns x values of -8 to 247 (which is 8px outside of the 0-239 range) and y values of -4 to 139 (which is 4px outside of the 0-135 range).

### DIFF
--- a/src/system/sdl/main.c
+++ b/src/system/sdl/main.c
@@ -524,14 +524,30 @@ static void processMouse()
 #if defined(CRT_SHADER_SUPPORT)
         if(crtMonitorEnabled())
         {
-            if(rect.w) input->mouse.x = (mx - rect.x) * TIC80_FULLWIDTH / rect.w;
-            if(rect.h) input->mouse.y = (my - rect.y) * TIC80_FULLHEIGHT / rect.h;
+            if(rect.w) {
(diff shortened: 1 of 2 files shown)


nesbox (approved):
Looks good, thank you.
One thing, please use `s32` type instead `int`.

Issue #1270, opened 2020-12-10
Floating point string representation in fennel with non english locale


adrian-castravete:
I'll start off by specifying some things that are more than likely relevant.
I'm on linux and my locale settings are the following:
```
$ locale
LANG=en_GB.UTF-8
LANGUAGE=
LC_CTYPE="en_GB.UTF-8"
LC_NUMERIC=ro_RO.UTF-8
LC_TIME=ro_RO.UTF-8
LC_COLLATE="en_GB.UTF-8"
LC_MONETARY=ro_RO.UTF-8
LC_MESSAGES="en_GB.UTF-8"
LC_PAPER=ro_RO.UTF-8
LC_NAME=ro_RO.UTF-8
LC_ADDRESS=ro_RO.UTF-8
LC_TELEPHONE=ro_RO.UTF-8
LC_MEASUREMENT=ro_RO.UTF-8
LC_IDENTIFICATION=ro_RO.UTF-8
LC_ALL=
```

I said this as there's something specific to the numeric representation of real (floating point) numbers. In most cases we all agree that this is a floating point number: `1.234` for the result of 1234 / 1000. The Romanian locale is `1,234` however.

Now i have these two programs that are identical, save for the fact that they're different languages:
```lua
-- title:  game title
-- author: game developer
-- desc:   short description
-- script: lua

function threadFunc(start)
 local test = start
 while test > 0 do
  print(test)
  test = test - coroutine.yield()
 end
end

local thread = coroutine.create(threadFunc)

coroutine.resume(thread, 1 + math.random())

function TIC()
 cls()
 if not coroutine.resume(thread, 0.01) then
  print("Coroutine ended")
 end
end
```
and
```scheme
;; title:  game title
;; author: game developer
;; desc:   short description
;; script: fennel

(fn thread-func [start]
 (var test start)
 (while (> test 0)
  (print test)
  (set test (- test (coroutine.yield)))))

(var thread (coroutine.create thread-func))

(coroutine.resume thread (+ 1 (math.random)))

(global TIC
 (fn tic []
  (cls)
  (when 
   (not (coroutine.resume thread 0.01))
   (print "Coroutine ended"))))
```

The problem is that the `fennel` one just seems to hang whereas the `lua` one continues normally.
So here's the **lua**:
![Lua version](https://i.imgur.com/6IMCYKu.gif)
And here's the **fennel**:
![Fennel version](https://i.imgur.com/2Y8Vphz.gif)

[...]

PR #1336, opened 2021-02-07
Update to Fennel 0.8.1.

technomancy:
This brings in the latest version of Fennel.

It includes a fix for #1270 which was caused by a bug in the compiler: https://todo.sr.ht/~technomancy/fennel/32

I've tested it with most of the TIC games written in Fennel.

Thanks!


nesbox (approved):
Thank you for the fix!

PR #1341, opened 2021-02-11
Show how to build pro version

joshgoebel:
Is this supposed to be well hidden?  Why would someone building from source (possibly contributing to the project, etc) not want the pro version by default?  Could all the build instructions be updated to build pro or explain how?

This was not clear at all and took me like 10 minutes to figure out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ You can download compiled versions for the major operating systems directly from
 To help support TIC-80 development, we have a [PRO Version](https://nesbox.itch.io/tic80).
 This version has a few additional features and binaries can only be downloaded on [our Itch.io page](https://nesbox.itch.io/tic80).
 
-For users who can't spend the money, we made it easy to build the pro version from the source code.
+For users who can't spend the money, we made it easy to build the pro version from the source code. (`cmake .. -DBUILD_PRO=On`)
 
 ### Pro features
 


nesbox (approved):
Ok thank you, also we could add info how to build a release with `-DCMAKE_BUILD_TYPE=MinSizeRel` option. Because we are getting Debug build by default and it works quite slow.

joshgoebel:
>  Because we are getting Debug build by default and it works quite slow.

How can I tell from the final build if it's a debug build or not?  I'm now following the steps in `build.yml` yet still I see `-dev` on my version... I would assume that building a production build should remove this?  

If -dev does not denote that could we add a if/else in `system.h` to add "Debug" to TIC_VERSION_LABEL?

nesbox:
> How can I tell from the final build if it's a debug build or not? I'm now following the steps in `build.yml` yet still I see `-dev` on my version... I would assume that building a production build should remove this?

Yes, the prodution build is on `stable` branch with removed `dev` prefix.

> If -dev does not denote that could we add a if/else in `system.h` to add "Debug" to TIC_VERSION_LABEL?

I think it's  a good idea. Thanks.

joshgoebel:
> could we add a if/else in system.h to add "Debug" to TIC_VERSION_LABEL?

I tried and failed to add this... I must now understand exactly what what a "debug" build means... neither DEBUG or EN_DEBUG seem to be set... i tried both.

PR #1348, opened 2021-02-14
libretro: Update libretro.h and code standards

RobLoach:
Small updates to the libretro header files, along with some small coding standards fixes.


nesbox (approved):
thank you

PR #1355, opened 2021-02-20
Fix using full path as cart name when loading cart

phlubby:
Since a39a206 (File Sytem refactoring, 2020-02-02) using cmdLoadCart()
results in a cart name containing the full path passed instead of only
the basename.

Fix by making it possible for isslash() to return true ("warning C6288:
Incorrect operator: mutual inclusion over && is always zero. Did you
intent [sic] to use || instead?"), and skipping copying of a path
separator.


nesbox (approved):
Thanks

PR #1361, opened 2021-02-26
libretro: Update the memory data slots

RobLoach:
This updates the libretro memory slots that are available within the TIC-80 memory.


nesbox (approved):
Thanks

PR #1368, opened 2021-03-04
Wren TIC.btn fixes

txgruppi:
# The fixes

Two fixes for the Wren's API

1. `TIC.btn()` was not defined in Wren and had the wrong return value
2. `TIC.btn(_)` was mixing controllers 1 & 3, 2 & 4

### `TIC.btn()`

Fixed 2 problems:

1. The handler `wren_btn` was checking for this call but it was not declared as part of the `TIC` class and it was not defined in `foreignTicMethods`
2. The handler `wren_btn` was setting the return value to `bool` instead of `double`

### `TIC.btn(_)`

The button index _bitmask_ was `0xf` instead of `0x1f`, this made tat gamepad 3 returned value for gamepad 1 and gamepad 4 returned value for gamepad 2.

### DIFF
--- a/src/api/wren.c
+++ b/src/api/wren.c
@@ -268,7 +268,7 @@ static void wren_btn(WrenVM* vm)
     }
     else if (top == 2)
     {
-        s32 index = getWrenNumber(vm, 1) & 0xf;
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Looks good.
Thank you.

PR #1369, opened 2021-03-05
Fixed Squirrel's print function fixed flag

txgruppi:
The "fixed" flag was being read from arguments 5, where it actually is in arguments 6.


nesbox (approved):
oops
thanks :)

PR #1386, opened 2021-04-02
Add "code" to the import usage documentation

RobLoach:
Looked like `code` was missing from the import list, so I've opted to add it and alphabetize the list.


nesbox (approved):
Thanks

PR #1393, opened 2021-04-05
libretro: Add features from jdgleaver

RobLoach:
@jdgleaver added a few features to the libretro core. This change cherry-picks those forwards to the upstream.

- Option to remove the screen border
- Gamepad Analog deadzone
- Emulated mouse pointer for devices without a mouse


nesbox (approved):
Awesome!
Thank you for support

PR #1427, opened 2021-04-26
fix dapper builds

cgrs:
This changes are now in sync with the manual build instructions. Building successfully in Ubuntu 21.04 and macOS Catalina 10.15.7 (19H524), though the binary executable it's only valid for Linux.

PS: Maybe it is possible to integrate Github actions with this file.


nesbox (approved):
Thanks

PR #1435, opened 2021-05-01
Fix CRT shader on macOS.

jminor:
On macOS Big Sur, if you enable the CRT shader with F6, it complains
```
Failed to load vertex shader: ERROR: 0:13: ';' : syntax error: syntax error
```

This PR fixes that by removing the extra semicolon after `main() {...}` in the CRT vertex shader.


nesbox (approved):
Thank you

PR #1442, opened 2021-05-03
Fix import command.

jminor:
The import command was broken since the switch to use `split()` and `strtok()` for parsing the command parameters. This PR makes import work again.


nesbox (approved):
Thank you

PR #1443, opened 2021-05-03
Enable CRT shader in wasm/WebGL build

jminor:
This PR enables the CRT shader in the wasm/WebGL build by patching the GLSL shader when it loads. This technique is fragile, and not as clean as I would like. If you have suggestions for a better way to accomplish this, I'm open to suggestions.

You can see it working here: https://pixelverse.org/experiments/natsukashii/


nesbox (approved):
Thank you.
I'm merging it, but going to change your PR a bit, I think it's better to add `-DBUILD_SDLGPU=On` here https://github.com/nesbox/TIC-80/blob/master/.github/workflows/build.yml#L322 instead enabling it directly in the cmake.

nesbox:
Could you pls add yourself to the Credits list https://github.com/nesbox/TIC-80#credits?
Thank you for the help :)

PR #1456, opened 2021-05-29
Fixes for 2 SIGSEGVs when section is NULL

Karethoth:
I ran into these SIGSEGVs when trying to load any cartridge on CentOS Stream 8.
After adding the null checks the cartridges load without a hitch.

#0  0x00007ffff69aa685 in __strlen_avx2 () from /lib64/libc.so.6
#1  0x00007ffff68d6802 in strdup () from /lib64/libc.so.6
#2  0x000000000043bca2 in onLoadCommandConfirmed ()

#0  0x00007ffff69aa685 in __strlen_avx2 () from /lib64/libc.so.6
#1  0x00007ffff68d6802 in strdup () from /lib64/libc.so.6
#2  0x000000000043b8ea in loadByHash ()

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -787,7 +787,7 @@ static void loadByHash(Console* console, const char* name, const char* hash, con
 {
     console->active = false;
 
-    LoadByHashData loadByHashData = { console, strdup(name), strdup(section), callback, data};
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Thank you

PR #1457, opened 2021-05-31
Adding a Progressive Web App build

schraf:
This pull request adds a new build target called webapp that works similar to the existing wasm build but will produce a webpage that is capability of being installed and run offline. The main use case for me was to be able to have TIC-80 work just like a native app on my ipad. 

There is a new github action called webapp.yml that will build and deploy the webapp to a 'webapp' branch every time a release is published. This way you can set the github pages to point to the webapp branch and it will be hosted by github for you. 

One thing I noticed is that on my ipad when I run the this the audio is lagging by multiple seconds. I don't think it is related to these changes and is more likely an emscripten thing.

I also want to do another change to this to make a better approach to importing/exporting projects.

You can test a deployment from my fork: https://schraf.github.io/TIC-80/


ChildishGiant:
The audio lagging on the wasm build is a common issue, not just on iPads. Doubt it's anything to do with the edits you've made. Currently this isn't really usable on a mobile device due to `edit` seeming to no longer being a command

schraf:
Didn't even thing about that. I have been using a hardware keyboard on iPad.

nesbox (approved):
This is awesome.
Please, don't forget to add yourself to the Credits section https://github.com/nesbox/TIC-80#credits.
Thank you for the help.

PR #1464, opened 2021-06-20
Installing and running instructions

Voklen:
This PR adds documentation on installing and running TIC-80 using the Linux build instructions.

The original docs didn't mention running `sudo make install` to install the program.`make -j4` is executed while running `sudo make install -j4`  so the former is redundant and has been removed.The docs also didn't mention how to run the program after installation so those instructions have been added as well.

The `make -j4` in the Mac build instructions could also be replaced with `sudo make install -j4` but I don't have a MacOS machine to test it on so somebody else could try that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ run the following commands in the Terminal
 sudo apt-get install git cmake libgtk-3-dev libgles1-mesa-dev libglu-dev -y
 git clone --recursive https://github.com/nesbox/TIC-80 && cd TIC-80/build
 cmake ..
-sudo make install -j4
+make -j4
 ```
 
 to install the latest CMake:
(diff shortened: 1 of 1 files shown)


DannyCork (approved):

RobLoach:
This is underneath "Build instructions", which would imply it's instructions how to build TIC-80, not install it.

nesbox (changes requested):
Hi, 
pls move `sudo make install` to the `Install` section
Thank you for the help

Voklen:
I created an "Install instructions" section and a "Linux" sub-section to allow contributers to add install instructions for other platforms as well. I also added links to the section from the Linux build instructions.

nesbox (approved):
thank you

PR #1483, opened 2021-07-12
#764: replace tile feature in map editor

### LINKED ISSUE #764: Map editor: replace tile feature
If there's ever a need to move tiles around in the tile editor, it might mess up maps I've already made.

An easy fix for this would be a tool in map editor to replace all tiles with id `x` with a tile of id `y`.

cogk:
In the Map Editor, use **ctrl+click** while in FILL mode to replace a specific tile with another tile across the whole map (or only in the selection).

This *Replace Tile* feature also works with rectangular sprite patterns.
The pattern is offset based on the mouse click position.

https://user-images.githubusercontent.com/10946971/125296665-cea34100-e326-11eb-9d5d-8ca4da13f92e.mp4

Fixes #764

### DIFF
--- a/src/studio/editors/map.c
+++ b/src/studio/editors/map.c
@@ -570,6 +570,26 @@ static void drawTileCursorOvr(Map* map)
     }
 
     drawCursorPos(map, pos.x, pos.y);
+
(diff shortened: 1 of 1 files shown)


RobLoach:
That's so cool!

nesbox (approved):
Awesome, thank you for the help :)

PR #1485, opened 2021-07-12
#1253: Load cart by hash

### LINKED ISSUE #1253: some way to load carts from website by number?
Often I'll see a cart I want to try out on the site, but it doesn't run well in the browser (understandable, my computers not great, and that's basically 2 levels of interpreters). So obvious thing to do would be to open the cart in the desktop version, but how to do it? you can either:
1. download it (now I have a bunch of "cart (N).tic" and I have to open it with a file popup)
2. try to find it using the `surf` command (not likely, and I have to know the right catagory)

It would be nice if I could just put in the cart number (from the url), and have it open, since it can already load remote carts for the `surf` command.

cogk:
This PR adds an optional parameter to the `surf` command. This allows you to open cartridges directly from the *Console*, using their unique identifier (numeric id or hash).

**TODO**
- [x] load by hash (`surf d02a74c75d80c9bc3b0721ee6f6c83ab`)
- [ ] load by numeric id (`surf 636`)
- [ ] update usage message for the `surf` command

Fixes #1253

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -1476,7 +1476,43 @@ static void onGameMenuCommand(Console* console)
 
 static void onSurfCommand(Console* console)
 {
-    gotoSurf();
+    if(console->desc->count == 0)
+    {
+        gotoSurf();
+    }
+    else if(console->desc->count == 1)
+    {
+        const char*   = console->desc->params->key;
+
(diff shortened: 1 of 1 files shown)


BuoYancYdabl:
It would be cool to load it by tic80.com id.

PR #1530, opened 2021-07-28
Handle emacs ^n ^p and ^e keys in the editor

trufae:


trufae:
It's my first contribution, and i didn't tested the PR too much, but i'm using tic80 with a 60% keyboard that doesnt have arrow keys, so supporting emacs keys for navigating around the code is kind of important to me.

I think this change should be done in other places like the system menu or the shell. Feedback is welcome :)

nesbox (approved):
ok, let's try this
thank you

PR #1533, opened 2021-07-31
Pump events once per loop, fixes #1506

### LINKED ISSUE #1506: [0.90 - linux] Tic-80 freezes the system when window focus is gained after a long time
(Hi there, long time no see)

Version : pro 4fdbdec
OS: ubuntu 21.04

On focus gained, the process seems to be stuck in the event loop, leaving me no option other than to wait (sometimes a few minutes) or restart the computer.

ddelemeny:
Fix for #1506 inspired by the following commit from SDL : https://github.com/libsdl-org/SDL/commit/ce8261dd6da0f1588ed608f392632252c7a5e693

As I understand it, the polling loop gets a crowd of events to process on focus gained and the queue may be topped up with more incoming events everytime SDL_PollEvents is called as it invokes SDL_PumpEvents under the hood, until the processing is finally able to catch up (and it may not). 

This commit changes the pump strategy to avoid topping up the event queue, seems to do the trick.

### DIFF
--- a/src/system/sdl/main.c
+++ b/src/system/sdl/main.c
@@ -1016,7 +1016,7 @@ void tic_sys_poll()
 
     // Workaround for freeze on fullscreen under macOS #819
     SDL_PumpEvents();
-    while(SDL_PollEvent(&event))
+    while(SDL_PeepEvents(&event, 1, SDL_GETEVENT, SDL_FIRSTEVENT, SDL_LASTEVENT) == 1)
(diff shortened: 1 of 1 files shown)


ddelemeny:
Argh, nope. Spoke too fast, just had a freeze again.

Need more brain juice.

nesbox:
ok, I'm closing it then

Issue #1541, opened 2021-08-09
Raspberry Pi 4, baremetal - bug with mkdir command


PetruCHIOrus:
There is one bug on the Raspberry Pi 4 build related to the **mkdir** command. 

The fact is that the command apparently creates a folder not in the tic80 folder, where the visible files are located, but in the actual root of the flash drive. You can go to this folder, but it is impossible to read its contents or write it. 

![photo_2021-08-09_21-02-13](https://user-images.githubusercontent.com/39710644/128759085-764ae31d-688c-417f-882c-910cdbbc2419.jpg)

The **demo** command creates a **bunny** folder also in the actual root of the flash drive, it cannot write cartridges there, and therefore the folder remains empty.

For example, normal command behavior:

![photo_2021-08-09_20-57-16](https://user-images.githubusercontent.com/39710644/128759529-6ce2ebb1-7f4a-4ada-8af4-0429ecbb4958.jpg)

msx80:
Hi there, i took a look at the Rpi3 version, it's buggy but in a different way 😅 It actually creates the folder in the right place but you can't "cd" into it.
If i find some time i'll look into it

nesbox:
Seems fixed here 22378fb300bf0f0b5711dd3ffa59ebff191ad2f5
Thanks to @msx80

PR #1582, opened 2021-09-01
fix path problems with baremeta #1541

msx80:
It was missing some passages throu tic_fs_path to normalize the path.  I've tested it and it seems ok now.


nesbox (approved):
Big thanks

PR #1590, opened 2021-09-08
Improvements for language support

msx80:
Developments for discussion #1289.

Changes: there's now a new file "languages.c", this is the designed place to "glue" all languages together. It defines a "Languages" global array of tic_script_config where all enabled languages go.

tic_script_config has been augmented with stuff that was defined in other places, like file extensions, demo and bunny mark cartridges, etc. (every implementation defines the new fields).

Most of the accesses with SCRIPT_LIST has been removed in favour of a loop on Languages (for which a macro is provided).

The static help text problem has been resolved by inserting placeholders in the static text and replacing them with a dynamic strings at runtime. It's terribly inefficient but since we're talking of a handful of lines, it should be ok. At least it works for now, maybe it can be improved in some way later.

The only remaining place where SCRIPT_LIST exists is in the definition of the vms in tic_core. This is my next step, but since is more delicate i'll do it in a different PR.

Some remaining languages switches (BUILD_WITH_JS etc) lives in console.c where demos are defined. Those could be moved to the appropriate tic_script_config too, if you're ok.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,6 +296,7 @@ macro(MACRO_CORE SCRIPT DEFINE BUILD_DEPRECATED)
     set(TIC80CORE_DIR ${CMAKE_SOURCE_DIR}/src)
     set(TIC80CORE_SRC
         ${TIC80CORE_DIR}/core/core.c
+        ${TIC80CORE_DIR}/core/languages.c
         ${TIC80CORE_DIR}/core/draw.c
         ${TIC80CORE_DIR}/core/io.c
         ${TIC80CORE_DIR}/core/sound.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -737,6 +737,7 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/studio/editors/music.c
     ${TIC80LIB_DIR}/studio/studio.c
     ${TIC80LIB_DIR}/studio/config.c
+    ${TIC80LIB_DIR}/studio/demos.c
     ${TIC80LIB_DIR}/studio/fs.c
     ${TIC80LIB_DIR}/studio/net.c
     ${TIC80LIB_DIR}/ext/md5.c
--- a/include/tic80_config.h
+++ b/include/tic80_config.h
@@ -42,50 +42,6 @@
 #   define TIC_BUILD_WITH_LUA 1
 #endif
 
-#if defined(TIC_BUILD_WITH_LUA)
-#   define TIC_LUA_DEF(macro) macro(lua, ".lua", "--", lua_State)
-#else 
-#   define TIC_LUA_DEF(macro)
-#endif
-
-#if defined(TIC_BUILD_WITH_MOON)
-#   define TIC_MOON_DEF(macro) macro(moon, ".moon", "--", lua_State)
-#else 
-#   define TIC_MOON_DEF(macro)
-#endif
-
-#if defined(TIC_BUILD_WITH_FENNEL)
-#   define TIC_FENNEL_DEF(macro) macro(fennel, ".fnl", ";;", lua_State)
-#else 
-#   define TIC_FENNEL_DEF(macro)
-#endif
-
-#if defined(TIC_BUILD_WITH_JS)
-#   define TIC_JS_DEF(macro) macro(js, ".js", "//", duk_hthread)
-#else 
-#   define TIC_JS_DEF(macro)
-#endif
-
-#if defined(TIC_BUILD_WITH_SQUIRREL)
-#   define TIC_SQUIRREL_DEF(macro) macro(squirrel, ".nut", "//", SQVM)
-#else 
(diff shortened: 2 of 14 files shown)


nesbox (changes requested):
ok, it works, only one thing is we have a dependency loop in the cmake now `tic80core -> build/assets/*.dat -> prj2cart -> tic80core`. Before, it looked like `tic80studio -> build/assets/*.dat -> prj2cart -> tic80core`.
Any thoughts on how to fix it?

msx80:
Ah yes i totally forgot to mention it. Indeed the build works, but if you "make clean" then it can't rebuild the dat. I didn't get it was a dependency cycle tho.
Would a compiler switch like "INCLUDE_ROMS" work? It could be set to true for tic80studio and false for prj2cart. Something like this:

static const u8 JsDemoRom[] =
{
#ifdef INCLUDE_ROMS
    #include "../build/assets/jsdemo.tic.dat"
#endif
};

But perhaps the tic80core target is built just once per run? not sure how the build process work in that regard.
I don't think it can be refactored out either, becouse prj2cart depend right on tic_script_config, so the cycle is real. Or we need to define a different target to build a subset of tic80core without the roms, to be used by prj2cart.

Otherwise the roms needs to be put back in tic80studio. i don't think it can be made completely dynamic tho. 
Let me know :)

nesbox:
Hard to say, on the one hand, we want to keep all the language-related things in one place, in another hand `tic80core` doesn't need to know anything about demo carts and marks, it's only the console related thing.
Maybe we could add `INCLUDE_ROMS` definition for the whole demo code
``` c
#ifdef INCLUDE_ROMS
static const u8 JsDemoRom[] =
{
#include "../build/assets/jsdemo.tic.dat"
};
#else
// lang implementation
#endif
```

and then include the file in the `console.c`
``` c
#define INCLUDE_ROMS
#include "api/js.c"
```
This could partly solve the issue.

msx80:
Uhm so basically js.c  would be two different files: when included by the full version it will be only the demos? It's a bit forced but could work. Problem is, you still need to wire JsDemoRom somewhere, statically. So that means switches and macros in some places.
I don't know, i feel like if we miss the objective of centralizing language related stuff on a single place then the whole thing kind of lose meaning. I mean, the for loop on Languages is certainly more readable than the SCRIPT_LIST kung fu, and that would stay, but..

What about this: we remove cart related stuff from _tic_script_config_, and create a new file with an new type _tic_script_config_extra_ (or something). Here we put the demos etc. Then we create an array of _tic_script_config_extra_ that parallels _Languages_, perhaps _LanguagesExtra_. This new stuff would be included only in studio, resolving the cycle. We end up with two places to define stuff instead of one, but they'll be pretty "parallel" so to speak. And a simple function could look up the right _tic_script_config_extra_ from a _tic_script_config_ to keep things dynamic.

If you like i can give it a try.

Ps if you feel like we're going in the wrong direction and want to throw this away just let me know, i'll understand :)

nesbox:
> What about this: we remove cart related stuff from tic_script_config, and create a new file with an new type tic_script_config_extra (or something). Here we put the demos etc. Then we create an array of tic_script_config_extra that parallels Languages, perhaps LanguagesExtra. This new stuff would be included only in studio, resolving the cycle. We end up with two places to define stuff instead of one, but they'll be pretty "parallel" so to speak. And a simple function could look up the right tic_script_config_extra from a tic_script_config to keep things dynamic.

Sounds good, I like it, demo things will link to the studio and we keep the core clean without unnecessary dependencies.
Thank you

msx80:
Should be ok now. New files are demos.h and demos.c, both under "studio". demos.c contains demos for all languages, it's just mechanical definitions, i don't think there's need to have a different file per language. Perhaps even proper demos (like fire, 3d etc, could be moved there).

i've also committed the modification i mentioned earlier since i was already working on it: the vm pointers at _tic_core_ are now unified, instead of a pointer per language, there are two fields: an opaque pointer (currentVM) and a tic_script_config* (currentScript) that record the information about which language the current vm belongs to, so we can close it. The opening and closing logic has been updated accordingly. This way tic_core doesn't know anything about the current language and is completely dynamic. (also it now never keeps more than one VM open at a time, previously it could have up to one VM per language open at the same time).

nesbox (changes requested):
Great, the dependency loop is fixed now.
One thing, it crashes when I use `demo` command, also I see the unnecessary output in the console, like `getConfigExtra`:
```
>demo
added carts:

fire.tic
font.tic
music.tic
p3d.tic
palette.tic
quest.tic
sfx.tic
tetris.tic
benchmark.tic
bpp.tic
getConfigExtra luaFound extra: luabunny/luamark.tic
getConfigExtra jsNOT Found extra
```
Could you pls fix these before the merge?

msx80:
oh man i even checked multiple times that i removed all printf 🤦 
fixed the crash too, i forgot the poor Javascript.

nesbox (approved):

PR #1596, opened 2021-09-16
Baremetal RPI: added Squirrel language which was previously disabled.

msx80:
Hi there, when i added Baremetal RPI i left out Squirrel becouse the toolkit was having trouble linking C++ stuff. I checked this again and found the problem: i simply wasn't setting C++ compiler flags (only C ones).

Now the baremetal version include all languages, and also any future C++ dependency we add will be able to work.


nesbox (approved):
Awesome 👍

Issue #1475, opened 2021-07-07
Add an additional parameter to peek/poke functions for diferent bit resolutions 


Kyuchumimo:
Certain TIC-80 RAM and VRAM areas has diferent bit resolutions, varying from 1-bit to 12-bit. see: https://github.com/nesbox/TIC-80/wiki/RAM
in this way, it would be more convenient to access certain areas of RAM and VRAM:

(For retrocompatibility)
* peek addr res  -> val
* poke addr val res

StinkerB06:
I think `peekb()` and `pokeb()` functions would do. Basically, `pokeb(0x20000,255,8)` would be equivalent to `poke(0x4000,255)`.

joshgoebel:
If anything I think you'd have peek2 and peek1 perhaps, for consistency. But I don't think either is necessary in the API because they are so trivial to implement.

nesbox:
3ef73db: added `res=1,2,4,8` parameter to `peek()/poke()` api
Now they looks like:
- `peek addr res=8 -> val`
- `poke addr val res=8`

joshgoebel:
@nesbox I was unaware of this.  Please review https://github.com/nesbox/TIC-80/wiki/peek and see if this seems correct... If so I'll update `poke` accordingly.

Of course this also isn't exactly the same [as what was requested here] since it allows no 1, 2, or 4 but precision access to the high bits, only the low bits... but it's nice and should still be documented.

nesbox:
Thank you, it looks good, one thing about addressing low/hi bits, the `address` parameter is aligned to bits depth.
For example, if you want to get `BITS` of `ADDR` pls call `peek(ADDR * 8 / BITS, BITS)`
for `1bits` call `peek(ADDR * 8, 1)`
for `2bits` call `peek(ADDR * 4, 2)`
for `4bits` call `peek(ADDR * 2, 4)`
for `8bits` call `peek(ADDR * 1, 8)`

another example
```
ADDR  VALUE
0     0b00000000
1     0b00000000
2     0b00000000
3     0b01100000
```
If you want to peek 2bits of the `ADDR=3` with offset `5` pls call `peek(3 * 4 + 5, 2)`

nesbox:
Additional API methods won't help here, because `peek4()` do the same, and `ADDR` is aligned by bits depth here too. :)
So `peek(ADDR, 4) == peek4(ADDR)` where `ADDR` is offset of 4bit value in the RAM.

[...]

PR #1616, opened 2021-10-06
Poke1 and Peek1

SuperPupperDoggo:
I think these would be useful, especially for compact single-bit flags in save files.

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -404,6 +404,29 @@ enum
         tic_mem*, s32 address, u8 value)                                                                                \
                                                                                                                         \
                                                                                                                         \
+    macro(peek1,                                                                                                        \
+        "peek1(addr) -> value",                                                                                         \
+                                                                                                                        \
+        "This function enables you to read single-bit values from TIC's RAM.\n"                                               \
+        "The address should be specified in hexadecimal format.",                                                       \
+        1,                                                                                                              \
+        u8,                                                                                                             \
+        tic_mem*, s32 address)                                                                                          \
+                                                                                                                        \
+                                                                                                                        \
+    macro(poke1,                                                                                                        \
+        "poke1(addr value)",                                                                                            \
+                                                                                                                        \
+        "This function allows you to write to the virtual RAM of TIC.\n"                                                \
+        "It differs from `poke()` in that it accesses the individual bits of memory.\n"                                     \
+        "Therefore, to address the highest bit of position 0x4000 you should pass 0x16000 as addr1, "                    \
+        "and to access the lowest bit (rightmost bit) you would pass 0x16007.\n"                                      \
+        "The address should be specified in hexadecimal format, and values should be given in decimal.",                \
+        2,                                                                                                              \
+        void,                                                                                                           \
+        tic_mem*, s32 address, u8 value)                                                                                \
+                                                                                                                        \
+                                                                                                                        \
     macro(memcpy,                                                                                                       \
         "memcpy(dest source size)",                                                                                     \
                                                                                                                         \
--- a/src/api/js.c
+++ b/src/api/js.c
@@ -535,6 +535,26 @@ static duk_ret_t duk_poke4(duk_context* duk)
     return 0;
 }
 
+static duk_ret_t duk_peek1(duk_context* duk)
+{
+    s32 address = duk_to_int(duk, 0);
+
+    tic_mem* tic = (tic_mem*)getDukCore(duk);
+    duk_push_uint(duk, tic_api_peek1(tic, address));
+    return 1;
+}
+
+static duk_ret_t duk_poke1(duk_context* duk)
+{
+    s32 address = duk_to_int(duk, 0);
+    u8 value = duk_to_int(duk, 1);
+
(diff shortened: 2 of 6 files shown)


joshgoebel:
Though I wonder if the API should perhaps be different at that level... `bitset`, `bitget` or some such?

StinkerB06:
This can already be done with existing `peek()` and `poke()`.

SuperPupperDoggo:
that is a good point. poke1 and peek1 is more consistent considering that poke4 and peek4 exist, but it is also a bit weird. perhaps `pokebit` and `peekbit` would be a good name?

joshgoebel:
> The same thing could be said for poke4 and peek4, and they still exist

I think it's about balancing the need for a useful API and how common certain operation are.  Since the graphics RAM is nibble based, peek4 and poke4 make a lot of sense for writing to VRAM.  That argument doesn't really hold up for peek1 and poke1, which are easy enough utilities to write by hand and wouldn't add the same amount of value to the same amount of people.

So it's not the same thing.

SuperPupperDoggo:
> > The same thing could be said for poke4 and peek4, and they still exist
> 
> I think it's about balancing the need for a useful API and how common certain operation are. Since the graphics RAM is nibble based, peek4 and poke4 make a lot of sense for writing to VRAM. That argument doesn't really hold up for peek1 and poke1, which are easy enough utilities to write by hand and wouldn't add the same amount of value for the same amount of people.
> 
> So it's not the same thing.

I mean, you do have a point, but once again the same thing could be said for `poke4` and `peek4`. They are very rarely used. Writing directly to VRAM is not very common, because if you want to interact with individual pixels on the screen, `pix` is smaller in terms of size and more efficient than writing an algorithm to calculate the address of a pixel, and the palette is made up of 24-bit entries. While there is border color, most people don't even know it exists, and I have never seen it used.

joshgoebel:
> While there is border color

Ah, yes, I forgot about nibble sized "registers" at various memory locations. :-)

vsariola:
I'll just chime in that where I come (demoscene), we use poke4 / peek4 a lot to access video memory.

But regarding the design choice of packing nibbles in memory: I would've preferred not packing two nibbles in the video memory in the first place, but do it like c64 color memory: higher nibbles never used. Similarly, registers should not pack nibbles / bits if possible. This way, only poke and peek would be needed. TIC-80 ain't really using only 64k of memory, not even close, so  why bother making it more hard to do things.

nesbox (changes requested):
Why not to add `peek2()/poke2()` in this case :)
Also, you can always call `peek(addr,1)` instead `peek1(addr)`, was added here #1475
Thanks

joshgoebel:
I think the availability of both `poke(x, 1)` and `peek(x,1)` likely mean we should close this with "already exists"?

Should `peek4` and `poke4` possible be deprecated in a future release?  I assume they would have never been added if this more flexible `poke/peek` API had already existed at that time?

vsariola:
Before deprecating them, let's a) profile the performance also, if the new poke api is slower to do the same thing or not. poke4/peek4 tends to be the hottest loops of them all: drawing fullscreen graphics for every pixel. b) consider if clean API is worth breaking existing cartridges, at least for a good while

joshgoebel:
> Before deprecating them, let's a) profile the performance also, if the new poke api is slower to do the same thing or not. 

Ah, this is indeed a good suggestion and would perhaps be an argument for keeping.  It's quite possible that `poke4` and `peek4` are actually SLOWER though if they are just using the underlying more complex `poke` API anyways (or that the different really isn't significant).  I think that's what I recall from reading the source, but I could be mistaken.

> consider if clean API is worth breaking existing cartridges

_There are many ways to deprecate things_, there is no need to "break existing cartridges".  Rather it could simply be that when creating new cartridges those APIs are no long accessible, etc...

nesbox:
I am against deprecating any API, it will break compatibility with old cartridges.
I'll merge this PR and going to add `poke2()/peek2()` since we have `peek4`

nesbox:
Added peek2()/poke2() api here 2db3134

joshgoebel:
Updating `peek` and `poke` API docs and consolidated the peek4/poke4 information there as well to avoid duplicating information everywhere.  I think we can explain everything better (and in more detail) in a single place.  I added redirects from the old API destinations (peek4/poke4).

- https://github.com/nesbox/TIC-80/wiki/peek
- https://github.com/nesbox/TIC-80/wiki/poke

The API docs do mention version 1.0 now, which is not released.... Should that we a link to a `versions` page where we explain which releases are out vs still in beta?  I think we need to keep the API up-to-date as we add things (even if they aren't yet released)... which means there will be mentions of 1.0.  I just wonder if that will lead to any confusion when someone gets looking for version 1 and can't find it?

nesbox:
I usually add a badge with the version number in which this function works, like `(0.90)`

joshgoebel:
Are you suggesting that instead of "added in 1.0"?

nesbox:
Personally, I just like `(1.0)` 😄 but I don't mind

joshgoebel:
Then I think I'll leave it for now, I like trying new things every once in a while - see how they feel. :)

Issue #714, opened 2018-09-01
Internal audio latency


triplefox:
When running TIC-80, across multiple platforms(Windows desktop, HTML5 player, Android), I always notice small pops and creaks when it plays sound, symptoms of frequent buffer underruns during audio playback. The current method it uses is to call SDL_QueueAudio from within the render tick and blit everything that has been rendered in that frame, which leaves the program heavily reliant on making every frame deadline every time.

Most audio programs run a separate audio thread with intentionally added frames of latency to smooth this out: instead of handing over all the samples that have been constructed right at that moment, they're repackaged into fixed-size audio frames for API consumption, and the renderer stays ahead of the API by a target number of frames. This adds a margin of safety, and the fixed size plays better with most audio APIs. Many apps also expose some user configuration options for buffer sizes.

nesbox:
In few words, we have to add empty buffer to the audio frames to allow slightly late for the sound system...
I'll try to add this, thanks.

borbware:
probably related to this: on my machine, if i leave tic-80 running for a long time, audio latency keeps building up, and before long, audio is played with a latency of several seconds. (0.70.6 pro, win10)

nesbox:
Does your machine go to sleep mode during this time?

borbware:
no, it does not! It happens steadily and gradually if I just keep TIC-80 open for a while (e.g., if it's on for a couple of hours, there is a guaranteed audio latency of a few seconds)

nesbox:
need investigation, thank you @borbware

runlow:
Experiencing the same problem as @borbware is describing but on Archlinux.

If I open a game or the music tracker for too long then I have to close TIC-80 and open it again. If I don't do this every 10 minutes or so the sound delay becomes greater and greater. It doesn't take hours - few minutes is enough. This makes the music tracker or any game with longer gameplay unusable/unplayable for me.

When I run TIC80 from the command line it would slowly keep showing the following message in the terminal but the program would keep running
`ALSA lib pcm.c:8545:(snd_pcm_recover) underrun occurred`

From alsa documentation:
"The underrun can happen when an application does not feed new samples in time to alsa-lib (due CPU usage)"

I've noticed that when I run pulseaudio this delay seems to go away and the sound on TIC-80 seems OK but running pulseaudio in turn causes other problems with some other programs that use the soundcard so I would prefer not to run pulseaudio.

I'm not sure I fully understand what the problem is but it seems to have something to do with audio buffering.

[...]

PR #1618, opened 2021-10-07
#714: fix crackling audio.

vsariola:
Previously, the sound registers were used to synthesize fixed amount of sound after every TIC() and this was pushed to SDL. The problem with this approach was crackling if TIC() takes too long to reach 60Hz. In this fix, the sound registers are instead pushed to a ring buffer and the audio is synthesized in SDL callback, to render audio. If the ring buffer does not have enough values, the sound synthesis uses the last good values from the ring buffer, so the music slows down yes, but at least it does not crackle.

The corollary to this is that calling tic_core_tick_end(tic) is not enough during the Studio sound export; one should also call tic_core_synthesize_sound which does the job of pulling the sound registers from the tail of the ring buffer and doing the sound synthesis.

The ring buffer is kept short on purpose, because if there slight desync between the TIC() frequency and sound frequency, the TIC() frequency might run faster and delay might build up. So, the ring buffer length sets a maximum on the delay that this technique can create.

Just to note that you might want to experiment with TIC_SOUND_RINGBUF_LEN and SDL_AudioSpec.samples to get good amount of buffering, without significant delays, and not too much overhead. SDL_AudioSpec.samples controls how many times a second the callback is called and presumably there's a small overhead with each call.

This fix also btw probably fixes the complaints about the sound getting desynchronized with the video; the short ring buffer ensures that the audio can never be too much out of sync with the video.

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -709,6 +709,7 @@ void tic_core_tick_start(tic_mem* memory);
 void tic_core_tick(tic_mem* memory, tic_tick_data* data);
 void tic_core_tick_end(tic_mem* memory);
 void tic_core_blit(tic_mem* tic);
+void tic_core_synthesize_sound(tic_mem* tic);
 
 void tic_core_blit_ex(tic_mem* tic, tic_blit_callback clb);
 const tic_script_config* tic_core_script_config(tic_mem* memory);
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -28,6 +28,7 @@
 
 #define CLOCKRATE (255<<13)
 #define TIC_DEFAULT_COLOR 15
+#define TIC_SOUND_RINGBUF_LEN 6 // in worst case, this induces ~ 6 tick delay i.e. 100 ms
 
 typedef struct
 {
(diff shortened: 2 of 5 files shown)


vsariola:
I'm getting reports that the studio might feel sluggish with this patch. I don't feel it but perhaps this is related to computer speed. One plausible explanation is that now with the audio synthesis done in the callback, maybe SDL only has one thread for doing event and audio handling? And if we are doing lot of stuff in the audio callback (the whole sound synthesis), then there might not be enough time to handle e.g. key events.

This is just a theory; please test and comment.

vsariola:
I reduced slightly the callback frequency; SDL_AudioSpec.samples is now 1024. I also made the ring buffer have a length of 6 items. Furthermore, I cleaned some unnecessary whitespace.

nesbox:
Hi, thank you for the PR, but unfortunately I'm in a hospital now and can't review/merge it. Sorry.

vsariola:
No problem! Get well soon!

nesbox (approved):
Just tested your version and it works perfectly, thank you very much for fixing this annoying issue.
One thing, we also have to update the `player.c` and other implementations, like `libretro/n3ds/sokol/...`

PR #1628, opened 2021-10-09
Optimize tic api call performance in LUA.

vsariola:
Previously, the tic_core* was associated with a global variable named "_TIC80". Finding the global variable with that name took surprising amount of time, according to profiling. With this optimization, each API function is associated with tic_core* through lua_pushcclosure mechanism. Closures are functions that are associated with certain "upvalues". Fetching an upvalue is very fast in LUA.


nesbox (approved):
Yes, it works faster now.
Thank you for the help.

Issue #800, opened 2019-01-26
Eval nonfunctional?


ElfEars:
When attempting to eval code on the command line, nothing after the command is actually run.
I've attempted to modify the map and the sprite data of the loaded program with it and nothing happens.

I think there should be better cohesion between the command line and the code editor ala pico-8 (making it easier to perform poke/memcopy operations loaded programs) but that's probably for another issue (awkward due to the multiple languages. Maybe a command for each language ie: lua mset(0,0,32) ??)

RobLoach:
eval code? What are you running?

nesbox:
@RobLoach we have `eval` command in the console, you can pause a game and call code with `eval <code>` and `resume`, for example.
Implemented only in Lua atm.

joshgoebel:
@ElfEars Did you `run` the cartridge first?  I found that some state seems to be necessary before eval does anything useful.

atesin:
> @ElfEars Did you `run` the cartridge first? I found that some state seems to be necessary before eval does anything useful.

this should be the "accepted answer" ...  i had the same problem that eval "does nothing", until i `run` default cart then exit with [esc] key

until not the ideal condition... but to prevent head smashings in the while, it should be documented here anyway -> https://github.com/nesbox/TIC-80/wiki/Console#evaluating-code

(or better being fixed)

technomancy:
I agree it's not very good the way it currently works without giving any feedback.

I tried to fix it, but unfortunately it just segfaults:

```
diff --git a/src/api/lua.c b/src/api/lua.c
index 696d8569..a9a633af 100644
--- a/src/api/lua.c
+++ b/src/api/lua.c
@@ -1778,7 +1778,10 @@ static void evalLua(tic_mem* tic, const char* code) {
     tic_core* core = (tic_core*)tic;
     lua_State* lua = core->currentVM;
 
-    if (!lua) return;
+    if (!lua) {
+      tic_api_trace(tic, "must run before eval", 1);
+      return;
+    }
 
     lua_settop(lua, 0);
 ```

I also tried this to automatically initialize, but it segfaults as well:

```
diff --git a/src/api/lua.c b/src/api/lua.c
index 696d8569..91a292cf 100644
--- a/src/api/lua.c
+++ b/src/api/lua.c
@@ -1778,7 +1778,10 @@ static void evalLua(tic_mem* tic, const char* code) {
     tic_core* core = (tic_core*)tic;
     lua_State* lua = core->currentVM;
 
[...]

PR #1633, opened 2021-10-14
Fix eval command

technomancy:
This fixes issue #800.

I am not 100% sure this is in line with the original intent of the fields, and I have very little C experience!

The CommandDesc struct has two relevant fields here: src and
command. The src field is supposed to contain the source of the entire
input while the command field is supposed to contain just the name of
the command being invoked; aka everything before the first space.

We would start with input like this:

    eval trace("hello" .. "world")

What we want is:

   src: eval trace("hello" .. "world")
   command: eval

The src field was set to a copy of the input, and then the command
field was set to running strtok on src to find the first space. But
even though strtok returns what looks like a string, it's actually
just pointing to the same copy of src which now happens to have been
terminated right where the first space used to be.

So what we got was just:

   src: eval
   command: eval

The fix is to copy the input into src and command fields separately
and only call strtok on command, allowing src to be left alone.

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -2318,7 +2318,8 @@ static void onEvalCommand(Console* console)
     if (script_config->eval)
     {
         if(console->desc->count)
-            script_config->eval(console->tic, console->desc->params->key);
+            script_config->eval(console->tic,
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Thank you.

PR #1653, opened 2021-10-30
Activate strict mode in Fennel when strict: true header is found.

technomancy:
Strict mode makes it so that it is a compile error to refer to a
global which doesn't exist yet. Using (global foo {}) will declare foo
as a global which can be used after the declaration, or you can always
bypass the checks using _G.foo.

This will catch a lot of errors earlier on during the compilation
process instead of at runtime. This should have been on by default
from the start. But better late than never. =)

We change the demo Fennel cart to add the strict header by default for
new games, but it's easy to delete for users who want the more
Lua-like global behavior.


technomancy:
I changed `demos/fenneldemo.fnl` and the binary version inside the `build/` dir, but I could not get my changes to the default `new fennel` cart to reflect the `;; strict: true` header when I build locally and run it.

Is this just a flaw in the build process that the old version of the cart has stuck around, or did I miss a file I need to change to update it?

nesbox (approved):
Thank you

nesbox:
Btw, just tried this card https://tic80.com/play?cart=2302 with your changes and got 
```
tic80.com>game:84: Compile error in 'match': runtime error: expected even number of pattern/body pairs
stack traceback:
  [C]: in function 'assert'
  src/fennel/macros.fnl:428: in function ?
  [C]: in function 'xpcall'
  [string "fennel.lua"]:2621: in function 'macroexpand_2a'
  [string "fennel.lua"]:2784: in function 'compile1'
  [string "fennel.lua"]:2852: in function 'compile_top_target'
  [string "fennel.lua"]:2873: in function 'destructure_sym'
  [string "fennel.lua"]:2957: in function 'destructure1'
  [string "fennel.lua"]:2969: in function 'destructure'
  [string "fennel.lua"]:1267: in function 'special'
  [string "fennel.lua"]:2635: in function ?
  [string "fennel.lua"]:3000: in function ?
  [string "fennel.lua"]:4014: in function ?
  [C]: in function 'pcall'
  [string "execute_fennel"]:1: in main chunk
```

Any chance you add backward compatibility with old carts?

technomancy:
Thanks!

I believe the problem you're seeing is not caused by this change but by the previous one; we fixed a bug, but this code was relying on that bug. I know the author of that cart so I sent them a fix here: https://github.com/peterhil/alchemy/pull/5

This PR should not have any adverse effects because it only triggers when the `;; strict: true` comment is found.

However, we are about to release Fennel 1.0.0 which has a few breaking changes in it intentionally. (Hopefully the last breaking changes ever.) Nothing big; similar to the change from Lua 5.3 -> 5.4. But I'm not sure how to address this problem.

I have a list of all the Fennel games I know about and have been testing them manually here: https://git.sr.ht/~technomancy/fennel-compendium (I actually found the problem you mentioned a few days ago but I only mentioned it to the author on IRC instead of opening a PR.) So if the number of Fennel carts is small maybe that's enough; we can just manually reach out to the authors. It won't scale indefinitely, but then again, we don't plan on making any backwards-incompatible changes after 1.0.0 either, so maybe it's OK.

Do you know of any other Fennel TIC-80 carts which aren't listed? https://git.sr.ht/~technomancy/fennel-compendium/tree/main/item/Makefile#L100

PR #1658, opened 2021-11-03
Command Request

mirkokral:
Please add a copy and move command because i don't want to open the folder in the system every time i want to rename/move/copy a file.


nesbox:
Pls use the Issues section for this
https://github.com/nesbox/TIC-80/issues

PR #1667, opened 2021-11-04
(enh) errors when mkdir fails

### LINKED ISSUE #1583: mkdir always returns successfully
Console command "mkdir" doesn't perform any check on the success of the underling folder creation. It responds with ok even if the folder is already existing, is a file or contains invalid characters. The behaviour can be traced to fs.c where the system appropriate "mkdir" primitive is called, but the response is ignored.

joshgoebel:
Closes #1583.


nesbox (approved):
Looks good, thank you for the help.

PR #1670, opened 2021-11-05
(chore) move moonscript into separate C file

joshgoebel:
Feels like every language should have their own C file in `api`... so I wonder if this should be broken out?

### DIFF
--- a/src/api/squirrel.c
+++ b/src/api/squirrel.c
@@ -1447,7 +1447,7 @@ static void squirrel_open_builtins(HSQUIRRELVM vm)
     sq_poptop(vm);
 }
 
-static void initLuaAPI(tic_core* core)
+static void initAPI(tic_core* core)
 {
     HSQUIRRELVM vm = core->currentVM;
     
--- a/src/api/squirrel.c
+++ b/src/api/squirrel.c
@@ -1447,7 +1447,7 @@ static void squirrel_open_builtins(HSQUIRRELVM vm)
     sq_poptop(vm);
 }
 
-static void initAPI(tic_core* core)
+static void initLuaAPI(tic_core* core)
 {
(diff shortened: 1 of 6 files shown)


joshgoebel:
Would this be good?  If so I can do the same with Fennell...

src/api/squirrel.c
```diff
@@ -1447,7 +1447,7 @@ static void squirrel_open_builtins(HSQUIRRELVM vm)
     sq_poptop(vm);
 }
 
-static void initAPI(tic_core* core)
+static void initLuaAPI(tic_core* core)
```
nesbox, line 1450: It looks like the function was accidentally renamed here, there should be no mention of Lua in the Squirrel module.

src/api/wren.c
```diff
@@ -1395,7 +1395,7 @@ static WrenForeignMethodFn bindForeignMethod(
     return foreignTicMethods(fullName);
 }
 
-static void initAPI(tic_core* core)
+static void initLuaAPI(tic_core* core)
```
nesbox, line 1398: Same here, please fix if not difficult.
Thank you.

nesbox (changes requested):

joshgoebel:
Yeah, no problem.  I'll fix this up and make sure the tests are green.  Just didn't want to throw more work at this if you aren't ok with the change in general.

nesbox (approved):
thank you

joshgoebel:
Yeah, not sure this was 100% ready yet. :)  You got to it too fast.

nesbox:
ok, waiting... :)

joshgoebel:
But you already merged it... did you get it working or should I just open a new PR? :-)

nesbox:
Honestly, I don't know what to do in this case :)

PR #1672, opened 2021-11-05
(enh) Music editor: Skip over sharp column when using cursor keys

### LINKED ISSUE #1625: tracker "sharp column" should be editable or not selectable
When you press *right* on a note - you end up on the "sharp" (hash symbol `#`) column which you can't interact with in any way. 

When you are in the tracker and on a note (first column), and then press *right* - it should take you directly to the octave number column.

Since you cannot interact with this "sharp column" - it doesn't seem warranted that you should be able to select it as it's an extra keypress when you move the cursor around.

Alternatively - you should be able to toggle whether a note is sharp or not when on the "sharp column".

joshgoebel:
Allows toggling of sharp.  I guess technically it doesn't even require you to be in the column...

Resolves #1625.

### DIFF
--- a/src/studio/editors/music.c
+++ b/src/studio/editors/music.c
@@ -280,6 +280,10 @@ static void leftCol(Music* music)
 {
     if (music->tracker.edit.x > 0)
     {
+        // skip sharp column
+        if(music->tracker.edit.x == 2) {
+            music->tracker.edit.x--;
+        }
         music->tracker.edit.x--;
         updateTracker(music);
     }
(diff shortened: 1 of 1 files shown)


runlow:
I still think it should not be possible to even "be in the column" if being in it doesn't do anything.

joshgoebel:
Rebased to only include the skipping behavior, which feels very nice and natural. (and tiny PR)

nesbox (approved):
Looks good, thank you

nesbox:
Please disable Draft mode for the PR if it's ready

joshgoebel:
Sorry!

runlow:
Thank you!

PR #1680, opened 2021-11-07
add MacOS application instructions

lzap:
Sup,

I am playing around with MacOS as well and launching the development version is a bit weird via Spotlight (it creates a terminal window by default). So here are instructions for hackers who would like to launch it conviniently.

I tested this on my latest MacOS (Monterey is it called?) and I noticed you guys provide also a DMG which installs as "tic-80" so I renamed this to be "TIC-80dev" so people can have both stable and dev versions installed and switch between each other.

Cheers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,20 @@ cmake ..
 make -j4
 ```
 
+to create application icon for development version
+```
+mkdir -p ~/Applications/TIC80dev.app/Contents/{MacOS,Resources}
+cp -f macosx/tic80.plist ~/Applications/TIC80dev.app/Contents/Info.plist
+cp -f macosx/tic80.icns ~/Applications/TIC80dev.app/Contents/Resources
+cat > ~/Applications/TIC80dev.app/MacOS/TIC80dev <<EOF
+#!/bin/sh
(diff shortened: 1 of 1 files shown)


RobLoach:
Should this be in install instructions instead, perhaps even in the install package? Something for homebrew?

lzap:
The package does this for you, this is meant for developers or users who would like to test out nightly builds and still have a nice way of launching the app. That was the idea.

nesbox (approved):
Thank you

nesbox:
I think we'll move all the **Build Instructions** from the `readme.md` to the **Wiki**...

nesbox:
As far as I remember we had two versions of the build instructions in the readme and in the wiki and they were all moved to the readme only. This is not a problem at the moment, but it is growing, I don’t know if it’s good to have that much text on the homepage.

joshgoebel:
See our README. :)  https://github.com/highlightjs/highlight.js

I mean we can move it back if we have to, but I kind of thought perhaps it benefited from the extra review that comes from being in the README rather than the wiki... I think a table of contents (which GitHub does now by default) is the answer to a long readme. :-)

PR #1684, opened 2021-11-09
suggest pro when trying to open source with standard

### LINKED ISSUE #1643: Print a more helpful message when trying to load text carts with non-pro version
Hello!

Thanks for TIC-80.

I'm happy to help fix this issue, but wanted to get it up before i forget. I was trying to open text carts (like cart.fnl) and just getting a "cart failed to load" error. I found out though #lispgames that it is because I don't have the pro version. It might be nice to indicate that as an error from TiC-80 for those who didn't manage to see it on the website?

joshgoebel:
Closes #1643.

Screen Shot 2021-11-09 at 9 07 55 AM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -748,9 +748,9 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/ext/png.c
 )
 
-#if(${BUILD_PRO})
+if(${BUILD_PRO})
     set(TIC80STUDIO_SRC ${TIC80STUDIO_SRC} ${TIC80LIB_DIR}/studio/project.c)
-#endif()
+endif()
 
 set(TIC80_OUTPUT tic80)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -748,9 +748,9 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/ext/png.c
 )
 
-if(${BUILD_PRO})
+#if(${BUILD_PRO})
     set(TIC80STUDIO_SRC ${TIC80STUDIO_SRC} ${TIC80LIB_DIR}/studio/project.c)
-endif()
+#endif()
 
 set(TIC80_OUTPUT tic80)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -738,7 +738,6 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/studio/editors/sfx.c
     ${TIC80LIB_DIR}/studio/editors/music.c
     ${TIC80LIB_DIR}/studio/studio.c
-    ${TIC80LIB_DIR}/studio/project.c
     ${TIC80LIB_DIR}/studio/config.c
     ${TIC80LIB_DIR}/studio/demos.c
     ${TIC80LIB_DIR}/studio/fs.c
@@ -748,6 +747,10 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/ext/gif.c
     ${TIC80LIB_DIR}/ext/png.c
 )
+
+#if(${BUILD_PRO})
+    set(TIC80STUDIO_SRC ${TIC80STUDIO_SRC} ${TIC80LIB_DIR}/studio/project.c)
+#endif()
 
 set(TIC80_OUTPUT tic80)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -738,6 +738,7 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/studio/editors/sfx.c
     ${TIC80LIB_DIR}/studio/editors/music.c
     ${TIC80LIB_DIR}/studio/studio.c
+    ${TIC80LIB_DIR}/studio/project.c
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -748,9 +748,9 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/ext/png.c
 )
 
-if(${BUILD_PRO})
+#if(${BUILD_PRO})
     set(TIC80STUDIO_SRC ${TIC80STUDIO_SRC} ${TIC80LIB_DIR}/studio/project.c)
```
joshgoebel, line 752: This is needed because we need `tic_project_ext` unless you wanted to factor that out into `tools`... I'm not sure what the harm is in building this file though - only the single function should actually get linked during compile.

CMakeLists.txt
```diff
@@ -738,6 +738,7 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/studio/editors/sfx.c
     ${TIC80LIB_DIR}/studio/editors/music.c
     ${TIC80LIB_DIR}/studio/studio.c
+    ${TIC80LIB_DIR}/studio/project.c
```
nesbox, line 741: `project.c` should only be built in Pro version, why did you add it here?

CMakeLists.txt
```diff
@@ -748,10 +749,6 @@ set(TIC80STUDIO_SRC
     ${TIC80LIB_DIR}/ext/png.c
 )
 
-if(${BUILD_PRO})
```
nesbox, line 751: Pls, restore these lines

src/studio/screens/console.c
```diff
@@ -926,7 +929,20 @@ static void onLoadCommandConfirmed(Console* console)
                 }
                 else printError(console, "\nfile not found");
 #else
-                printError(console, "\ncart loading error");
+                if(tic_project_ext(name)) {
+                    printError(console, "\nproject loading error");
+                    printFront(console, "\nThis version only supports binary cartridges (GIF, PNG, or TIC).");
```
nesbox, line 934: We don't have GIF cartridges :)
Only, PNG and TIC

CMakeLists.txt
nesbox: I saw below you added this for the sake of `tic_project_ext()` function, maybe better move it to the `src/tools.c`

nesbox (changes requested):

src/studio/screens/console.c
joshgoebel: Aren't their own GIF cartridge out there?

CMakeLists.txt
joshgoebel: I could... but curious, what is the harm in building it if it's never linked?  I don't see any real harm.

src/studio/screens/console.c
```diff
@@ -926,7 +929,20 @@ static void onLoadCommandConfirmed(Console* console)
...
+                    printLine(console);
+                    printFront(console, "\nTIC-80 ");
+                    consolePrint(console,"PRO",tic_color_light_blue);
+                    printFront(console, " is needed for projects.");
```
technomancy, line 938: Thanks for looking into this.

It seems like this message is conflating "projects" and "text files". I guess within the code those are synonymous, but to an end user who has never read the implementation, there is no way to understand this.

The message should use the word "text" rather than "projects" because that's the term that the user will understand.

joshgoebel: I see your point, but I don't love that either.  I think perhaps we just repeat the extension of the file?

- `...is needed for LUA projects.`
- `...is needed for WREN projects.`
- `...is needed for JS projects.`

Is that clearer? And if projects is still confusing I support we could replace it with `files`.

technomancy: I don't think it's really any clearer, because it's possible to have Lua, Wren, JS, etc games which are stored in binary carts or in text files. If you specifically want to avoid the word "text" for some reason then `needed for .lua files` would be better than putting the language in all-caps, since LUA is not an acronym. But saying "text files" is definitely clearer.

CMakeLists.txt
nesbox: Why should we include it in the project if its code is not used in the PRO build?
Another way, you can add it but pls wrap all the functions with 
```c 
#if defined(TIC80_PRO)
...
#endif
```

src/studio/screens/console.c
nesbox: Never heard.

src/studio/screens/console.c
nesbox: I also vote for the "text files" or "text carts"

joshgoebel:
Screen Shot 2021-11-10 at 7 55 47 AM

CMakeLists.txt
joshgoebel: >  if its code is not used in the PRO build?

_But it's code IS used in the build now_... so including it makes sense.  The functions that aren't referenced are never included always because the linker will just silently discard them.

You're way overthinking this one, but I went ahead and just moved the function to `tools`...

nesbox:
Looks good, thank you.

nesbox (approved):

PR #1700, opened 2021-11-20
No need for a heading

Tyler887:
We do not need a heading at the very top of our markdown file. We could get rid of it.


RobLoach:
Disagree. Having a heading means the heading will be there, meaning a title is available. Without it it just starts the file with "Our Pledge", which is not the content's title.

PR #1706, opened 2021-11-22
Automatically mark stale issues and PRs

Tyler887:


joshgoebel:
Personally I think this is better handled by more active curation which I've been trying to help out a bit more with lately. Just my 0.02.

nesbox:
I agreed, there is no problem handling them manually, there are not many of them. Maybe in the future when the project will be grown to a huge size and we have thousands of issues 😄 Sorry

Issue #1687, opened 2021-11-10
Drawing VRAM trails glitch


technomancy:
On the master branch there's a graphical glitch in certain games (such as https://tic80.com/play?cart=692) which does not occur on the stable release. Moving the cursor leaves behind trails and doesn't clear it.
![powder-bug](https://user-images.githubusercontent.com/141/141035612-94c9036d-634a-4b12-bf4e-23c218eec255.gif)

Someone on IRC told me that they bisected this problem to this commit: https://github.com/nesbox/TIC-80/commit/973dd00a3227036addb073472f0ec49757664784

nesbox:
I reproduced it, will try to fix, thank you

nesbox:
#1709 dup

Issue #1508, opened 2021-07-21
clip() state should be in RAM


ddelemeny:
Actually it should probably be in VRAM, but memory is scarce up there. Anyway, it seems odd that this state cannot be peek'd/poke'd.

Use case:
One can quite easily implement true-coordinates sprite drawing function using clip/spr, but this discards and overwrites any existing clip state.

nesbox:
I moved the `BLIT SEGMENT` near the `BORDER` and got 4 free bytes for the `CLIP` data a54f4b1.
VRAM looks now
```
>help vram
+-----------------------------------+
|          16KB VRAM LAYOUT         |
+-------+-------------------+-------+
| ADDR  | INFO              | BYTES |
+-------+-------------------+-------+
| 00000 | SCREEN            | 16320 |
| 03FC0 | PALETTE           | 48    |
| 03FF0 | PALETTE MAP       | 8     |
| 03FF8 | BORDER / BLIT SEG | 1     | <<<
| 03FF9 | SCREEN OFFSET     | 2     |
| 03FFB | MOUSE CURSOR      | 1     |
| 03FFC | CLIP              | 4     | <<<
+-------+-------------------+-------+
```

joshgoebel:
😮 And now we have 0 free VRAM. 

Oh wowsers...

joshgoebel:
From chat:

> Will this affect compatibility with previous games that use Blit Segment on TIC-80 0.80.1344 and  0.90.1723?

@nesbox

nesbox:
Yes, it will affect if you used **blit segment** in 0.80 and 0.90, but here I'm ready to put up with it, almost nobody uses it.
Also, I think, we have to release the dev version (0.91) asap to prevent using the old address.

joshgoebel:
Oh ok. :-)   If you're fine with it in this case, that's good. :-)

nesbox:
I removed CLIP data from the VRAM here a6b4b7e, a lot of complaints from users.
Sorry.

joshgoebel:
Why, brave it also broke border color?

nesbox:
Yes, also a lot of carts uses `poke(0x03FF8, color)` instead `poke(2*0x03FF8, color)` and this changes the `blit` segment.
So, as usual, we can't change the spec to not break compatibility.

joshgoebel:
[...]

PR #1708, opened 2021-11-26
Update Fennel to version 1.0.0.

technomancy:
This updates Fennel to the latest stable release, 1.0.0: https://fennel-lang.org/v1.0.0/#section

Note that our intent is that this is the last Fennel release ever to contain intentionally breaking changes, other than for potential security issues. We also expect to slow down on development of the language and focus more on ecosystem stuff after this, so I should be pestering you with fewer PRs like this in the future. =)

I have tested this with all known Fennel TIC-80 games. I found two problems.

One of them is that Amethyst Waters has a graphical glitch:

amethyst-glitch

The other is that the player sprites in Butterfly Effect are not rendered: https://alloy-d.itch.io/butterfly-effect In particular the calls to `spr` from the `draw-player` function appear to have no effect.

However, both these bugs also manifest on the current git build from before this patch is applied, so they do not seem to be related to the upgrade to Fennel 1.0.0.


nesbox (approved):
Great, thank you.
I'll check these games later.

technomancy:
Thanks for merging this.

I went back to check and it seems that the Amethyst Waters bug is a duplicate of #1687; going back to the previous commit as the one mentioned there makes the problem disappear, so that one at least is covered.

The other issue was caused by #1508 so I can open a separate issue about that.

PR #1721, opened 2021-12-06
Add Shift+Return shortcut to the code editor that auto-closes a block

blinry:
This is a convenience feature shamelessly stolen from PICO-8: when pressing Shift+Return in the code editor, insert an `end` statement, and put the cursor in an indented line in-between. For example, we go from

     for x=0,239 do|

to

    for x=0,239 do
     |
    end

I would love to use this during speedcoding! Is this a welcome feature? Does this PR look alright?

### DIFF
--- a/src/studio/editors/code.c
+++ b/src/studio/editors/code.c
@@ -1157,6 +1157,24 @@ static void doTab(Code* code, bool shift, bool crtl)
     else inputSymbolBase(code, '\t');
 }
 
+// Add a block-ending keyword or symbol, and put the cursor in the line between.
+static void newLineAutoClose(Code* code)
+{
+    newLine(code);
+
+    const char* blockEnd = tic_core_script_config(code->tic)->blockEnd;
+    if (blockEnd != NULL)
+    {
+        newLine(code);
+        for(size_t i = 0; i < strlen(blockEnd); i++)
+            inputSymbol(code, blockEnd[i]);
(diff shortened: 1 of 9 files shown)


src/studio/editors/code.c
```diff
@@ -1157,6 +1157,19 @@ static void doTab(Code* code, bool shift, bool crtl)
     else inputSymbolBase(code, '\t');
 }
 
+// Add an "end" keyword, and put the cursor in the line between.
+static void newLineWithEnd(Code* code)
```
nesbox, line 1161: TIC-80 also supports other scripts langs like JS, Squirrel, Wren, Fennel, and Moonscript. Please add the function "end" depending on lang, you can define it in `tic_script_config` struct (for example `const char* end = tic_core_script_config(tic)->functionEnd;`).
Thank you.

nesbox (changes requested):

nesbox:
Any progress here?

blinry:
Good point about the multiple langs! I called the struct entry `blockEnd`, because it could also be used to close if statements and for loops, for example.

For Lua and Ruby, it's `end`, for JS, Squirrel and Wren, it's `}`, and Moon and Fennel just insert an indented line for now.

Ready for review again! :)

nesbox (approved):
Looks good, thank you 👍

PR #1723, opened 2021-12-06
Update to v0.90.1723 and disable dingux (WIP)

RobLoach:
This updates to v0.90.1723, brings in more architectures, and disables the dingux build.

**DO NOT MERGE**


RobLoach:
Oops, wrong repo. Ignore!

PR #1725, opened 2021-12-08
Fix OpenDingux builds

RobLoach:
This is a libretro downstream change from @jdgleaver to fix the OpenDingux builds of TIC-80.


nesbox (approved):
thank you 👍

PR #1734, opened 2021-12-12
Fix buffer overflow in onHelp_api

remi6397:
There are too many API functions for them to fit in the buffer 😳


joshgoebel:
Doesn't this just truncate? Should we instead (or in addition) also increase TICNAME_MAX?

nesbox:
I added a better fix here f029595 where I calculate the buffer size on compile-time and we are not depending on `TICNAME_MAX`.
Thanks.

PR #1745, opened 2021-12-16
When listing language options in the spec, use "x, y, or z" format

blinry:
I noticed that there was a space behind each language option, leading to an improper-looking list in the CODE section of `help spec`. Before:

screen10

After:

screen9

Also seems to fix the indentation, I didn't touch that. Implemented in the most straightforward way I could think of, does that seem alright?

Also removes the last "|..." in `help new`, as I don't think there are other options, right? Before:

screen12

After:

screen11


RobLoach (approved):
Seems fine.

nesbox (approved):
👍

PR #1753, opened 2021-12-18
On Linux, use Ctrl+L to clear the console, per default of other shells

### LINKED ISSUE #1741: Add Ctrl+L as a keybinding for clearing the console
This is another thing I tried to do intuitively, coming from a Linux background, where it's the default shortcut for "clear" in many shells.

Reading the source code, I saw that Ctrl+K is already a keybinding for that, which seems to be a MacOS thing. Would it hurt to also add a Ctrl+L keybinding? Very willing to implement that! :)

blinry:
Closes #1741.


nesbox (approved):
Thanks

Issue #1746, opened 2021-12-16
Add help entries for keys and buttons


blinry:
I very often look up functions and RAM addresses in the built-in help. The only thing I need to look up externally is the list of key and button IDs.

Would a feature that adds the "keys" and "buttons" help topics be welcome, which show tables from https://github.com/nesbox/TIC-80/wiki/Key-Map and https://github.com/nesbox/TIC-80/wiki/key?

joshgoebel:
Closing with #1754.

PR #1754, opened 2021-12-18
Add help topics for keys and buttons

blinry:
As per #1746, this is an implementation attempt.

screen13

The help for the keys is arranged in two columns to save some space:

screen14
screen15

Open questions:

- Should the topics be mentioned in the help texts for `btn(p)` and `key(p)`? I think probably not – for example, `poke` also doesn't mention the RAM and VRAM pages.
- Should the tables have headers, like RAM and VRAM?
- Should they be included when exporting the help?

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -2922,6 +2922,16 @@ static void onExport_help(Console* console, const char* param, const char* name,
         FOR(const ApiItem*, api, Api)
             ptr += sprintf(ptr, "\n### %s\n`%s`\n%s\n", api->name, api->def, api->help);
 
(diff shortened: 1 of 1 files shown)


nesbox:
> Should the topics be mentioned in the help texts for btn(p) and key(p)? I think probably not – for example, poke also doesn't mention the RAM and VRAM pages.
Should the tables have headers, like RAM and VRAM?

It's up to you :)

> Should they be included when exporting the help?

I'd like to see it in the exported help :)

Thanks

blinry:
I added them to the exported help, and I'm happy with this PR as-is! :)

nesbox (approved):
Thanks

Issue #1771, opened 2022-01-01
Do we want to "own" the per-language WASM bindings/libs?


joshgoebel:
I've created the first, for Zig: https://github.com/joshgoebel/zig_bunnymark/blob/main/src/tic80.zig

Related: #1259 

I imagine people will over time show up to write bindings for all the possible languages - look at the WASM-4 project for a list of what's already possible to compile to WASM pretty simply:

- C/C++
- D
- C# (in progress)
- Nelua
- Go
- Nim
- Odin
- Rust
- Wat
- Zig

WASM-4 "owns" the bindings (and build scripts) for each of these in it's own repo: https://github.com/aduros/wasm4/tree/main/cli/assets/templates

Although because of this they also kind of keep the API to "least common denominator" and the functions are all very C like - even when the languages might allow for much nicer APIs... Personally I'd hope we could be a little more flexible in that regard, but it does have concerns like documentation, etc... ie, the "original" API looks one way while the Zig or Odin API has "more niceness".  Perhaps we could support both "raw" and nicer... Mostly I'm imagining useful "aliases" (such as `noclip()`, if we don't decide to add it more broadly) and named arguments:

```zig
// behind the scenes calls clip(-1,-1,-1,-1)
tic.noclip() 
// named arguments via a struct
tic.print("hello world", x, y, .{ scale: 3, fixed: false, smallfont: true, color: 4 })
```

nesbox:
yes, we want it :)

joshgoebel:
Where in the source tree should we place such things?

- examples?
- projects
- templates?
- examples/templates?

nesbox:
I'd select something like `src/api/wasm/[c d go nim odin ...]`, but anyway it's your choice.

joshgoebel:
Closing this as asked and answered and continuing with #1784.

PR #1772, opened 2022-01-01
WASM runtime, BINARY chunk support

### LINKED ISSUE #1259: Adding WebAssembly runtime support via BINARY chunks
hi there! is there any interest in adding WebAssembly as a supported "language"? I've been experimenting with [this wasm interpreter](https://github.com/wasm3/wasm3), it can run binary wasm file by interpreting the bytecode. It's pretty easy to attach custom api functions and claims to be pretty fast.
Obviously it won't be editable inside tic-80 editor (we could show a base64 encoding or something) but it would open up a lot of possibilities, like making games in any language that targets WebAssembly.

wasm3 looks like a well written software, doesn't have dependencies and is standard c, should compile on most platforms. 

i'd like to give it a try eventually :)

### LINKED ISSUE #1693: Request: BINARY chunk, support external compilers
Related #1007.

Could we add a BINARY chunk and external compiler support?    The intention is to support fantasy (or not) emulated CPUs and to allow usage of external (quite complex) compiler toolchains (Assembly, Rust, C, etc) to produce the compiled machine code.

See #1694 for further background.  Of course this wouldn't apply to just 68000.  It would apply to any language that compiles down to machine code / bytecode (And this isn't suitable for storage in the CODE chunk - because the code itself might be stored there).

This also paves the road for distributing cartridges/games that no longer contain the source they were build from.

**How it might work with cartridges:**

```
tic80 --cmd "load cart.tic & import binary cart.bin & save"
```

Would take the data from `build.bin` and load it into the BINARY chunk of the cartridge.

**How it might work with project:**

The binary data would simply be included as any other binary data is:

```
-- <BINARY>
001:eccccccccc888888caaaaaaaca888888cacccccccacc0ccccacc0ccccacc0ccc
002:ccccceee8888cceeaaaa0cee888a0ceeccca0ccc0cca0c0c0cca0c0c0cca0c0c
003:eccccccccc888888caaaaaaaca888888cacccccccacccccccacc0ccccacc0ccc
...
```

When working on a project with an external build toolchain it would be the build processes job to compile the source in the project source file, and then re-write the source file so that than `BINARY` section contained the latest release.  For easy development this would be done with some type of file watch process most likely.

Would could also add a `BINARY_ZIP` chunk for parity (though I think this is a bad idea - the amount of binary storage "ROM" should be a fixed amount, not subject to random size issues depending on compressibility)... obviously for distribution you could imagine a cartridge with only a BINARY chunk and no CODE chunk...

@nesbox Thoughts?

### LINKED ISSUE #1776: WASM demo cartridge
Something slightly different needs to be done here... My first pass at just the "above the fold"...

Screen Shot 2022-01-02 at 1 50 32 PM

I think for WASM we need to tell the user what they've gotten themselves into and then they could go to the Wiki or look at the sample starter projects in various languages to learn more.   I went with the `--` comments from Lua for their hopeful familiarity since Lua is our original language.

This "in cart help" could be as short or long as we wanted but it may be best to keep it brief and then direct them elsewhere, to somewhere like the wiki where the content is easier to maintain.

Thoughts?

### LINKED ISSUE #1777: Making TIC RAM a pointer so that runtimes can provide their own RAM
I'm frustrated to have gotten so far and then realized the memory is more of an issue than I thought... WASM3 (by default) wants to "own" it's own memory... so the simplest thing would be to let it do that and then hand that to TIC-80 and say "hey, here is the memory to use for the machine"... if a runtimes doesn't need that,  you could provide the default allocation yourself (as is done now).  This has the side benefit of allowing some runtimes to have more RAM than others, depending on their needs.

I actually though maybe I could handle this myself inside `wasm.c` since all the API functions take `tic_mem*`, but it's a bit misleading... several of them cast it back to `tic_core` to get access to `state` and other internal pieces of TIC... not to mention the other internal pieces which would still be looking at the "original" RAM.

So then the question is could we perhaps just replace the WASM3 ram with TIC-80s RAM.  This looks more doable since it appears to purposely be a pointer... but we require a minimum of 128kb for WASM3 (WASM spec works in 64kb chunks)... so would you just want to change the entire TIC-80 to 128kb?  Then what if we decide for "fairness" with interpreted languages that WASM needs 256kb?  Then see the ongoing discussion in #1259 where some think the memory limit should be MUCH higher - to match the same "memory limit" of the other runtimes, which is effectively uncapped.  I'm not sure we'd want to share that same extra allocation with the runtimes though - which have their own RAM already - although I'm also not sure it would hurt - although it does increase the overall memory demand if we just have a lot of allocated but unused RAM just laying around.

So that's lead up to:

_Would it be possible if anywhere we referenced `tic_ram` it was via a pointer?_  This would allow the WASM runtime to replace the internal RAM with it's own at `init` and then swap it back at `close`...

And a related question:

_Also, why does pausing copy the RAM and state into a whole other data structure?_  Is something bad going to happen to the original copy during the pause?

### LINKED ISSUE #1778: Low-level API concerns (WASM, etc)
Much of the TIC-80 API is potentially problematic because of default values/variable args.  (ie our scripting languages allow passing a variable number of arguments, and in WASM that is not easily possible)

In other places the API is problematic because using `-1` as a "default placeholder" would mix types, such as when the value is supposed to be a boolean. Ie, in Zig for example there is no way a function can return an integer in one place and a boolean in another... it's one or the other, those are not compatible types.

- btn() - the default form that returns a numeric, the `btn(x)` variant returns a boolean
- font (take booleans)
- print (take booleans)

Or you have cases when the input CAN be negative but the value is also optional:

- map - the `sx` and `sy` arguments, etc

Many others take only positive numeric arguments and I think using `-1` could work there such as:

- cls
- clip
- etc
- sfx
- pix
- etc

---

_Originally posted by @msx80 in https://github.com/nesbox/TIC-80/issues/1259#issuecomment-1004032808_ :

> I feel like the "simple just works" TIC approach boiled down to WASM might would be -1 for "i dunno" or "just do the right thing"... ie, no one would need to know the defaults to call the API - that would be all handled INSIDE TIC, not WASM. So to call sync(), effectively sync(0,0,false) you would need to know that -1 was "default" and the number of arguments (to push the right number of args onto the stack).

So the problem with how to map the API to simpler languages (like C or wasm) is tougth. Maybe we could do a separated issue for this?

Anyway, i think there are two ways: one where the defaults are implemented in the language and one were they're implemented in TIC. I also think the second one offer more value (you avoid repeating all the defaults values), but it's a bit harder to implements. It would require changes to the tic80 layer (which is not much trouble).

The most delicate thing would be to choose a value to communicate "default" to tic80. You propose -1 which works for most cases but what if -1 is also a valid option for the parameter? I think tic80 accepts negative values for coordinates in api. In "map" you can pass negatives to sx and sy, -1 would be ambiguous as the default is 0.

PS we already have some kind of "c api" in [api.h](https://github.com/nesbox/TIC-80/blob/master/src/api.h). I think it's sensible to use it (or at least take it into consideration)

### LINKED ISSUE #1780: WASM storage vs RAM philosophy... equivalency with other runtimes or locked down tight?
Continued from #1777. Related #1259.

Probably should have started this first, but I guess I assumed that WASM was "different" would be clear and some decisions easy, but this RAM pointer discussion has me a little concerned when we keep talking about the RAM allocation as if it should be static.  **This is not true of any of the other runtimes.**

> None of the other runtimes statically allocate their runtime RAM, I'm not sure why WASM should either... if Lua needs 256kb while it runs, then when you stop it and switch VMs (to say a JS cartridge), that memory is released. If WASM needs 256kb while it runs, why do you want that 256kb wasted all the time, even if the user never plays a WASM cartridge?
  
> Also WASM (like the other runtimes) allows dynamic memory allocation. It's possible that one cartridge might needs 128kb and another needs 256kb... that's something we could easily support (with my approach) - like we support with every other runtime. But not if the allocation is static.

So two approaches have been suggested here:

- We lock down WASM (more like WASM-4 does).  Tiny ROM, tiny RAM.  This would make it VERY different from the other runtimes which have unlimited RAM.  This would be the perspective that TIC-80s "unlimited" RAM nature is bad and this might be the first step in course correction.
- We do our best to make WASM equivalent to the other runtimes. (more like classic TIC-80)  That means if Lua can request 512kb of RAM, then WASM should be able to do the same.  IE: You shouldn't be able to write a game in Lua that's impossible to write in WASM solely because we artificially limited WASM's RAM.

I see benefits to both approaches.  But I don't think we should lock WASM down UNLESS that's part of a serious, larger strategy to lock the other runtimes down more in the future.  IE, WASM shouldn't be crippled for no reason.

joshgoebel:
Closes #1259.  Closes #1693. Closes #1776. Closes #1777.  Closes #1778.  Closes #1780.

Building on the great work already done by @msx80.  What I've done here:

- rebased the existing work onto the latest TIC-80 codebase
- upgrading WASM3 to newer version
- flushed out the API almost fully (though there are some questions)
- implemented Bunnymark in Zig/WASM
- adds `import binary` and `export binary` commands
- wrote Zig language bindings for the TIC-80 WASM API
- `wat2wasm` stuff has been removed, could be added back as `script: wat` in the future

Unanswered questions (see #1259) / TODOS

- [x] What is the max size of a WASM binary? **256kb.** 
   #1780
- [x] How much memory should the runtime have access to? Currently 256kb.  128kb wasn't enough to even run Bunnies benchmark (1200 sprite limit) since now all 96kb is "reserved"...
- [x] Can core.memory.ram become a pointer?  **For now, may be changed later.**
   #1777
- [x] What exactly should the editor do (other than provide the magic `--script: wasm` comment...  **It provides some simple docs, license, README, etc..**
  #1775
- [ ] We may still need to further tweak some of the API functions specifically for WASM.
   #1778 
- [x] What to do about "projects"... I'm not sure they fit with the WASM workflow. **Binary chunks are currently not saved in projects.**
    #1775
- [x] BINARY chunk should be saved in multi-chunks if over 64kb, same as CODE
- [x] WASM demo cartridge 
  #1776 
- [x] move WASM bunny bench into `demo.c`, etc
- [ ] can we support `remap` callback on `map`? This one may be over my head.  WASM3 author thinks technically yes, but it sounds like a chore - and WASM3 doesn't provide the support pieces yet.
- [ ] Add BOOT API callback 
   #1773 
- [x] Figure out where to put per-languages TIC header libs
   #1771 

You can find the Zig project in `examples/zig_bunnymark`.  You'll need a very recent version of Zig to build it... someone is welcome to come in here and write a C/C++ version and library wrapper.  I'd recommend you use the same Makefile's that WASM4 is using and it should be simple.


joshgoebel:
Doesn't seem to build Windows... an issue with WABT?

```
D:\a\TIC-80\TIC-80\vendor\wabt\src\config.h(59,10): fatal error C1083: Cannot open include file: 'alloca.h': No such file or directory [D:\a\TIC-80\TIC-80\build\wabt.vcxproj]
1661
  stream.cc
1662
D:\a\TIC-80\TIC-80\vendor\wabt\src\config.h(59,10): fatal error C1083: Cannot open include file: 'alloca.h': No such file or directory [D:\a\TIC-80\TIC-80\build\wabt.vcxproj]
```

If we're agreed that `wat` doesn't need to be a part of this first pass I can rip this out for now... or if someone else could jump in and help out?

nesbox:
I agree, `wat` can be ripped out for now :)

joshgoebel:
```
In file included from /__w/TIC-80/TIC-80/vendor/wasm3/source/m3_exec_defs.h:11,
874
                 from /__w/TIC-80/TIC-80/src/api/wasm.c:36:
875
/__w/TIC-80/TIC-80/vendor/wasm3/source/m3_core.h:36:25: error: conflicting types for 'u32'
876
   36 | typedef uint32_t        u32;
877
      |                         ^~~
878
In file included from /__w/TIC-80/TIC-80/include/tic80.h:26,
879
                 from /__w/TIC-80/TIC-80/src/tic.h:25,
880
                 from /__w/TIC-80/TIC-80/src/api.h:25,
881
                 from /__w/TIC-80/TIC-80/src/core/core.h:25,
882
                 from /__w/TIC-80/TIC-80/src/api/wasm.c:23:
883
/__w/TIC-80/TIC-80/include/tic80_types.h:31:22: note: previous declaration of 'u32' was here
884
   31 | typedef unsigned int u32;
885
      |                      ^~~
```

Only the Raspberry baremetail build is failing now because on that platform evidently `uint32` and `unsigned int` are not the same... and WASM3 also aliases `u32`... Is there same way to be more explicit on Raspberry PI?  I'm not sure why we need these conditionals in `tic80_types.h`?

Help?

keithohara:
This looks awesome!

How do you run the zig example?

This didn't work:

```
 ../../build/bin/tic80 bench.tic
[string ";; title:  game title..."]:1: function arguments expected near 'title'
```

joshgoebel:
Are you running the branch with this specific change?  This is not released anywhere, and it's not merged.

You'll need `--` comments, let me update.

keithohara:
> Are you running the branch with this specific change? This is not released anywhere, and it's not merged.
> 
> You'll need `--` comments, let me update.

Pretty sure I am:

```
 git branch -a
* wasm_support
```

joshgoebel:
That should fix it... I rebased onto master and @nesbox changed the memory map out from underneath me... grrrrr. :-p  So now I've added 1024 bytes padding at the end until we sort that out.

@nesbox Once we start actually using the free RAM for compiled software we can't just add it randomly - that would break all compiled cartridges.  If that memory at the end is truly reserved instead of "free" then we need to document that.  

Do you want to just reserve the entire 12kb for your own use or should we just flag off another 1-2KB that's reserved and "do not touch"?

keithohara:
> That should fix it... I rebased onto master and @nesbox changed the memory map out from underneath me... grrrrr. :-p So now I've added 1024 bytes padding at the end until we sort that out.

Thanks, that did the trick!

joshgoebel:
Let me know if you have any questions!  I'm on Discord also.

joshgoebel:
Done:

https://github.com/nesbox/TIC-80/wiki/RAM

But now that means that no cartridges (other than WASM) have access to any actual device memory at all... that's a change... should we bump the core memory from 96kb to 128kb -giving users 32kb back for hardware hacking in non-WASM runtimes?  I've used that memory before for temporarily copying portions of the screen buffer and such things.  There are nice tricks you can do with this and bank switching... but if you take away the free RAM you leave no one anywhere to `memcpy`.  

Thoughts?

nesbox:
This memory was never free, it was always assumed that sooner or later it will be occupied as it develops, let's leave it as is.

joshgoebel:
Once this is ready is merging it as experimental (behind a build flag that's disabled by default) reasonable? I'd love to allow a few people to build and test all (after merging) the API bits without also (yet) being locked into binary compatible for life... ie if we need to change a few things early on I'd like to do it without fear of breaking published cartridges in the wild.

TLDR: _Can we merge it but also label it "beta" and still subject to change slightly?_

src/core/core.c
```diff
@@ -444,6 +447,13 @@ void tic_core_tick(tic_mem* tic, tic_tick_data* data)
 
             data->start = data->counter(core->data->data);
 
+            // TODO: does where to fetch code from need to be a config option so this isn't hard
+            // coded for just a single langage? perhaps change it later when we have a second script
+            // engine that uses BINARY?
+            if (strcmp(config->name,"wasm")==0) {
```
joshgoebel, line 453: Or we could move this into the `config` struct, `usesBinaryVsCode` or some such, but perhaps that's premature...

nesbox:
I tried to build it and got this screen
image

Something is broken, possibly the problem is somewhere in the cart loading.

joshgoebel:
What exactly did you do to get that screen?

nesbox:
Nothing special, just compiled the `wasm_support` branch and ran.

joshgoebel:
Hmm.... I pushed my latest rebase so we're using the exact same commits, but mine has been building and running without issue all along...

joshgoebel:
Can you pul the latest and try again the commit number showing in your build is quite a while back...

nesbox:
I see that CMake is changing `config.tic.dat` at compile time and I think that's the reason I got this black screen.
image

nesbox:
I did some refactoring with the Studio here 3dc2feacbeb2f1db8e7184eb548d381d7e690efc which touched a lot of project files, pls don't worry, your PR can be easily merged with it 😟

joshgoebel:
> I see that CMake is changing config.tic.dat at compile time and I think that's the reason I got this black screen.

I'm assuming that's your new refactoring things?  I haven't made any changes to CMakeList that should touch config.tic.dat that I'm aware of...?   If you need me to make any changes or do any specific work here, please let me know.

nesbox:
Finally, I merged the PR with the `main` branch here [wasm-merge](https://github.com/nesbox/TIC-80/tree/wasm-merge).
I decided to split the cart builders to `prj2cart` and `wasmp2cart`, also I fixed an error, instead `fread(buffer, size, 1, file);` should be `fread(buffer, *size, 1, file);`, after that everything worked and finally I saw a normally rendered TIC-80 screen.
Seems all work except barematal build, don't know how to fix it properly.
image
https://github.com/nesbox/TIC-80/runs/5289437702?check_suite_focus=true#step:6:876

joshgoebel:
> I decided to split the cart builders to prj2cart and wasmp2cart

Yeah, if it was up to me tiny helpers like this would be in a more accessible language, not C, so I'm glad you decided to rejigger that stuff...

Does Raspberry pi really not define `uint32_t`, that seems strange...?

nesbox:
I fixed RPI build with the patch, don't see another better way.
So, seems we are ready to merge the PR, there are two ways to do it, I can merge [wasm-merge](https://github.com/nesbox/TIC-80/tree/wasm-merge) branch directly to the `main` and close the PR or we could merge it with your `wasm_support` fork and then properly push it to the main repo.
What do you think?

nesbox:
Ok, I merged it directly to the `main` aeabb55bef63a11014107d2d0a369ac7ec324530.
Pls do something with the PR.

PR #1786, opened 2022-01-06
Create SECURITY.md

joshgoebel:
I went to see what our security policy was, and we didn't have one, so this is a beginning.  I just used the default template and filled in the very basics.

Not sure if you'd prefer a nicer email like `security@tic80.com`?  

Also is there any reason someone should still be running 0.80 or 0.70?  Could we mark anything below 0.90 as obsolete?  This also has wiki implications as we don't need to mention "added in version 0.2" if someone isn't using version 0.1 anymore, etc...

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,23 @@
+# Security Policy
+
+## Supported Versions
+
+Only the latest releases are supported and may receive security updates.
+If you aren't using the latest stable version, please upgrade.
+
+| Version    | Status   | Supported          |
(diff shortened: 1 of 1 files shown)


nesbox:
> Not sure if you'd prefer a nicer email like security@tic80.com?

I'd prefer, but the mail server isn't configured yet, let's change it later.

> Also is there any reason someone should still be running 0.80 or 0.70?

Yes, I always see people who don't want to update for an unknown reason.

nesbox (approved):

PR #1787, opened 2022-01-07
fix: prevent cartridges hijacking keyboard events

### LINKED ISSUE #1785: Hacking KEYBOARD registers in "RAM" can affect input the IDE sees
_It seems one can trivially break the IDE from within the VM by writing to the KEYBOARD registers._ (even if they can't trigger a keypress it seems they could perhaps trigger a key release -  therefore making the IDE response to MULTIPLE keypress events when a key was only pressed a single time)

**Confirmed**  I'm not sure how large this exploit is but it's definitely possible to hack/break the keyboard state from inside the game and break the IDE.  _This should not be possible._

**API affected**

Anything that can write to the KEYBOARD registers in RAM:

- poke
- memset
- memcpy

**Examples**

Trivial example:

```lua
function TIC()
  poke(0x0FF88, 66)
end
```

This simulates ESC being constantly held (never released) and prevents you from exiting a game - you have to quit TIC-80 completely... I'm not sure if other exploits are possible (such as injecting actual keypress rather than preventing key releases)...  IMHO the keyboard handling of the "kernel/IDE" should not depend at all on the state of the fantasy computer's memory registers (that can be hacked by the game) -  ie, keyboard handling at the IDE level should never reference `tic->ram`.

You can also trigger key repeats easily (the original issue I filed):

```lua
function TIC()
	poke(0x0FF88, 0)
	poke(0x0FF89, 0)
	poke(0x0FF90, 0)
	poke(0x0FF91, 0)
end
```

Now someone hitting run, save, etc... that keypress will be triggered multiple times, perhaps 5 or 6 (even for a short keypress) based on tricking the IDE into thinking the key was released, then pressed multiple times.

---

_Older description before we narrowed this down:_

**Observed behavior:**

Holding down CMD-R (even very briefly) will cause the cartridge to be started 5-6 times... continuing to hold it down will cause the cartridge to be started over and over indefinitely.

**Expectation:**

Hitting CMD-R should start the cartridge once, regardless of how long it's depressed.  Ie, we want `btnp` here, not `btn`. 😁

joshgoebel:
Resolves #1785.

This fix captures the current (`input.ram.keyboard`) state in
`tick_start` and immediately restores it in `tick_end`. This
prevents cartridges from injecting corrupt "previous" states
into the keyboard state engine that could be used to hijack
keyboard input.

When we trust `input.ram.keyboard` after user code has run
we open ourselves up to a cartridge:

- repeating keyboard events by fake "lifting" a key
  (which is still held by the user, then registering as
  separate repeating keypress/releases until the user
  finally releases the key)
- preventing key up events by fake "holding" a key
  (this prevents the state engine from ever seeing the
  key release).

---

Originally this was a lot larger, but I figured it was going to be too hard to review and certify with just a glance.   This takes your "don't let someone modify KEYBOARD" idea and simplifies it to preserving rather than protecting.  We preserve the "now" state during a tick and then restore it immediately afterwards, vs trusting whatever (potentially incorrect) state could be present in the RAM.

Since the real problem is making sure the integrity of KEYBOARD state is guaranteed for "system" bits, this does so by making sure that immediate after `pre-tick` the "now" state is restored, guaranteeing that:

- historic keyboard state is never corrupted in the first place (fixing the bug)
- KEYBOARD state is guaranteed to be accurate immediately pre-tick because we haven't run any user code yet that could alter it

I hope this is clear.  If you have any questions, ask.

---

I'm still open to doing a larger refactor here, but I thought this was pretty clean.  The only gotcha is that code like `handleShortcuts()` must be run before the user code, and I've added a comment to that effect:

- pretick
- Studio handles keystrokes (KEYBOARD is guaranteed correct here)
- tick (user code)
- post tick

### DIFF
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -101,6 +101,7 @@ typedef struct
     struct
     {
         tic80_gamepads previous;
+        tic80_gamepads now;
 
         u32 holds[sizeof(tic80_gamepads) * BITS_IN_BYTE];
     } gamepads;
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -108,6 +108,7 @@ typedef struct
     struct 
     {
         tic80_keyboard previous;
+        tic80_keyboard now;
 
         u32 holds[tic_keys_count];
     } keyboard;
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -294,7 +294,18 @@ void tic_api_reset(tic_mem* memory)
 {
     tic_core* core = (tic_core*)memory;
 
+    // keyboard state is critical and must be preserved across API resets.
+    // Often `tic_api_reset` is called to effect transitions between modes
+    // yet we still need to know when the key WAS pressed after the
+    // transition - to prevent it from counting as a second keypress.
+    //
+    // So why presev `now` not `previous`?  this is most often called in
+    // the middle of a tick... so we preserve now, which during `tick_end`
+    // is copied to previous. This duplicates the prior behavior of
+    // `ram.input.keyboard` (which existing outside `state`).
+    u32 kb_now = core->state.keyboard.now.data;
     ZEROMEM(core->state);
+    core->state.keyboard.now.data = kb_now;
     tic_api_clip(memory, 0, 0, TIC80_WIDTH, TIC80_HEIGHT);
 
     resetVbank(memory);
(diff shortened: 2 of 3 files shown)


src/core/core.h
```diff
@@ -108,6 +108,7 @@ typedef struct
     struct 
     {
         tic80_keyboard previous;
+        tic80_keyboard now;
```
nesbox, line 111: I just realized we should do the same for the `tic80_gamepads previous;` data

joshgoebel: I didn't think the gamepad could be used for studio control/input, so it's not a security risk, but it's easy enough to add to this PR if you like.  I'll do it.

nesbox: Gamepad used in `Surf` and `Game Menu`, so you also can hack them :)
Pls add the same for the gamepad too.
Thanks.

joshgoebel:
You'll have to test, I don't have gamepads.  But patch is simply enough.

nesbox (approved):
Thank you

PR #1789, opened 2022-01-07
Console: `cd` with no args returns home

### LINKED ISSUE #1742: When entering "cd" without arguments, return to the base directory
This is another thing I expected to work intuitively, because it's the default for the `cd` command in Linux shells to return to the home directory in that case.

Looking at the help of the TIC-80 "cd" command, I found that `cd /` already does that. Would it be a welcome addition to trigger the same behavior for `cd` without arguments?

joshgoebel:
Resolves #1742.


nesbox (approved):

nesbox:
Thanks

PR #1790, opened 2022-01-07
add default license, version, site tags to demos

### LINKED ISSUE #387: License for cartridges.
It would be nice to see from a cartridge if it can be modified and redistributed and on what conditions (the software license).

Also it would be nice to have some manifest in the cartridge with a link e.g. to github project and contact to the author.

I am wondering if I can modify existing game to create a split screen version. Currently the only way is to comment under the game on web page.

joshgoebel:
Resolves #387.

Once we agree on the template I'll modify all the demos (not just Lua)...  does this look good?

### DIFF
--- a/demos/luademo.lua
+++ b/demos/luademo.lua
@@ -1,7 +1,10 @@
--- title:  game title
--- author: game developer
--- desc:   short description
--- script: lua
+-- title:   game title
+-- author:  game developer, email, etc.
+-- desc:    short description
+-- site:    website link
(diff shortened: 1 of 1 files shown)


joshgoebel:
I went with "MIT License (change this to your license of choice)" vs just "license of choice" because technically the demo code is licensed... but that might not really be relevant so that could be shortened if necessary.

nesbox (approved):

PR #1791, opened 2022-01-07
support CTRL-A, CTRL-E console editing keystrokes

joshgoebel:
I try to do this all the time, so it should probably work. :)


nesbox (approved):
ok, thanks :)

Issue #1739, opened 2021-12-15
Anbernic RG351M with ArkOS - most games lag terribly


Mr-Bajs:
I just found out about TIC-80 so i downloaded a few games from https://tic80.com added it into the folder on the SD cards. But most games just studders terribly or slow motion.

From maybe 10 games that I tried only Witch 'em Up and Cauliflower Power runs smooth.

v. 0.90.1748

nesbox:
It mostly depends on how game devs use the TIC resources, don't know what we could do here.
Also, please try to test the nightly build, it should run a bit faster https://nightly.link/nesbox/TIC-80/workflows/build/master.
Thanks

desttinghim:
I am having the same experience, running TIC-80 through retroarch. The biggest problem so far has been with sprites - in bunnymark I get roughly 75 before the music starts to break. My uneducated suspicion is that each sprite issues its own draw call which quickly tanks performance.

version: 0.90.1748

desttinghim:
@nesbox I am using the retroarch tic80 core on the anbernic, but I don't see the core in the nightly releases, or instructions on how to build the libretro core for arm. How would I test the nightly build?

nesbox:
Maybe @RobLoach can answer the question :) He works with the retroarch build

RobLoach:
Is that build through Lakka? How are you getting the tic-80 core? The hardware is pretty low, and 75 sprites is a pretty good amount 😋

desttinghim:
I'm running Arkos. I got the tic80 core through retroarch. I feel like the hardware could do more than 75 sprites, but I could be wrong.

joshgoebel:
It is, but the level of indirection in our drawing code is... Let's take scale!=1 (bunny mark), which is worse case I think:

- tic_api_spr()
  -  drawSprite()
    - drawTile()
      - loop x,y
        - rotate and orientation x, y swaps
        - tic_tilesheet_gettilepix()
        - palette[] lookup of color
        - drawRect()
          - loop
            - drawHLine
              - loop
                - tic_api_poke4
                  - tic_tool_poke4

This is likely easily 10-20x slower than what it would take to just move the pixels onto the raw hardware. But for all the flexibility we offer there is a reason for all of these abstraction layers.

[...]

PR #1794, opened 2022-01-08
(enh) speed up poke4 API hot path

joshgoebel:
This is a hot path and used by almost all our internal drawing code so
it gains a real 13% performance boost from not going thru the same
`tic_api_poke` abstraction as the other poke* API calls do.

Related: #1739.

All of these should really be `u32` in my mind (and then we avoid the `<` comparison... but I wasn't seeing a measurable speed difference so I tried to go for "smallest change".  I did change the order of the `switch` in drawTile, but that could just be my imagination.  It doesn't hurt though.

_I can make them all u32 for consistency if you'd like, just let me know._

Before:

Screen Shot 2022-01-08 at 10 52 20 AM

After:

Screen Shot 2022-01-08 at 10 59 50 AM

Here is a run showing 17.7% increase in rendering - 13-14% was more typical in my testing. We see the uptick in almost all the draw functions because they all route thru `poke4`.

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -481,7 +481,7 @@ enum
         2,                                                                                                              \
         0,                                                                                                              \
         void,                                                                                                           \
-        tic_mem*, s32 address, u8 value)                                                                                \
+        tic_mem*, u32 address, u8 value)                                                                                \
                                                                                                                         \
                                                                                                                         \
     macro(memcpy,                                                                                                       \
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -108,9 +108,14 @@ void tic_api_poke2(tic_mem* memory, s32 address, u8 value)
     tic_api_poke(memory, address, value, 2);
 }
 
-void tic_api_poke4(tic_mem* memory, s32 address, u8 value)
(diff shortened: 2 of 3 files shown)


joshgoebel:
I'm sure you could get similar with with poke1 and poke2 (removing the code resuse) but don't think it's worth bothering since they aren't hot paths for most games.

nesbox:
Why are you testing the debug build without any compiler optimizations?
I've tested the changes on the release and I don't see any boost, almost the same result in both cases
Before
image
After
image

Could you pls build with `-DCMAKE_BUILD_TYPE=MinSizeRel` CMake option and test the release on your side?

joshgoebel:
> Why are you testing the debug build without any compiler optimizations?

🤦🏻‍♂️

Cause I do a lot of debugging?  Sorry.  I'll test again and if there is nothing here I'll close.  It is easy to believe the compiler somehow optimizes this away, so you're probably right.

joshgoebel:
If there is something here it's within the margin of error now.  Closing.  Sorry, again for testing debug build. 🤦🏻‍♂️

PR #1800, opened 2022-01-09
(enh) moonscript supports eval now

joshgoebel:


src/api/moonscript.c
```diff
@@ -91,6 +108,9 @@ static bool initMoonscript(tic_mem* tic, const char* code)
             return false;
         }
 
+        lua_setglobal(lua, "_ms_loadstring");
+        lua_getglobal(lua, "_ms_loadstring");
```
joshgoebel, line 112: Dunno if there is a better way to do this... I was looking for `lua_tee` or something, but couldn't find.

nesbox: Honestly, I don't understand what's going on here :)
Also, could you pls move the `"_ms_loadstring"` to `static const char _ms_loadstring[] = "_ms_loadstring"` variable?

joshgoebel: Well, Lua is stack based, so setting a global removes it from the stack so you have to put it back on the stack... I was looking for a "duplication item on top of stack" API, but didn't see it.

nesbox (approved):
Thank you

PR #1801, opened 2022-01-09
faster drawTile, remove repeat clipping checks

### LINKED ISSUE #1796: Doesn't drawTile already check clipping itself?
In drawTile:

```
    if (scale == 1) {
        // the most common path
        s32 sx, sy, ex, ey;
        sx = core->state.clip.l - x; if (sx < 0) sx = 0;
        sy = core->state.clip.t - y; if (sy < 0) sy = 0;
        ex = core->state.clip.r - x; if (ex > TIC_SPRITESIZE) ex = TIC_SPRITESIZE;
        ey = core->state.clip.b - y; if (ey > TIC_SPRITESIZE) ey = TIC_SPRITESIZE;
```

It appears to be taking clipping into account, yet in `setPixel` *for every single pixel* we check the clipping individually... it seems this is not necessary.  If I replace the drawTile pixel paints with `setPixelFast` (not clip checked) I see the official benchmark (modified to remove square root noise) go from:

- before: 31500
- after: 32900

About a 4.4% win... for a hot path.  This is release, min size.  I think this is real win because I can't really see how the compiler could optimize this check away in any case.  I modified the demo cart to scale=1 and moved the guy around and he seems to clip just fine with no issues.

Thoughts?  Should I push PR?

joshgoebel:
Resolves #1796 .

`drawTile` appears to be taking clipping into account, yet in
setPixel for every single pixel we check the clipping
individually... it seems this is not necessary.

This adds a new `setPixelFast` that avoids the repeat checks.

I see about a 4-5% increase in benchmarks for just sprite draws.

### DIFF
--- a/src/core/draw.c
+++ b/src/core/draw.c
@@ -67,6 +67,15 @@ static void setPixel(tic_core* core, s32 x, s32 y, u8 color)
     tic_api_poke4((tic_mem*)core, y * TIC80_WIDTH + x, color);
 }
 
+static void setPixelFast(tic_core* core, s32 x, s32 y, u8 color)
+{
+    const tic_vram* vram = &core->memory.ram.vram;
+
+    // does not do any CLIP checking, the caller needs to do that first
+
+    tic_api_poke4((tic_mem*)core, y * TIC80_WIDTH + x, color);
+}
+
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Thanks

PR #1804, opened 2022-01-09
order langauges by alphabetical order

joshgoebel:
If we aren't playing favorites anymore then this order really should be alphabetical, not historic.

### DIFF
--- a/src/core/languages.c
+++ b/src/core/languages.c
@@ -31,33 +31,35 @@ extern tic_script_config WrenSyntaxConfig;
 
 tic_script_config* Languages[] = {
 
-	#if defined (TIC_BUILD_WITH_LUA)
-	&LuaSyntaxConfig,
-	#endif
+    #if defined(TIC_BUILD_WITH_FENNEL)
+    &FennelSyntaxConfig,
+    #endif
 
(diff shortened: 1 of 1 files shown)


nesbox (approved):

remi6397:
This made fennel the default language. :grin::v:

joshgoebel:
Default in what sense? New does not have a default anymore.

remi6397:
Not the language, I mean the demo that's loaded by default, it used to be the Lua hello. Maybe we should randomize it? :smile:

joshgoebel:
Oops, that was unintentional.  @nesbox Should we go back and hard-wire Lua as the starter language when the app first opens?

nesbox:
> Should we go back and hard-wire Lua as the starter language when the app first opens?

of course, Lua has to be the first, also I need to be more careful when reviewing PRs

nesbox:
reverted here 7364cbbad4cfee4897a5deb8ae00700fad51d0f0

joshgoebel:
Could we have just have fixed the default issue without reverting the ordering?

nesbox:
Please make new PR with the fixed ordering and Lua by default, I don't see other variants.
Sorry for the confusion.

PR #1813, opened 2022-01-16
split out main menu code from studio

joshgoebel:
This is the very beginning of my work to try and split the codebase up into smaller, more modular pieces.  It follows our discussion about privacy and studio not having full access to `core` (even internally), etc... it seems to then follow that the menu menu shouldn't have full access to the huge `impl` struct that is essentially our entire application (that currently resides in studio)... also studio has tended to be a bit of a "catch all" file,  so pulling out functionality into separate files would be good thing on it's own I feel.

This removes all menu menu related code from `studio`.  It leaves all state management code in studio (`impl.mode`)... the things that main menu still needs access to:

- tic_mem (pointer passed in `init`)
- config (pointer passed in `init`) - this would not be necessary if we moved `rwConfig` to `studio` and just allowed access to that for consumers.  We could also have headers for different privilege levels... so that you'd only include `writes_config.h` for things that needed `r/w` access to the config, etc.
- menu, game_menu (pointers passed in `init`)
- impl.gamepads (still extern `impl` for now)

For `impl.gamepads` we could copy a pointer as well or add an API for that...  The `mem` access feels like something we could hide behind an API in the future (it's only used to get `script: menu` and for the menu callback), but I was trying to take things one step at a time.

Splitting out `studio_impl.h ` was necessary to avoid a circular dependency in the header files.

The code in `main_menu.c` is 99% cut and paste from studio with just glue needed at the edges (`initMainMenu`, etc.) and to make the references to our internal state rather than the global studio `impl` object.

---

Commits:

- slightly nicer start up stages
- fewer allocations is better
- Separate main menu from studio
- isolate menu menu from studio a bit more
- it is main menu really, not game menu


src/system/sdl/main.c
```diff
@@ -1617,9 +1617,11 @@ static void emsGpuTick()
 
 static s32 start(s32 argc, char **argv, const char* folder)
 {
-    SDL_Init(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK);
```
nesbox, line 1620: We have to initialize SDL first here to detect the keyboard layout for the gamepad mapping we need in `studioInit()`. Pls revert or find a better way to init the config, because it breaks keyboard input for the gamepad.

src/studio/screens/main_menu.c
```diff
@@ -0,0 +1,381 @@
...
+// SOFTWARE.
+
+#include "../studio.h"
+#include "../studio_impl.h"
+#include "menu.h"
+#include "main_menu.h"
+
+// TODO: we should be able to shrink this access 
+// currently this is only used for impl.gamepad
+extern StudioImplementation impl;
```
nesbox, line 30: maybe pass `impl.gamepad` pointer directly to `initMainMenu()`

nesbox (changes requested):

src/studio/screens/main_menu.c
joshgoebel: Or, so me the bigger problem was that we're using `impl`, not how we get ahold of it. :-)  But yes, we can def pass it instead and remove the extern - that's a hair nicer.

src/system/sdl/main.c
nesbox: > but why do we need SDL to just load the config

Because `config` calls `tic_sys_default_mapping()` when initializing.

joshgoebel: Right, that just copies static default enumerable values into `&config->data.options.mapping`... I'm still not seeing why we need SDL init *at that point*...?  Config doesn't even know about SDL (one of many platforms) so I'm not sure I"m understanding the dependency you think is here?

joshgoebel: _Oh wait, looking at the wrong source file._ *looking closer*

CMakeLists.txt
```diff
@@ -1098,6 +1099,7 @@ if(BUILD_STUB)
             src/system/sdl/main.c
             src/studio/screens/run.c
             src/studio/screens/menu.c
+            src/studio/screens/main_menu.c
```
joshgoebel, line 1102: Not sure here, because I'm not sure what `BUILD_STUB` does?

nesbox: It's ok, I use the `BUILD_STUB` option to build the export binaries without editors.

joshgoebel: Ah so that's part of the regular build process but to test it I'd want to do a Mac export say and run that individually?

nesbox: Unfortunately, it's not set up for local testing yet, you can only test it when I upload the stubs to the server.
I'll have to think about how to test stubs locally.

joshgoebel: Wait, so aren't the export binary part of the compiled project - aren't they needed if I do an export?

nesbox: TIC-80 downloads export binaries from the server, we have about 48 binaries for every supported OS and script and can't embed all of them to the main binary :)

nesbox:
Can we merge your PR or are you going to make some other changes?

joshgoebel:
> Can we merge your PR or are you going to make some other changes?

It's good to go I think.  I can make more PRs.  Small PRs are good I think.

nesbox (approved):

PR #1823, opened 2022-01-24
mruby: Make Ruby Build Optional

RobLoach:
When building TIC-80 on architectures that don't have Ruby installed, the build fails. It would be great if we were still able to build TIC-80, just without the mruby support. While I understand it's possible to cross-compile it, it can be difficult to handle the target across all the platforms we're hitting.

## Implementation

- Adds an option named `BUILD_WITH_MRUBY`, which defaults to `TRUE` when `rake` is found
- Uses `BUILD_WITH_MRUBY` to determine whether or not to enable mruby

### DIFF
--- a/src/api/mruby.c
+++ b/src/api/mruby.c
@@ -22,6 +22,7 @@
 // SOFTWARE.
 
 #include "core/core.h"
+
 #if defined(TIC_BUILD_WITH_MRUBY)
 
 #include <stdlib.h>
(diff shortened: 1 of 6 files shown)


src/api/mruby.c
```diff
@@ -22,7 +22,6 @@
 // SOFTWARE.
 
 #include "core/core.h"
-
```
RobLoach, line 25:
``` suggestion

```

PR #1850, opened 2022-02-24
fix(zig) fix template compiling and add TextureSource

joshgoebel:


joshgoebel:
@nesbox Spot on you had it, but we really should have an enum for texture source, so this adds that as well as fixes the template example.

demos/wasm/src/tic80.zig
```diff
@@ -29,6 +29,13 @@ const RGBColor = extern struct {
     b: u8,
 };
 
+const TextureSource = enum(i32) {
+    TILES = 0,
+    MAP,
+    VBANK1
```
nesbox, line 35: It is not necessary to use `VBANK1`, it can also be `VBANK0`.
So `textri()` is using an inactive vbank for rendering, maybe just call it `VBANK`?

joshgoebel: vbank(0) would use bank 1 and vbank(1) would use bank 0?

I'll reconsider the name.  Technically it's SCREEN_INACTIVE_VBANK or some such. :-)

nesbox: ok :)

nesbox (approved):

Issue #1773, opened 2022-01-01
Request: Add `start` callback for compiled/WASM runtimes


joshgoebel:
Many compile languages can't have a concept "top-level" code (all the code must be in functions)... so there is no opportunity to perform start-ups tasks... having to do a manual check in `TIC` isn't terrible, but is kind of an ugly pattern IMHO.  If we supported a `init` or `start` or `boot` hook or something like PICO-8 we would eliminate this problem.

For example from my Zig cartridge:

```zig
export fn TIC() void {
    if (!started) { start(); }

  // actual TIC code here
}
```

It could be nice if this wasn't necessary and we could instead do:

```zig
export fn BOOT() void {
  // startup code
}
export fn TIC() void {
  // per tick code
}

```

nesbox:
Ok, let's add `BOOT()` function.

paul59:
Can we get an INIT() for regular cartridges using Lua etc?

nesbox:
> Can we get an INIT() for regular cartridges using Lua etc?

Yes, the callback will work everywhere, if we add it, `INIT()` or `BOOT()` doesn't matter.

paul59:
> > Can we get an INIT() for regular cartridges using Lua etc?
> 
> Yes, the callback will work everywhere, if we add it, `INIT()` or `BOOT()` doesn't matter.

INIT() would be my preference - that's the function name used by many other 'game engines' (eg Defold, Godot, Pico, Love2d, Unreal. 'Boot' implies loading the system itself (and out of the user's control).

joshgoebel:
>  'Boot' implies loading the system itself 

You boot a cartridge in my thinking... I think many early systems didn't have much of an OS, so I actually think it's a nice throwback name... but also I think if many other engines use 'INIT' that would be enough of an argument for me.

I see this as useful for, but entirely separate from the WASM work... so if someone else wanted to pick this up and run with it, that'd be fine.  It's also not *required* for WASM, but just a "very nice to have"... so may not be something I get to myself.

nesbox:
Added `BOOT()` callback to all the supported scripts and WASM aeffb6d4ca72db22540bcc3c56ab81230f5583bd

PR #1863, opened 2022-03-21
(chore) update Zig source to use new BOOT()

joshgoebel:
Update Zig examples to use the new `BOOT` API.

Follow-up on #1773


nesbox (approved):
Thank you.

PR #1890, opened 2022-04-23
Fix crash when internet is disabled

OldiLo:
When I run TIC-80 without the Internet, I get a segmentation fault


nesbox (approved):
Thank you

PR #1891, opened 2022-04-24
Improved D template

### LINKED ISSUE #1888: Wasm: Type mismatch between function signatures and implementations
In ```src/api/wasm.c```, in ```linkTicAPI```, all function parameters are declared to be either ```i``` or ```*```. But some functions, e.g., ```wasmtic_line```, are actually reading floats off the stack, not integers.  For ```wasmtic_line```, its correct signature is ```v(ffffi)```.

I've verified that ```line()``` doesn't work with the current incorrect signature, and works with the fixed signature.

Will send PR once I verify the other handful of functions.

PierceNg:
- Add Makefile and dub.json for D.
- Improved README.
- Fix function signature type mismatches.

Fixes #1888.

### DIFF
--- a/templates/d/buildcart.sh
+++ b/templates/d/buildcart.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+rm -f game.tic
+make clean
+make
+tic80 --fs . --cmd 'load wasmdemo.wasmp & import binary cart.wasm & save game.tic & exit'
+tic80 --fs . --cmd 'load game.tic & run & exit'
--- a/templates/d/buildwasm.sh
+++ b/templates/d/buildwasm.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+make clean
+make
+tic80 --fs . --cmd 'load wasmdemo.wasmp & import binary cart.wasm & run & exit'
--- a/templates/d/dub.json
+++ b/templates/d/dub.json
@@ -1,0 +1,38 @@
+{
+  "name": "cart",
+  "targetType": "executable",
+  "configurations": [
+    {
+      "name": "raw",
+      "dflags": [
+        "-betterC"
+      ],
+      "lflags": [
+        "--strip-all",
+        "--allow-undefined",
+        "--stack-first",
+        "--no-entry"
+      ]
+    },
+    {
+      "name": "wasi",
+      "dflags": [
+        "-betterC",
+        "-Xcc",
+        "$WASI_SDK_PATH/share/wasi-sysroot"
+      ],
+      "lflags": [
+        "--strip-all",
+        "--allow-undefined",
+        "--stack-first",
+        "--no-entry",
+        "$WASI_SDK_PATH/share/wasi-sysroot/lib/wasm32-wasi/libc.a"
+      ]
+    }
+  ],
+  "toolchainRequirements": {
+      "dmd": "no",
+      "gdc": "no",
+      "ldc": ">=1.11.0"
+  }
+}
--- a/templates/d/src/tic80.d
+++ b/templates/d/src/tic80.d
@@ -4,12 +4,54 @@ extern(C):
 
 struct MouseData {
     short x; short y; 
-    short scrollx; short scrolly;
+    byte scrollx; byte scrolly;
     bool left; bool middle; bool right;
 }
 
 const int WIDTH = 240;
 const int HEIGHT = 136;
+
+// These are pointers.
+const FRAMEBUFFER_PTR = cast(ubyte*)0;
+const TILES_PTR = cast(ubyte*)0x4000;
+const SPRITES_PTR = cast(ubyte*)0x6000;
+const MAP_PTR = cast(ubyte*)0x8000;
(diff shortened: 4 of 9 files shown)


templates/d/dub.json
```diff
@@ -0,0 +1,38 @@
...
+    {
+      "name": "raw",
+      "dflags": [
+        "-betterC"
+      ],
+      "lflags": [
+        "--strip-all",
+        "--allow-undefined",
+        "--stack-first",
+        "--no-entry"
```
joshgoebel, line 14: We also need to pull over:

 ```
        "--initial-memory=65536",
        "--max-memory=65536",
        "-zstack-size=14752"
```

But with TIC-80 specific values, which I think would mean 256kb memory and 96kb "reserved"... which I THINK means stack-size should be `96*1024 + 8 * 1024` (assuming we want an 8kb actual stack)... does that sound right?

File: templates/d/buildwasm.sh [hunk 2]
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+make clean
+make
+tic80 --fs . --cmd 'load wasmdemo.wasmp & import binary cart.wasm & run & exit'
```
joshgoebel, line 4: Just me, but nothing about `build` to me implies "run" also...  I think perhaps we need `build_and_run` or something...? Thoughts?

templates/d/src/tic80.d
```diff
@@ -4,13 +4,55 @@ extern(C):
...
+const WAVEFORMS_PTR = cast(ubyte*)0xFFE4;
+const SFX_PTR = cast(ubyte*)0x100E4;
+const MUSIC_PATTERNS_PTR = cast(ubyte*)0x11164;
+const MUSIC_TRACKS_PTR = cast(ubyte*)0x13E64;
+const SOUND_STATE_PTR = cast(ubyte*)0x13FFC;
+const STEREO_VOLUME_PTR = cast(ubyte*)0x14000;
+const PERSISTENT_MEMORY_PTR = cast(ubyte*)0x14004;
+const SPRITE_FLAGS_PTR = cast(ubyte*)0x14404;
+const SYSTEM_FONT_PTR = cast(ubyte*)0x14604;
+const WASM_FREE_RAM_PTR = cast(ubyte*)0x18000;
```
joshgoebel, line 33: Do you really think we should provide this out of the box?  Does D not allocate globals and such things starting here automatically?

templates/d/buildcart.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+rm -f game.tic
+make clean
+make
+tic80 --fs . --cmd 'load wasmdemo.wasmp & import binary cart.wasm & save game.tic & exit'
+tic80 --fs . --cmd 'load game.tic & run & exit'
```
joshgoebel, line 6: Why do we need to invoke tic80 twice vs doing it all in a single session?

templates/d/dub.json
PierceNg: Let me run some tests with the numbers.

templates/d/buildwasm.sh
PierceNg: I'm fine calling it ```build_and_run```. As is typical, this started out with the make commands, and then I added the tic80 command line too to run. As the README states, these are just convenience scripts.

templates/d/src/tic80.d
PierceNg: I suppose you are referring to WASM_FREE_RAM_PTR? Since TIC-80 allows peeking and poking anywhere, I put this in as well. But yes, generally the game programmer will leave to D to manage memory.

templates/d/buildcart.sh
PierceNg: My intention is to demonstrate the two separate actions -- creating a .tic file from necessary artifacts, and running the .tic directly without reference to said artifacts -- for the programmer who is new to this console.

templates/d/src/tic80.d
joshgoebel: I know. I'm just not sure we want to _encourage_ this.  Someone who knows what they are doing can add that line to their own source easily enough without us including it in the core examples.   And if they do not know what they are doing they probably shouldn't be using `WASM_FREE_RAM_PTR` in the first place... 

Thoughts?

nesbox (approved):

nesbox:
I OK with the PR, @joshgoebel please let me know when we are ready to merge.
Thank you.

templates/d/dub.json
PierceNg: So the numbers for TIC-80 would be, for 8k actual stack:
```
       "--initial-memory=262144",
       "--max-memory=262144",
       "-zstack-size=106496"
```

templates/d/src/tic80.d
PierceNg: Let's leave it out then.

templates/d/dub.json
joshgoebel: Did we ever get this changed?

PR #1912, opened 2022-05-04
Refactor and improve browser versions (tic-html, export stubs, webapp)

tranxuanthang:
This PR refactor & fix various problems in browser/HTML versions:

- Build emscripten script as ES6 module, so that emscripten doesn't pollute the global namespace and can be easier to import to SPA projects (vue, react,...)
- Change `SDL_EMSCRIPTEN_KEYBOARD_ELEMENT` to the canvas element so that emscripten won't block all keyboard input in the rest of the webpage (fixing the problem that make typing or backspace-ing comment content on tic80.com not possible after starting the game)
- Add tabindex to canvas element so that canvas can be focused (and unfocused) and capture the keyboard input properly
- Refactor html template to be more minimal, and show a simple "play" icon (instead of "click to play" text)
- Small changes to html build in github action's script

This is my first PR so forgive me if I missed anything.


joshgoebel:
What is the goal of:

```
-s EXTRA_EXPORTED_RUNTIME_METHODS=['ENV'] 
```

Also, we should state clearly this is going to break some older browsers that don't support ES6 modules... how far bad do we want our browser support to extend?  Do we have any stats from the website showing us the browser distribution (to perhaps use as a rough metric instead of just spitballing).  I'm not *super* worried about this, but it's worth mentioning since if it was more than a few users I'm personally not sure the benefits listed here are worth breaking compatibility.

joshgoebel:
Also if we're adding sub-directories now I'm not sure `html` and `html-export` fully clarify things (at all)... could we use more descriptive names that make it clear what we are building in the two different folders?  (good naming is my pet peeve) :-)

tranxuanthang:
@joshgoebel I don't think there is any problems in compatibility, ES6 modules feature (import/export) is supported by most modern browsers, and has roughly same browser support to webassembly, so if browser does not support ES modules then it probably does not support webassembly anyway.

These are browser compatibility information of ES6 modules amd webassembly:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#browser_support
https://developer.mozilla.org/en-US/docs/WebAssembly

And also it is possible to support even older browsers with javascript build tools and bundlers (webpack, esbuild) when necessary.

tranxuanthang:
The argument `-s EXTRA_EXPORTED_RUNTIME_METHODS=['ENV'] ` is used so that we can dynamically define custom environment variable in preRun:
https://github.com/nesbox/TIC-80/pull/1912/files#diff-f1f3a943063f166d85dab81f0e9a1a1054fbb930f44c996018497c39534ed583R79-R83

It is not needed without `-s MODULARIZE=1`, because this way ENV is defined as a global variable (https://emscripten.org/docs/porting/connecting_cpp_and_javascript/Interacting-with-code.html#environment-variables). But with `-s MODULARIZE=1` I have to add `-s EXTRA_EXPORTED_RUNTIME_METHODS=['ENV']` so that ENV is created as a property inside Module object and editable in preRun (https://github.com/emscripten-core/emscripten/issues/9827).

tranxuanthang:
> Also if we're adding sub-directories now I'm not sure html and html-export fully clarify things (at all)... could we use more descriptive names that make it clear what we are building in the two different folders? (good naming is my pet peeve) :-)

@joshgoebel I'm super bad at naming things, do you have any suggestions? ;-)

nesbox (approved):
I'm not a web developer and I don't understand some things, but anyway let's merge it.
Thank you for the help. 👍

PR #1926, opened 2022-05-18
Enable scrolling Code Editor by key-combinations like Ctrl-P and Ctrl-N.

sakisakira:
In Code Editor, we can scroll the the code with cursor keys.
For example, if the cursor is at the end line of the screen and if we press down arrow key,
the code will scroll up.
We also can move the cursory by pressing several key-combinations like Ctrl-P, Ctrl-N, Ctrl-E, Alt-Left, Alt-Right.
On the other hand, the key-combinations does not scroll up/down.
This PR will enable the scrolling with the key-combinations.

### DIFF
--- a/src/studio/editors/code.c
+++ b/src/studio/editors/code.c
@@ -1626,11 +1626,9 @@ static void processKeyboard(Code* code)
                 goNextBookmark(code, code->src);
         }
     }
-    else usedKeybinding = false;
-
-
-    if(ctrl || alt)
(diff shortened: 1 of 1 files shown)


nesbox (approved):
👍

BuoYancYdabl:
Wow, I was wondering where my cursor goes everytime I accidentally press ctrl+p instead of ctrl+o :D

PR #1936, opened 2022-05-27
Update fedora build instructions

rktjmp:
Update fedora build instructions for latest release.

I did this in a docker container, so it may over-specify vs what comes with a base install but the base should just not install any already existing packages.


nesbox (approved):

nesbox:
Thank you.

PR #1943, opened 2022-05-30
Support OSX PPC, GameCube, Wii and WiiU as libretro core

phcoder:
Unlike my previous attempt with big-endian this one minimizes code changes and instead fixes structs. Only chunk size and sound register needed special handling beyond structure change


include/tic80.h
```diff
@@ -76,14 +77,25 @@ typedef union
...
     {
-        bool up:1;
-        bool down:1;
-        bool left:1;
-        bool right:1;
-        bool a:1;
-        bool b:1;
-        bool x:1;
-        bool y:1;
+#if RETRO_IS_BIG_ENDIAN
```
joshgoebel, line 80: Why is there so much of this, as long as all the code is on the same endianness, can't the compiler just sort this out itself?  Is the problem use code inside the VMs or?

src/core/draw.c
```diff
@@ -207,8 +207,8 @@ static void drawSprite(tic_core* core, s32 index, s32 x, s32 y, s32 w, s32 h, u8
     if (index < 0)
         return;
 
-    rotate &= 0b11;
-    flip &= 0b11;
+    rotate &= 3;
+    flip &= 3;
```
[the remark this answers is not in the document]
phcoder: OSX PPC default compiler is gcc 4.2 which doesn't support binary notation. Other than normal int a solution could be to have defines for BINARY_11 and similar

include/tic80.h
phcoder: Problems arise from 2 sources:
* Loading little-endian cartridge (few places)
* Union between bit-field and plain types like u*. E.g. you can write through gamepads.first.x and read through gamepads.data. Then the result will depend on endianness unless we do those changes

phcoder:
Added ngc/wii/wiiu as they are only a small patch on top of osx-ppc

joshgoebel:
I'm also afraid of such huge patches for endianess. Are they no better solutions?

phcoder:
> To be honest, I'm a little afraid of such changes, a lot of overhead for defining structures, isn't it? Maybe we should add some unit testing for big endian support, to not break it accidentally in the future.

Where are the existing unit tests? Unit tests usually would catch big-endian issues as long as they are executed in big-endian environment.

> Also, Is it still in progress? if yes, pls convert it to draft.

Not in progress. Pretty much finished.

nesbox (approved):

nesbox:
THan you for your help, pls add yourself to the [credits](https://github.com/nesbox/TIC-80#credits) list.

joshgoebel:
@phcoder Several bugs have been reported that seem likely to have originated here... would you be willing to triane and work on fixes for them perhaps?  If so I can do a quick sweep and drop a short list here.

PR #1946, opened 2022-06-02
Feature: Optional auto-closing of delimiters

benthor:
This PR adds optional auto-closing of opening brackets, braces and parentheses. It's off by default but can be enabled by setting `AUTO_DELIMITERS` to `true` in `config.lua`. If enabled, this feature will simply insert the matching closing symbol in the direction of the (advancing) cursor whenever an opening bracket, brace or parenthesis is entered.

(Background: This feature was primarily motivated to ease development of Fennel code, where it is most useful. Lua users might be less interested but could still find it helpful)


nesbox (approved):
Thank you.

PR #1957, opened 2022-06-14
Add metainfo and fix case typo in desktop category.

kevinsmia1939:
Hi,
This PR adds metainfo file which contain information that might be useful to end user such as description, links, screenshots, etc. It is useful when TIC-80 is put is an app store such as Flathub for Linux which display these infos.
I name it as `com.tic80.TIC_80.metainfo.xml` so it would conform to reverse DNS scheme and dash is replace with underscore as recommended here:
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#file-naming

I change the `Categories=education` in desktop file to` Categories=Education` because it is case-sensitive as mention here: 
https://specifications.freedesktop.org/menu-spec/menu-spec-1.0.html#category-registry

I would also like to submit TIC-80 to Flathub, which allow linux user to install it very easily, let me know what you think.
https://flathub.org/home
Thanks.

### DIFF
--- a/build/linux/com.tic80.TIC_80.metainfo.xml
+++ b/build/linux/com.tic80.TIC_80.metainfo.xml
@@ -10,15 +10,18 @@
         <p>
            TIC-80 is a free and open source fantasy computer for making, playing
            and sharing tiny games.
-
+        </p>
+        <p>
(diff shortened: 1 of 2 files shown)


nesbox (approved):
Thank you

PR #1967, opened 2022-06-25
Add smooth line drawing in sprite editor

solarized-fox:
currently when drawing in tic80's sprite editor, only the point at the current mouse position is "stamped", which leads to undesirable gaps in the user's stroke:
before

this pull request changes the behavior to draw a line between the previous mouse position (now cached as a member of the Sprite struct) and the current one, resulting in an unbroken stroke:
after

hopefully this change is useful, if there is anything that I should change please feel free to let me know, too.

### DIFF
--- a/src/studio/editors/sprite.c
+++ b/src/studio/editors/sprite.c
@@ -175,7 +175,7 @@ static void paintPoint(Sprite* sprite, u8 color, s32 x, s32 y)
             tic_tilesheet_setpix(&sprite->sheet, x+i, y+j, color);
 }
 
-static void paintLine(Sprite* sprite, s32 Size, u8 color, s32 x1, s32 y1, s32 x2, s32 y2)
+static void paintLine(Sprite* sprite, u8 color, s32 x1, s32 y1, s32 x2, s32 y2)
 {
     float dx = x2 - x1;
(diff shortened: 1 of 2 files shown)


src/studio/editors/sprite.c
```diff
@@ -165,6 +166,61 @@ static void processPickerCanvasMouse(Sprite* sprite, s32 x, s32 y, s32 sx, s32 s
...
+static void paintPoint(Sprite* sprite, u8 color, s32 x, s32 y)
+{
+    s32 pixels = sprite->brushSize;
+
+    for(s32 j = 0; j < pixels; j++)
+        for(s32 i = 0; i < pixels; i++)
+            tic_tilesheet_setpix(&sprite->sheet, x+i, y+j, color);
+}
+
+static void paintLine(Sprite* sprite, u8 color, s32 x1, s32 y1, s32 x2, s32 y2)
```
nesbox, line 178: You can call `tic_api_line()` instead the own line dwawing implementation.

nesbox: Oh, sorry, I see, you draw line depending on brush size.

nesbox (approved):

Issue #1784, opened 2022-01-05
Compiled Languages support / WASM language bindings


joshgoebel:
Creating a master topic to link to from our wiki/documentation for those wishing to contribute language templates/libraries.

We should be able to support all the same languages as a [WASM-4](https://wasm4.org) project.  The list:

- [ ] C
- [ ] C++
- [ ] D
- [ ] Go
- [ ] Nelua
- [ ] Nim
- [ ] Odin
- [ ] Rust
- [ ] Wat
- [x] Zig

If you'd like to help speak up on this thread.

PierceNg:
I've written the low-level D bindings. The standard hello demo works on Linux and macOS. PR incoming soon.

joshgoebel:
Awesome.  Looking forward to it!

joshgoebel:
@keithohara If you find the time to work on your C/C++ libs again too in the future that'd be awesome.

PierceNg:
A quick experience report, based on translating Lua examples on the wiki.

The tri() example uses sin and cos functions. 

- With D, Wasm3 errs at runtime that "WASM must export a TIC function". Generating a .wat file from the .wasm shows BOOT() and TIC() are exported, and also that sinf() and cosf() are expected to be provided by the Wasm runtime. The .wasm file is 1773 bytes long. I suspect the error is because sinf and cosf aren't available.

- In Zig, the trig functions are part of Zig's runtime, and the compiled Wasm works. 

One of the mouse examples constructs a simple string indicating current mouse coordinates.

- With D, using the built-in library function format() pulls in unresolved dependencies.

- I know even less Zig than I do D, but based on checking the interwebs, it seems Zig doesn't come with string handling, so this needs to be implemented.

Possibility: Modify TIC-80 so that Wasm3 exports more libc functions like trig and string handling functions. 

I'm also experimenting with the Free Pascal Wasm compiler. Currently it generates .wasm files that are too big for TIC-80.

As an aside, Free Pascal also has a Javascript transpiler named pas2js. The examples implemented in Pascal and transpiled into Javascript work fine.

joshgoebel:
>  Modify TIC-80 so that Wasm3 exports more libc functions like trig and string handling functions.

[...]

PR #1971, opened 2022-07-01
Fix WASM font signature

soxfox42:
I found another instance where the WASM function signature was incorrect. The line in `linkTicAPI` had only 8 arguments for `font`, the Zig and D bindings had 9, and the `wasmtic_font` implementation itself had 10. I've updated everything so that all 10 arguments are exposed everywhere. This shouldn't break existing binary compatibility, as the templates would already have been producing incompatible binaries when the `font` function was used.

I haven't been able to test either version fully, because I can't work out how to get the Zig template to compile at all, and the D template seems to still have a slightly incorrect function signature for `font`. Because of that, I'll leave this as a draft for now while I look into the issues further.


soxfox42:
The issue with D seems to have just been an issue with function signatures, specifically a missing `const` qualifier, and an incorrect type for three of the four `transcolors` arguments. I was able to build a simple cart using `font` after fixing those minor issues.

Zig, on the other hand, has a lot of weirdness going on. I worked out that it definitely doesn't build with Zig 0.9.1, only the master branch for now. Even after managing to get a build of that working on my system, the demo seems to be written for the raw API, and the user-friendly API seems to have some syntax errors. For now, I think the updated `font` signature should be fine, so I'll mark this pull request as ready, and move discussion of the Zig template back over to #1784.

templates/d/src/tic80.d
```diff
@@ -63,12 +63,12 @@ void exit();
 void elli(int x, int y, int a, int b, int color);
 void ellib(int x, int y, int a, int b, int color);
 bool fget(int id, ubyte flag);
-int font(char* text, int x, int y, ubyte transcolors, int colorcount, int width, int height, bool fixed, int scale);
+int font(const char* text, int x, int y, uint* transcolors, int colorcount, int width, int height, bool fixed, int scale, bool alt);
```
joshgoebel, line 66: Are you SURE this is cont char*?

src/api/wasm.c
```diff
@@ -972,12 +972,12 @@ M3Result linkTicAPI(IM3Module module)
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "exit",    "v()",           &wasmtic_exit)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "fget",    "i(ii)",         &wasmtic_fget)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "fset",    "v(iii)",        &wasmtic_fset)));
-    _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "font",    "i(*iiiiiii)",   &wasmtic_font)));
+    _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "font",    "i(*iiiiiiiii)", &wasmtic_font)));
```
joshgoebel, line 975: Wow can't believe you keep finding these, but looks correct (your change).

templates/d/src/tic80.d
soxfox42: Not really to be honest, I've never really done much with D. I added `const` because 1. that's what `print` and `trace` use and 2. it made my test compile.

I *think* `const` here just means the function won't be able to change the value, not that a constant value has to be passed.

src/api/wasm.c
soxfox42: Haha, we *should* be good now. I checked each line against the corresponding `wasmtic_*` functions, and I couldn't find any other mismatched bindings.

nesbox (approved):

PR #1974, opened 2022-07-02
Improve the Zig template

soxfox42:
As mentioned in https://github.com/nesbox/TIC-80/issues/1784#issuecomment-1171940804, I found a few problems with the current Zig template.

With this PR, there is no longer an example of using the raw API, as I couldn't come up with a suitable structure that supported multiple examples without duplicating tic80.zig. I don't think this is a major issue though, as I imagine most users of the WASM templates will use the wrappers anyway.

### DIFF
--- a/templates/zig/src/tic80.zig
+++ b/templates/zig/src/tic80.zig
@@ -210,7 +210,7 @@ const MapArgs = struct {
     h: i32 = 17,
     sx: i32 = 0,
     sy: i32 = 0,
-    transparent: []const u8 = .{},
+    transparent: []const u8 = &.{},
     scale: u8 = 1,
     remap: i32 = -1, // TODO
 };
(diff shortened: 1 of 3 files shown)


nesbox (approved):
Thank you

templates/zig/src/tic80.zig
```diff
@@ -269,7 +269,7 @@ pub const trib = raw.trib;
 
 const TextriArgs = struct {
     texture_source : TextureSource = TextureSource.TILES,
-    transparent: []const u8 = .{},
+    transparent: []const u8 = &.{},
```
joshgoebel, line 272: This is a bit strange, what version of Zig were you using again?

soxfox42: Just the latest master, since 0.9.1 is missing the `export_table` field on `LibExeObjStep`. 

Without the `&`, it seems to be an array literal (type omitted), which can't be coerced to a slice. The pointer to the array can be coerced though. ([Zig Documentation](https://ziglang.org/documentation/master/#Type-Coercion-Slices-Arrays-and-Pointers))

Issue #1956, opened 2022-06-14
WASM runtime memory methods cannot access all RAM


joshgoebel:
That does raise another question though, why do any of the templates use TIC-80's `memcpy` and `memset`? Implementing them natively in the source language should allow them to access the full 256 KB of memory used in WASM mode, but the TIC-80 versions appear to be limited to the 96 KB in `tic_ram`. The only reason I can see is possibly performance, since I would guess that the native C versions are faster than anything that runs in the WASM interpreter.

_Originally posted by @soxfox42 in https://github.com/nesbox/TIC-80/issues/1784#issuecomment-1151771461_

---

I've confirmed from reading source that this seems to be correct.  The memory size can't be assumed to be 96kb anymore.  It's 96kb on scripting runtimes and 256kb on the WASM runtime.

This is probably something that needs to be setup once (at runtime startup - or perhaps it's a static value inside each config) so that it's fast to access... anyways we need a way to access it when all we're given is a `*tic_mem` pointer, like in most API calls.

PR #1977, opened 2022-07-09
Add a Rust WASM template

soxfox42:
Based on the WASM-4 Rust template, this should allow WASM TIC-80 cartridges to be built using Rust.

It should be pretty much as functional as the Zig template, including safe wrappers for the TIC-80 API, as well as the raw bindings, and raw memory access.

One issue still remains, and while it doesn't prevent the use of this template, it does limit it. TIC-80's memory layout uses address 0, which is basically impossible to access in Rust. Currently, the template defaults to building without any optimisations, which does allow access to the `FRAMEBUFFER` array, but also significantly increases cart size and reduces performance.
One possible way I can see to fix this would be to implement an alternate memory layout. It would require an extra pointer or offset so that `memcpy`, `memset`, `peek`, and `poke` could access the correct region of memory even if the MMIO wasn't at the beginning of memory, but since extra data about the memory needs to be stored to solve #1956, I think this would be okay. I'm happy to implement this, probably along with a fix to #1956, would the extra logic to handle this be acceptable? 

Anyway, whatever happens with the memory situation, this template should be in a perfectly usable state, even if it doesn't have ideal performance.


nesbox (approved):
Thank you for the Rust template 👍

joshgoebel:
This is really a Rust problem IMHO... we're effectively "legacy" hardware, Rust has to deal with us as-is... it doesn't make sense to add complexity for a single compiled language in a single runtime.  Rust could add a `FRAMEBUFFER1` pointer and just avoid every writing to the first byte of the display... yeah it sucks, but...

> One possible way I can see to fix this would be to implement an alternate memory layout. 

So I think we've (@nesbox) been pretty against this in the past - I'd guess because of the complexity and edge cases this introduces.   If you're really chomping at the bit AND the change was small and non-invasive (a few lines somewhere) I'd go to bat for you, but I fear it is likely to be neither.  How were you planning to solve this?

> along with a fix to #1956

A fix for this would be a great PR though.

soxfox42:
Yeah, I do see that it's a specific language issue, and Rust is now usable here, so it's not too big an issue if it can't be fixed.

> How were you planning to solve this?

Each of the memory access APIs currently accesses ram by finding a pointer using something like: `u8* ram = (u8*)memory->ram;`. I was thinking I could change this to instead store the pointer inside `tic_mem` (and also the memory size for #1956), then just use that pointer directly. Then, `initWasm` could detect a specific metatag to decide the value of `core->memory.ram`. The changes would be spread across each of the functions that access RAM, but I think it's still a fairly minor change if implemented that way.

joshgoebel:
>  (and also the memory size for #1956),

It might be better to track this in the per-runtime configuration, though I'm forgetting how easy that is to access during API calls.

> was thinking I could change this to instead store a second pointer inside tic_mem 

A pointer to where?  We're not even in charge of that allocation... WASM allocates it... so it isn't possible to over-allocate... you'd really need to subtract from the pointer (point earlier) to move the MMIO higher, wouldn't you?  And if we're not in charge of the allocation - then we can't front pad it with the extra RAM we'd need.  Am I missing something?

soxfox42:
> A pointer to where?

To `wasm_ram` (from `m3_GetMemory`), exactly as it is now. It's actually the pointer to `tic_ram` that I would have to offset.

> pointers to RAM would also need to know about the shift

See above, `core->memory.ram` would be updated, which all other code is already using.

> It might be better to track this in the per-runtime configuration
Fair point, I'll look into that, thanks.

joshgoebel:
> It's actually the pointer to tic_ram that I would have to offset.

Yes, but then wouldn't that expose internal data structure and require extra bounds checks on the low side?  And how does that solve pointers inside the VM runtime, which isn't using the tic_ram pointer at all?

joshgoebel:
> Quite possible, but I'm happy to give it a shot anyway.

All right then... :-)  Good luck.

Issue #1540, opened 2021-08-09
File access android. Google pixel 5.


Gouitsu:
Follow up post from an email.

Basically. I can't access the TIC-80 folder to copy files from and to. I can access the the "files" folder, the directory above but im not sure if im be able load my .pngs and .tic files from that folder.

I'm using 0.90pro

Thanks.
![Screenshot_20210809-130858](https://user-images.githubusercontent.com/88678056/128745089-f215c27e-a8f6-4ccc-8124-b32544fc1f17.png)

nesbox:
What version of Android do you have?

Gouitsu:
I'm on Android 11. It's also just hand an update, first thing I checked was access to the folder. Unfortunately no change. I'm using EX File Explorer for browsing my folders, permissions has been setup for it also. I shall plug my phone into my PC tonight, see if I get any files access that way.

Gouitsu:
Is it possible for TIC-80 to have another accessible working directory on android, outside the android/data files?

Enemby:
Yeah, same here. I can actually access TIC-80's local folder with the system files app (not google files), on my Pixel 2, but it's impossible, at least on my phone, to copy files into anything under android/data, as it's intentionally hidden. This means, in all likelihood, that on android you can export files.. but you can't download anything and run it, so it's a pretty reduced experience.

ciseur68:
Hi guys, same problem for me on Xiaomi Mi 9 Lite, Android 11. This is because of Android 11 or above which has restrictions on android/data and obb.

Gavin-Paolucci-Kleinow:
Using the nightly build on my Pixel 5a with Android 13, I was able to mount the TIC-80 folder in [Material Files](https://github.com/zhanghai/MaterialFiles) and copy to and from the folder.

Here's the path I mounted:
content://com.nesbox.tic.documents/tree/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.nesbox.tic%2Ffiles%2FTIC-80

At first I mounted the wrong path and it seemed like it worked but I got security errors and it copied the files with empty contents:
content://com.nesbox.tic.documents/tree/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.nesbox.tic%2Ffiles

PR #1998, opened 2022-07-30
Add DocumentsProvider for Android.

sakisakira:
In Android 11 and later, file access from the other app got strict.  So we cannot edit files under the home directory of TIC-80.
This PR will enable to edit source code file under the home directory of TIC-80 by any Android text editor with Storage Access Framework.
https://developer.android.com/guide/topics/providers/document-provider?hl=en
Screenshot_20220730-224015
Screenshot_20220730-224028

These screenshots are of the file open dialog of a text editor app (Jota+).  I confirmed that I can edit hello.rb with the app and then the editing result is affected by pressing Ctrl-R on TIC-80.
I expect this PR might solve some part of #1540

### DIFF
--- a/build/android/app/src/main/AndroidManifest.xml
+++ b/build/android/app/src/main/AndroidManifest.xml
@@ -95,6 +95,17 @@
             </intent-filter>
             -->
         </activity>
+
+        <provider
+            android:name="TICDocumentsProvider"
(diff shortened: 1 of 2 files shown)


nesbox (approved):
Awesome, finally someone did it 😄 
Thank you.

Issue #1904, opened 2022-05-03
Time() bug?  1.0.2164 Pro 


Morokiane:
There seems to be an issue with time(). The functions I have that are calling time() are much slower on the desktop version of TIC-80, but still work correctly on the web version.

Further testing is showing this is affecting MacOS on both M1 and Intel...and is happening in Linux. Runs as expected in Windows

nesbox:
We have started to use `clock()` https://en.cppreference.com/w/c/chrono/clock std function here to measure time in 1.0, maybe you have a piece of code to profile it?

Morokiane:
This screenshot shows an FPS counter that utilizes time(). To the left is the desktop version FPS compared to the web version. The MacOS version stays around 22-26 fps while the web and Windows version will stay around 60. The cart is here [Platform Engine](https://tic80.com/play?cart=2367)

![Screen Shot 2022-05-04 at 5 16 22 PM](https://user-images.githubusercontent.com/4334870/166840371-f2af028f-9633-4b2d-8b1a-00379ab0505a.png)

nesbox:
Could you pls go to the game menu and check `VSYNC` option, pls try to enable/disable it and see what happens?

Morokiane:
Same result with vsync off.

nesbox:
Sorry to bother you, did you restart the app after disabling VSYNC?

Morokiane:
Yes.

massung:
Confirming the same issue on Linux and MacOS (VSYNC on/off doesn't matter). The example program at https://github.com/nesbox/TIC-80/wiki/time shows that time progresses much slower.

andreyorst:
This also happens in Android build. Can be seen with the following cart:

```lua
-- title:  game title
-- author: game developer
-- desc:   short description
-- script: lua

pt=0

function TIC()
	dt=(time()-pt)/1000
	pt=time()
	cls(13)
	print(dt,0,0)
end
```

On the correct build it shows the value of ~0.02, but on Linux and Android it shows the value of ~0.001

MineRobber9000:
Notably, this bug does not occur on Windows. If we ask MSDN what `clock` does, it will tell us:
[...]

PR #2007, opened 2022-08-13
override some scale settings; finer-grained SDL init

dhorlick:
# Problem

TIC-80's uiScale currently defaults to 4. This turns out to be a poor choice for everybody's favorite maybe-still-supported platform, the PocketC.H.I.P., which lacks a screen large enough to accommodate it.

For whatever reason (maybe [issue 1904](https://github.com/nesbox/TIC-80/issues/1904)?), recent TIC-80 builds also perform very badly at scale 4 on the PocketC.H.I.P., taking more than a minute to start up and then missing many keystrokes.

misframed

I recorded the image above after twice failing to "exit". Note that the cursor is not visible.

Additionally, sound is currently broken for TIC-80 on the PocketC.H.I.P. (but not for other SDL2 applications.) That is not a big problem by itself, but will complicate the implementation of a screen-framing fix slightly.

# Solution

A sophisticated user can circumvent the screen-framing problem on PocketC.H.I.P. by explicitly reconfiguring uiScale to a lower value. This pull request seeks to address the problem (and indirectly, also the worst of the performance issues) for all PocketC.H.I.P. users by conditionally overriding the uiScale setting, without prompting. Here's how:

   1. Attempt to check the screen bounds with SDL2 on startup. If they are reported, wait until defaults have been applied and file configuration has been loaded. Then, if screen bounds were reported, determine whether the prescribed uiScale would fit. If they would not, reduce uiScreen to the largest agreeable value. Note that the user will still be permitted to specify any scale they want (without a fail-safe) from the command line.
   1. Currently, TIC-80 attempts to initialize all SDL2 subsystems in tandem. Because the sound is broken, this initialization technically fails. That failure seems to compromise SDL2's ability (at least immediately) to report on screen bounds. The workaround is to initialize SDL2 subsystems separately. If the video subsystem fails to initialize, report the problem and quit immediately, because there's no chance TIC-80 will be able to perform productive work. If the sound (or joystick) subsystem fails to initialize, report the problem but keep going.

I have tested this successfully on Ubuntu 20.04.4 and of course on the PocketC.H.I.P., whose new effective default uiScale of 2 is still somewhat slow, but no longer accompanied by lots of missed keystrokes.

I expect most systems should see no change from this pull request, but I can imagine a few others (smartwatches, LED boards, etc.) for which it might be useful.

Feel free to repurpose code from this pull request if that would be more convenient than merging all of it.

### DIFF
--- a/src/studio/studio.c
+++ b/src/studio/studio.c
@@ -2367,7 +2367,7 @@ bool studio_alive(Studio* studio)
     return studio->alive;
 }
 
-Studio* studio_create(s32 argc, char **argv, s32 samplerate, tic80_pixel_color_format format, const char* folder)
+Studio* studio_create(s32 argc, char **argv, s32 samplerate, tic80_pixel_color_format format, const char* folder, s32 maxscale)
 {
     setbuf(stdout, NULL);
 
@@ -2483,6 +2483,13 @@ Studio* studio_create(s32 argc, char **argv, s32 samplerate, tic80_pixel_color_f
     tic_fs_makedir(studio->fs, TIC_LOCAL_VERSION);
     
     initConfig(studio->config, studio, studio->fs);
+
+    if (studio->config->data.uiScale > maxscale)
+    {
+        printf("Overriding specified uiScale of %i; the maximum your screen will accommodate is %i", studio->config->data.uiScale, maxscale);
+        studio->config->data.uiScale = maxscale;
+    }
+
     initStart(studio->start, studio, args.cart);
     initRunMode(studio);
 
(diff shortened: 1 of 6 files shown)


dhorlick:
Looks like there isn’t any interest in this. Will close next week if there hasn’t been any activity.

asiekierka:
This is a good PR! @nesbox The world of various ARM handhelds is one worth supporting.

nesbox:
I'm now distracted by another project, this weekend I'll try to review and merge.
Sorry for the inactivity, I hope to be back on TIC-80 soon.

dhorlick:
No worries, we all can relate!

Thanks, everybody.

nesbox (approved):
Let's merge it, sorry for so long code review.

PR #2010, opened 2022-08-21
Support fn+enter alias for insert on mac keyboards

fiskWasTaken:
The problem as it stands, there's no way to emulate an insert press on a Macbook keyboard, which makes composing music on the TIC-80 a little tricky. 

fn+backspace works fine as 'delete', but support for fn+enter is sporadic at best and for some reason it is equivalent to SDLK_KP_ENTER. So, this change just enables fn+enter as an insert keypress on Mac versions of the application.


nesbox (approved):
thanks

PR #2019, opened 2022-10-05
Use cross-platform timing method to fix time() bug

### LINKED ISSUE #1904: Time() bug?  1.0.2164 Pro

MineRobber9000:
Fixes #1904, using my original suggestion of moving sokol_time into the core.

Notable changes include:
 - `time()` should no longer be bugged
     - Sokol's time module deals with wall-clock time, unlike the `clock` function which is supposed to deal with CPU time
 - Sokol headers were updated
     - Because of how TIC-80 is architected, I have no way of ensuring `stm_setup()` only gets run once during the program's lifetime, short of making sure every single system calls `stm_setup()`. With the latest sokol headers, repeating `stm_setup()` is no longer a crash.

I've tested the build on my Windows machine, and soxfox has tested the build on Mac.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -420,6 +420,7 @@ macro(MACRO_CORE SCRIPT DEFINE BUILD_DEPRECATED)
         ${TIC80CORE_DIR}/core/draw.c
         ${TIC80CORE_DIR}/core/io.c
         ${TIC80CORE_DIR}/core/sound.c
+        ${TIC80CORE_DIR}/core/sokol_time_impl.c
         ${TIC80CORE_DIR}/api/js.c
         ${TIC80CORE_DIR}/api/lua.c 
         ${TIC80CORE_DIR}/api/moonscript.c 
@@ -451,6 +452,7 @@ macro(MACRO_CORE SCRIPT DEFINE BUILD_DEPRECATED)
             ${THIRDPARTY_DIR}/fennel
         PUBLIC
             ${CMAKE_SOURCE_DIR}/include
+            ${THIRDPARTY_DIR}/sokol
             ${CMAKE_SOURCE_DIR}/src)
 
     target_link_libraries(tic80core${SCRIPT}
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -23,6 +23,7 @@
 #include "api.h"
 #include "core.h"
 #include "tilesheet.h"
+#include "sokol_time_impl.c"
 
 #include <assert.h>
 #include <string.h>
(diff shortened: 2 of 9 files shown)


MineRobber9000:
Build failures are currently:

 - Sokol
     - Turns out, `sokol_gfx.c` did not take too kindly to me updating sokol from underneath it. I'm working on this one right now.
 - Baremetal
     -  The reported error is that CLOCK_MONOTONIC is not defined, which... is strange to say the least. soxfox seems to be working on figuring this one out.

soxfox42:
Yeah, looking at baremetal currently. Also, I've now tested it on Linux, all seems fine there.

soxfox42:
As for baremetal:

For some reason, the build system currently doesn't add the circle include directories to the include path. I got a little further by directly including circle's time.h when building for baremetalpi, but that fails when it expects to find circle headers in the include path. I haven't yet looked into tweaking the build to fix the include path, but I figured I'd put an update here before I finish up for the night, in case anyone has any input.

MineRobber9000:
I have hit a wall in debugging the failing Sokol build. There is a validation error in `sg_apply_pipeline` that causes a failure. Without access to the console and its debug messages, I cannot determine the cause of the failure.

CMakeLists.txt
```diff
@@ -420,6 +420,7 @@ macro(MACRO_CORE SCRIPT DEFINE BUILD_DEPRECATED)
         ${TIC80CORE_DIR}/core/draw.c
         ${TIC80CORE_DIR}/core/io.c
         ${TIC80CORE_DIR}/core/sound.c
+        ${TIC80CORE_DIR}/core/sokol_time_impl.c
```
nesbox, line 423: TIC-80 core should be independent from 3rd party libs, pls don't use Sokol here.
We have to find another way to measure time, for example we could add `time()` function to the `tic80` struct and implement it somewhere on the rendering side, where we use SDL and Sokol libs.
```c
    struct
    {
        u64 (*time)();

        void (*trace)(const char* text, u8 color);
        void (*error)(const char* info);
        void (*exit)();
    } callback;
```
https://github.com/nesbox/TIC-80/blob/d23ffe1daebf33780f6b8f7bc6d7c83c2716af0c/include/tic80.h#L60

nesbox (changes requested):

CMakeLists.txt
MineRobber9000: I feel like that's a waste of coding time, to make sure every independent system has its own way to get time. No matter what library you use, at the end of the day, you'll end up calling a function that basically does what Sokol's time library does, so why not just... use *that* one and avoid annoying platform-specific timing code?

If you're unwilling to yield on "don't use Sokol" then I'll just close this PR and start over. I personally don't see the point in writing our own function for something that's already a solved issue, but if "make a callback function in every single system TIC-80 supports for getting time" is the solution you genuinely want, I've done way too much in this PR to turn around and do that.

MineRobber9000:
Closing this PR. I will eventually get around to the add-a-callback-for-every-single-system-to-implement solution, but with finals fast approaching, I don't have time to deal with this right now.

Issue #2042, opened 2022-12-04
Multiple bugs with PNG cartridges


bztsrc:
Hi,

Very cool console you have! I like it very much! I only have issues with the PNG files. The way how TIC-80 currently handles them is unacceptable for several reasons:

- First, it's not supported everywhere, only in studio and in surf (but not from command line for example).
- Second, it does not work on systems with *case-sensitive* file systems (due to the LoadPngCart call depends on file extension!)
- Third, and this is the biggest problem, a real no-go for me, is the way how TIC-80 embeds data generates *CORRUPT* png files! Tools are refusing to resize the image because of that, or worse, they might corrupt the cartridge data silently.

Here's a proposed solution that fixes all of these problems at once and as a side effect improves performance a lot. This renders the use of `LoadPngCart` completely unnecessary, it Just Works(tm) and gets the cartridge from a png file *without* actually decoding the data:
```diff
diff --git a/src/cart.c b/src/cart.c
index 9be32c7..1c6079d 100644
--- a/src/cart.c
+++ b/src/cart.c
@@ -87,6 +87,28 @@ void tic_cart_load(tic_cartridge* cart, const u8* buffer, s32 size)
 
 #define LOAD_CHUNK(to) memcpy(&to, ptr, MIN(sizeof(to), chunk->size ? retro_le_to_cpu16(chunk->size) : TIC_BANK_SIZE))
 
+    // check if the cartridge buffer is a PNG file
+    if (!memcmp(buffer, "\x89PNG", 4))
+    {
+        s32 siz;
+        const u8* ptr = buffer + 8;
+        while (ptr < end)
+        {
+            siz = ((ptr[0] << 24) | (ptr[1] << 16) | (ptr[2] << 8) | ptr[3]) + 12;
+            if (!memcmp(ptr + 4, "caRt", 4))
+            {
+                buffer = ptr + 8;
+                size = siz;
+                end = buffer + size;
+                break;
+            }
+            ptr += siz;
+        }
+        // error, no TIC-80 cartridge chunk in PNG???
+        if (ptr >= end)
+            return;
+    }
+
     // load palette chunk first
     {
         const u8* ptr = buffer;
```
The advantage is, this patch is extremely simple, dependency-free and it does not require to decode the png at all! It just iterates through png chunks without interpreting them.

[...]

PR #2045, opened 2022-12-08
Adding png chunk cartridge support

bztsrc:
As explained in isse #2042 steganography has certain limitations. This patch add support storing the cartridge data in `caRt` PNG chunk as well. For backward compatibility, it still tries to use steganography if the cartridge fits into the pixels, and on load if no `caRt` chunk found, still tries steganography.

### DIFF
--- a/src/ext/png.c
+++ b/src/ext/png.c
@@ -1,6 +1,7 @@
 // MIT License
 
 // Copyright (c) 2021 Vadim Grigoruk @nesbox // grigoruk@gmail.com
+// Copyright (c) 2022 bzt png chunk stuff
 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -28,6 +29,7 @@
 #include <png.h>
 #include "tic_assert.h"
 
+#define EXTRA_CHUNK "caRt"
 #define RGBA_SIZE sizeof(u32)
 
 png_buffer png_create(s32 size)
@@ -48,7 +50,7 @@ static void pngReadCallback(png_structp png, png_bytep out, png_size_t size)
     stream->pos += size;
 }
 
-png_img png_read(png_buffer buf)
+png_img png_read(png_buffer buf, png_buffer *cart)
 {
     png_img res = { 0 };
 
@@ -102,6 +104,28 @@ png_img png_read(png_buffer buf)
 
         free(rows);
 
+        // Read in cartridge data from chunk if possible
+        if (cart)
+        {
+            png_unknown_chunkp unknowns = NULL;
+            int num_unknowns;
+
+            png_set_keep_unknown_chunks(png, PNG_HANDLE_CHUNK_ALWAYS, EXTRA_CHUNK, 1);
+            png_read_end(png, info);
+            num_unknowns = png_get_unknown_chunks(png, info, &unknowns);
+            for(s32 i = 0; i < num_unknowns; i++)
+                if (!memcmp(unknowns[i].name, EXTRA_CHUNK, 5))
+                {
+                    cart->size = unknowns[i].size;
+                    cart->data = (u8*)malloc(cart->size);
+                    if (cart->data)
+                        memcpy(cart->data, unknowns[i].data, cart->size);
+                    else
+                        cart->size = 0;
+                    break;
+                }
+        }
+
         png_destroy_read_struct(&png, &info, NULL);
     }
 
@@ -121,7 +145,7 @@ static void pngWriteCallback(png_structp png, png_bytep data, png_size_t size)
 
 static void pngFlushCallback(png_structp png) {}
 
-png_buffer png_write(png_img src)
+png_buffer png_write(png_img src, png_buffer cart)
 {
     png_structp png = png_create_write_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
     png_infop info = png_create_info_struct(png);
@@ -141,6 +165,17 @@ png_buffer png_write(png_img src)
         PNG_FILTER_TYPE_DEFAULT
     );
 
+    // Save cartridge data in a chunk too. This supports bigger cartridges than steganography
+    if (cart.data && cart.size > 0 && cart.size <= 0x7fffff){
+        png_unknown_chunk unknowns = { 0 };
+        memcpy(&unknowns.name, EXTRA_CHUNK, 5);
+        unknowns.data = cart.data;
+        unknowns.size = cart.size;
+        unknowns.location = PNG_AFTER_IDAT;
+        png_set_unknown_chunks(png, info, &unknowns, 1);
+        png_set_keep_unknown_chunks(png, PNG_HANDLE_CHUNK_ALWAYS, EXTRA_CHUNK, 1);
+    }
+
     png_write_info(png, info);
 
     png_bytep* rows = malloc(sizeof(png_bytep) * src.height);
@@ -148,7 +183,7 @@ png_buffer png_write(png_img src)
         rows[i] = src.data + src.width * i * RGBA_SIZE;
 
     png_write_image(png, rows);
-    png_write_end(png, NULL);
+    png_write_end(png, info);
 
     png_destroy_write_struct(&png, &info);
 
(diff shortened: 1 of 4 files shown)


src/ext/png.c
```diff
@@ -141,14 +165,25 @@ png_buffer png_write(png_img src)
         PNG_FILTER_TYPE_DEFAULT
     );
 
+    // Save cartridge data in a chunk too. This supports bigger cartridges than steganography
+    if (cart.data && cart.size > 0 && cart.size <= 0x7fffff){
```
nesbox, line 169: Here you store the cart data in the `caRt` chunk in addition to the steganographic data and it looks redundant, we are doubling the cart data in two places. I suggest letting the user decide how to save the `.png` by adding console command options like `save cart.png [way=(steno | caRt | both)]`.
What do you think?

dulsi: (bztsrc is having issues logging into github and asked someone to forward this response.)

    > Here you store the cart data in the caRt chunk in addition to the steganographic data and it looks redundant, we are doubling the cart data in two places.

    Only for now to provide backward compatibility.

    > I suggest letting the user decide

    I think it would be much wiser to abandon steganography entirely instead. My reasons:
    - steganography has very limited storage capacity (98k tops)
    - it silently fails if the cartridge data happens to be bigger than that
    - it is extremely vulnerable to data corruption
    - has no means at all to detect data corruption

    On the other hand, chunk based storage
    - has unlimited storage capacity (up to 2^31)
    - can't silently fail on save, it always works
    - has a CRC to detect data corruption (handled by libpng automatically)

joshgoebel: I agree.  I think we need to maintain backwards compatibility - to continue to READ old steganography but if you make changes and save that cartridge it should be encoded as PNG chunk metadata.  Giving the user 3 different choices is kind of the word of all approaches.

bztsrc:
Ah, that stupid verification code email finally arrived, I can answer again.

> to continue to READ old steganography but if you make changes and save that cartridge it should be encoded as PNG chunk metadata.

Agree. I haven't removed the stegano writer part yet, because we have to wait until everybody starts using the new version, but I agree after a transition period stegano-write should go, leaving only stegano-read for backward compatibility.

bztsrc:
@nesbox: Review needed please.

Plus here's a little addition to cart.c, which fixes issue #2048 It loads the cartridge data without using libpng (which is not linked with libtic80core.a), and this adds support for PNG cartridges in libretro (and probably on other systems as well). Needs this PNG chunk writer PR to be merged first of course, as it can only read the chunk.
```diff
diff --git a/src/cart.c b/src/cart.c
index 9be32c7..2d66fe1 100644
--- a/src/cart.c
+++ b/src/cart.c
@@ -85,6 +85,37 @@ void tic_cart_load(tic_cartridge* cart, const u8* buffer, s32 size)
     memset(cart, 0, sizeof(tic_cartridge));
     const u8* end = buffer + size;
 
+    // NOTE by bzt: we do not have libpng support here in libtic80core.a, so steganography is
+    // completely out of question, but we can find and uncompress a chunk ourself pretty easily
+    u8 *chunk_cart = NULL;
+
+    // check if this cartridge is in PNG format
+    if (!memcmp(buffer, "\x89PNG", 4))
+    {
+        s32 siz;
+        const u8* ptr = buffer + 8;
+        // iterate on chunks until we find a cartridge
+        while (ptr < end)
+        {
+            siz = ((ptr[0] << 24) | (ptr[1] << 16) | (ptr[2] << 8) | ptr[3]);
+            if (!memcmp(ptr + 4, "caRt", 4) && siz > 0)
+            {
+                chunk_cart = malloc(TIC_BINARY_SIZE);
+                if (chunk_cart)
+                {
+                    size = tic_tool_unzip(chunk_cart, TIC_BINARY_SIZE, ptr + 8, siz);
+                    buffer = chunk_cart;
+                    end = buffer + size;
+                }
+                break;
+            }
+            ptr += siz + 12;
+        }
+        // error, no TIC-80 cartridge chunk in PNG???
+        if (!chunk_cart)
+            return;
+    }
+
 #define LOAD_CHUNK(to) memcpy(&to, ptr, MIN(sizeof(to), chunk->size ? retro_le_to_cpu16(chunk->size) : TIC_BANK_SIZE))
 
     // load palette chunk first
@@ -219,6 +250,10 @@ void tic_cart_load(tic_cartridge* cart, const u8* buffer, s32 size)
                 }
         }
     }
+
+    // if we have allocated the buffer from a PNG chunk
+    if (chunk_cart)
+        free(chunk_cart);
 }
```
Cheers,
bzt

bztsrc:
Happy Xmas!

@nesbox: any chance of a review please?

Cheers,
bzt

nesbox:
@bztsrc hi, sorry for the delay with the review, too little time due to moving to another country.
In my opinion we should handle both scenarios here (caRt + shorthand) or nothing at all, ideally it could be somewhere in one place and I don't see any problem to link `libpng` with `tic80core`.

bztsrc:
@nesbox: no worries!

> In my opinion we should handle both scenarios here (caRt + shorthand) or nothing at all

Not sure what you mean. This patch handles both, and with the cart.c patch it even handles both tic and png cartridges in libretro (a feature that's not working ATM).

> ideally it could be somewhere in one place

Definitely, but that would require a refactor, out of scope of this PR which is about to introduce png chunk storage. Could be done a future PR, but first this functionality should be added.

> I don't see any problem to link libpng with tic80core.

But I do. Unnecessary bloat for nothing. Implementing steganography in tic80core requires a lot of work for a technology that's obsoleted by this PR.

This PR isn't about hacking stegano into libtic80core, this is about the replace stegano with a much better solution (that as a side effect, solves the libtic80core png loading issue too). But the emphasis is on replacing the unreliable, vulnerable and limited capacity stegano in the first place.

Cheers,
bzt

joshgoebel:
> This PR isn't about hacking stegano into libtic80core, this is about the replace stegano with a much better solution (that as a side effect, solves the libtic80core png loading issue too). 

Exactly, 100% agree. The new proposed solution paves the road for *easily* integrating PNG loading support in additional places, with almost (exactly?) zero new dependencies - that's a win win.  

If someone else LATER wanted to come along and add steganography support to libtic80core - that would be a sep PR.  But if this new method is better in all regards I also don't see the point in continuing to push steganography, esp. with the states known issues and incompatibilities.

---

Do you disagree with the new method is much better?

bztsrc:
Happy new year!

> with almost (exactly?) zero new dependencies

Exactly zero. The cart.c patch only needs `tic_tool_unzip`, which libtic80core.a already has, no additional libraries/functions required.

> Do you disagree with the new method is much better?

Yeah, I'd like to hear your opinion too,  if you don't want this PR, I wouldn't force it any more.

Cheers,
bzt

joshgoebel:
@bztsrc Bigger picture would you be interested in doing any of the work to deprecate/remove steganography if we decided to go this direction?  For example (one idea): removing the functionality completely - and replacing it with a tiny utility (or script) that converts a PNG from steganography to a DATA chunk.  A webpage could possibly even do this...

I wish we know how commonly cartridges are even passed around as PNGs out there in the wild...

bztsrc:
> Bigger picture would you be interested in doing any of the work to deprecate/remove steganography if we decided to go this direction?

Considering that I've already done the bigger part in this PR, yes, sure. If this PR gets merged, I'll add a new PR that refactors png loading, removing stegano-write (I'd keep stegano-read for compatibility), and using one common function to decode png.

> replacing it with a tiny utility (or script) that converts a PNG from steganography to a DATA chunk. A webpage could possibly even do this...

I've already created this tool, and it is also available as a web-service (via wasm) [here](https://bztsrc.gitlab.io/p8totic/), but you have my permission to add it to the TIC-80 webpage. Originally it was created to convert PICO-8 cartridges to TIC-80, but it can also convert .png files (either with stegano or chunk) to .tic and vice versa.

If the input is a .tic file, then the output is a .png which includes the data as stegano (if it fits, for backward compatibility) as well as in a chunk. This tool uses the default cartridge [image](https://github.com/nesbox/TIC-80/blob/main/build/cart.png) and the default [font](https://github.com/nesbox/TIC-80/blob/main/src/core/font.inl), grabs the title and the author from the script's comments, so the resulting png is 100% pixel correct with the png files saved using TIC-80.

Cheers,
bzt

nesbox (approved):
maybe the problem is me, because I implemented the stenography, I don't want to refuse it and because of this I don't see the advantages of the new way :)
ok, let's merge the PR  and see how it goes

joshgoebel:
> because of this I don't see the advantages of the new way

Isn't it pretty clear steno doesn't have enough bytes for large cartridges?

PR #2063, opened 2022-12-23
Fix rare segfault during shutdown

jezze:
Sometimes player-sdl segfaults during shutdown. By reordering the shutdown sequence the segfault seem to have disappeared.


jezze:
Compared to my previous pull request which is a bit dodgy in its implementation, this one is very straight-forward.

joshgoebel (commented):
Makes sense to me that we should first close things before tearing the world down.

nesbox (approved):

PR #2068, opened 2023-01-01
Don't show "unknown topic" when help message is displayed

locriacyber:
Remove the error "unknown topic" when typing `help api` and similar help commands.

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -3548,13 +3548,12 @@ static void onHelpCommand(Console* console)
         FOR(const struct Handler*, ptr, Handlers)
             if(strcmp(ptr->cmd, param) == 0)
             {
+                foundTopic = true;
                 ptr->handler(console);
                 break;
             }
 
-        if (foundTopic) {
-            commandDone(console);
-        } else {
+        if (!foundTopic) {
             printError(console, "\nunknown topic: ");
             printError(console, param);
         }


joshgoebel:
Please upload snap of the error case your solving here.

locriacyber:
Curiously, it's not broken at https://tic80.com/create

On branch 6ef9fda:

[Uploading a.webm…]()

src/studio/screens/console.c
```diff
@@ -3548,13 +3548,12 @@ static void onHelpCommand(Console* console)
         FOR(const struct Handler*, ptr, Handlers)
             if(strcmp(ptr->cmd, param) == 0)
             {
+                foundTopic = true;
                 ptr->handler(console);
                 break;
             }
 
-        if (foundTopic) {
-            commandDone(console);
```
joshgoebel, line 3556: The fix on 3651 looks good (setting the flag in a situation I missed in my patch), but what makes you think this needs to be removed?  I'm not sure this change is necessary.

locriacyber: `commandDone` is called twice (one more time below).

locriacyber:
Unrelated: The website is using a newer version (from the version number) of TIC80. Is that fix private?

src/studio/screens/console.c
joshgoebel: Ah yes, I see!

joshgoebel (approved):

PR #2069, opened 2023-01-02
Janet Support

locriacyber:


src/api/janet.c
```diff
@@ -0,0 +1,140 @@
...
+static const char* DYN_TIC_MEM = ":tic_mem";
+
+#define JANET_FN_S_tic(CNAME) \
+    static const int32_t _cfun_##CNAME##_sourceline_ = __LINE__; \
+    static Janet _cfun_##CNAME (int32_t argc, Janet *argv) { \
+        tic_mem* tic = janet_unwrap_pointer(janet_dyn(DYN_TIC_MEM)); \
+        assert(tic != NULL);
+
+JANET_FN_S_tic(btn)
+    janet_fixarity(argc, 1);
```
joshgoebel, line 32: Just to be clear this param is optional, not required.

src/api/janet.c
```diff
@@ -0,0 +1,140 @@
...
+    .blockStringStart = NULL,
+    .blockStringEnd = NULL,
+    .singleComment = "#",
+    .blockEnd           = NULL,
+
+    .keywords = JanetKeywords,
+    .keywordsCount = COUNT_OF(JanetKeywords),
+};
+
+#endif /* defined(TIC_BUILD_WITH_WREN) */
```
joshgoebel, line 140:
```suggestion
#endif /* defined(TIC_BUILD_WITH_JANET) */
```

src/api/janet.c
locriacyber: that's not in the doc

joshgoebel: I would agree it's unclear.

`btn([id]) -> pressed`

The `[]` means the argument is optional.  If no specific button is passed I believe the entire state of the button register (bitfield) is returned. (you can read the source to confirm)

I'm speaking largely of Lua, it might be useful to check and see if this has been handled correctly across all runtimes.

joshgoebel: Yes, `btn` and `btnp` have the same behavior for 0 arguments:

https://github.com/nesbox/TIC-80/blob/main/src/api/lua.c#L601

locriacyber: it seem to always return `true` for me in the Lua version.

AlecTroemel:
@locriacyber cool to see this MR, I actually spent some time on this same thing a while ago (branch here for comparison  https://github.com/AlecTroemel/TIC-80/tree/charlotte_janet_support). I'd love to help get this MR ready however possible! some questions

- Do syntax errors contain line numbers & other info like relevant function names? I could never get that sorted
- You have an opinion on the tic80 function names that collide with the janet standard library functions? Specifically `print` and `map`?

locriacyber:
Thanks for picking this up
- Janet prints error to stderr. It's hard coded in. Syntax error is considered compile-time error.
- I recommend making it a module, so the user can choose between `(import tic80)` or `(use tic80)`

Issue #2065, opened 2022-12-29
segfault on libretro core aarch64


dhwz:
The core crashes when a game is loaded. Happening with latest commit.

@jezze
Since this change the retroarch core is instantly crashing
https://github.com/nesbox/TIC-80/commit/2a700111a0dfaec8d92d3bf715bb1d1170367882#diff-2fb8122f88c730713f251fb6c985b2401f11b9f6053f9fc5f55329c98767a7cdR809

jezze:
I assume this is libretro? Yeah I wrote in the commit message that this probably was not a good solution because libretro needs to implement the two callback functions that is currently just set to NULL so I think it was a bit early to merge it but its not too late to fix! What is weird though is that libretro should not have worked before my change either because it also didnt set up those two callbacks and should have crasched just like the player. I can try to see if I can fix it possibly. How do I reproduce this crash? I know very little about retroarch.

dhwz:
I think it's already related to this change:
0cd38fd3044fb856be1c9a9db4fc2e1191c696e9

jezze:
Yeah if it is that change that introduced the problem then yes that makes total sense to me.

jezze:
Ok! Knowing what they did I think I can actually whip up a fix for this. Give me some minutes. Are you able to test it for me @dhwz later? Like I said I am not sure how to reproduce it.

jezze:
I created a fix here, I think this should work? https://github.com/nesbox/TIC-80/pull/2072

dhwz:
Thx, I'll try it and let you know if it's working.

dhwz:
@jezze yeah great that did it! I'll apply that as a patch till it's getting merged. Thank you very much! :)

nesbox:
fixed here #2072

PR #2072, opened 2023-01-06
Fix counter and freq for libretro

jezze:
This is the follow up to my previous fix to make sure libretro works as well.

### DIFF
--- a/src/system/libretro/tic80_libretro.c
+++ b/src/system/libretro/tic80_libretro.c
@@ -78,6 +78,22 @@ struct tic80_state
 static struct tic80_state* state;
 
 /**
+ * TIC-80 callback; Request counter
+ */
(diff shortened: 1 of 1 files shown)


jezze:
This have been validated to work on retroarch and fixes https://github.com/nesbox/TIC-80/issues/2065

joshgoebel (approved):
Looks about like what I'd expect.

MineRobber9000:
Wouldn't this cause the same issues we were having earlier with `clock()` not being wall-time (see #1904)?

joshgoebel:
Oh wait, yes what is `clock()` in libretro here?  If it's just the same old LIBc stuff then no, it won't work.  We need to be counting wall time.

nesbox (approved):

PR #2079, opened 2023-01-11
janet support

AlecTroemel:
possible replacement for https://github.com/nesbox/TIC-80/pull/2069, builds off the work of @dressupgeekout. All functions bindings are complete, things left to do:

- [x] make tic80 a module so the user can choose between `(import tic80)` or `(use tic80)`. This is to avoid function name collisions of `print` and `map`
- [x] fix bug where random characters show up at end of errors 
- [x] get all builds passing


nesbox:
Great, new language support is a good reason to release a new TIC-80 version.
Thank you 👍

AlecTroemel:
To the people watching this PR, I think I'm going to need some assistance with the failing builds.

- for Windows builds, I'm really flying blind here.. the logs in the build pipeline don't really make much sense to me.. and VS in a windows VM no being no clearer. Janet's desired way to be built in windows is to run `build_win.bat`, and my best bet is cmake is having problems running that script? https://github.com/nesbox/TIC-80/actions/runs/3969162081/jobs/6803278488#step:4:1780
- for RPI baremetal and 3ds, it looks like there's some sort of datatype mismatch between the main header file and actual source (not sure if I'm even saying that right). Its all related to IO/file related functions https://github.com/nesbox/TIC-80/actions/runs/3969162081/jobs/6803278106#step:6:500, which is maybe the real problem.. are file operations available on baremetal?

any help or points in the right direction would greatly appreciated, as I'm pretty stumped..

AlecTroemel:
With the help of the great people over in the janet Gitter chat, all platforms builds are now passing! I've marked this PR as ready for review

nesbox (approved):

PR #2117, opened 2023-02-05
Code Editor upgrades: structured s-expresion edition and emacs-mode

sthilaid:
Hi! This pull request adds two main features:

- = Structured edition =
This functionality mimics some functionality provided by the great paredit-mode in emacs. It helps s-expression (lisp like language code structure) edition by giving you tools to create / remove expressions and enforces (to a certain extent) the balance of parenthesis. It is still missing an automatic code alignment, but this might come in later. 

- = emacs-mode =
This code editor mode changes a lot of bindings to be a lot more compatible with emacs users. It also brings some new features like deleteLine, startLine and some modified functionnality for others. It also fixes what I believe is a bug in endLine (which before the fix would bring you to the end of the code buffer). To enable this, I added a new submenu called Code Editor Options which has this "emacs mode" option embeded.

- = tic_script_config additions =
I added options in the language config to customize the isalnum function used when parsing their code. For instance, scheme allows a lot of normally forbidden characters in their symbols (stuff like ':' or '#'). This allows any language to customize it. In a similar fashion, there is also an option to override the API function names for that language so that the code highlight will correctly recognize them if they were modified for that language (for example, if there is a prefix). Finally, there is an option to enable structured edition for that language.

Let me know what you think!


AlecTroemel:
@sthilaid what are your thoughts moving the cursor to the inner s-expression  `ctrl-down`? I havent done much "structural editing", but It was surprising to not be able to undo a `sexpify` without first hitting `right`.

AlecTroemel:
oh also, would we be able to add quotations `"` to the list of "can only delete if the innards are cleared"?

sthilaid:
@AlecTroemel, very good suggestions, i'm going to implement both of those right away!

sthilaid:
done and done! it was a bit more work to add the double quotes to structured edition but it was an involuntary omition on my end! good catch! From my tests, it seems to be working well, hopefully there isn't too many holes in my logic.

src/studio/config.c
```diff
@@ -238,6 +238,7 @@ static void setDefault(Config* config)
             .volume         = MAX_VOLUME,
             .vsync          = DEFAULT_VSYNC,
             .fullscreen     = false,
+            .emacsMode      = false,
```
nesbox, line 241: Could you pls move all the emacs mode declarations under `#if defined(BUILD_EDITORS)` definition, which will allow not including this option for special builds without editors?

src/studio/screens/mainmenu.c
```diff
@@ -218,7 +242,8 @@ static const MenuItem OptionMenu[] =
     {"VSYNC",           NULL,   &VSyncOption, "VSYNC needs restart!"},
     {"FULLSCREEN",      NULL,   &FullscreenOption},
     {"VOLUME",          NULL,   &VolumeOption},
-    {"SETUP GAMEPAD",   showGamepadMenu},
+    {"SETUP GAMEPAD",       showGamepadMenu},
+    {"CODE EDITOR OPTIONS", showCodeEditorMenu},
```
nesbox, line 246: The same here, pls wrap the option declaration with `#if defined(BUILD_EDITORS)`

src/studio/system.h
```diff
@@ -128,6 +128,7 @@ typedef struct
         bool vsync;
         s32 volume;
         tic_mapping mapping;
+        bool emacsMode;
```
nesbox, line 131: And here, please :)

nesbox (changes requested):

sthilaid:
There seems to be some issues with the actions today, I don't know how to resolve these problems. For instance:
```Err:1 http://archive.ubuntu.com/ubuntu bionic/main amd64 fonts-lato all 2.0-2
  Connection failed [IP: 185.125.190.39 80]```

sthilaid:
lol, okay that worked... seems like the build action or not super reliable! ;p

nesbox (approved):
Seems everything looks good
Thank you

sthilaid:
I would like to add some documentation about these changes, when should I modify the wiki to explain them. When will version 1.1 be released? (no point in adding the doc if it's not in the users' hands)

sogaiu:
I was looking at [this](https://github.com/nesbox/TIC-80/projects/6).

nesbox:
Yes, GitHub Issues is our roadmap, 
as for adding documentation, pls create a new page in the Wiki and add a link to it on the Code Editor page https://github.com/nesbox/TIC-80/wiki#code-editor

sthilaid:
Excellent, I added the info to the wiki, thanks!

Issue #2131, opened 2023-02-17
WASM map() function can't be called with sx or sy = -1


Madadog:
# Problem
From the WASM API (tested with Rust), calling the `map()` function with sx or sy set to -1 replaces them with the value 0. This is not consistent with the Lua API, where -1 is a valid value. It also creates a visual discontinuity when sx or sy cross over -1.

Demonstration of visual glitches resulting from this:
![video9](https://user-images.githubusercontent.com/53168336/219581017-b7fb775d-9f42-4d39-9804-89656f04d2f2.gif)

# Solution
Remove this default-setting code from the WASM API:
https://github.com/nesbox/TIC-80/blob/b4935a63528b152fd3ccd015fee22da6b9430e6c/src/api/wasm.c#L688-L689

This does not affect any of the languages with an existing TIC80 template (C, D, Rust and Zig) because all arguments to `map()` must be supplied by the caller, or there will be a compilation error.

For any languages where function parameters can be omitted, this might change the default value of sx/sy to -1 (assuming `m3ApiGetArg()` returns -1 if an argument is missing, though I am not actually sure about this) but this can probably be worked around at the language API level when their template is added.

joshgoebel:
I see two major choices here.  Don't have defaults - and at the WASM binary API level you MUST specify the full params...leaving wrapper functions to make the API nicer (or not).  I've already tried to adopt this approach in Zig with it's structs and pushing defaults values into the lib.

Or (since WASM is really 32bit minimally) you could change all int type params at the WASM level to i32 or u32 and have a very large sentinel value that's used as a stand-in for "DEFAULT".  Of course that only works for int/bools... pointers would get more complex...

I think the first may be preferable and push this complexity into the wrapper libs... perhaps languages provider a `mapAll()` for "all defaults" and `map(*args)` for the full API, and nothing in-between... static languages often simple suck at this sort of thing more than dynamic languages. Or perhaps languages adopt a fully struct based param passing for non-required args (as the Zig wrapper does).

nesbox:
In this case, I'd like to vote for **"Don't have defaults - and at the WASM binary API level you MUST specify the full params"** and define all the defaults on the wrappers level.

Grissess:
[...]

PR #2132, opened 2023-02-17
Remove defaults for sx/sy in WASM map() function

Madadog:
Solves https://github.com/nesbox/TIC-80/issues/2131.

This enables calling `map()` from the WASM API with sx or sy=-1, which is consistent with the Lua API. 

This does not break any of the current languages with templates (C, D, Rust, Zig) because arguments to `map()` cannot be omitted. However, for future languages which allow optional arguments, this may require workarounds at the language API level to get sx/sy=0 by default, if `m3ApiGetArg()` defaults to -1 on missing arguments (I do not understand how `m3ApiGetArg()` works, but the other defaults in wasm.c seem to assume a `int32_t` with a value of `-1` from `m3ApiGetArg()` means a missing argument).

### DIFF
--- a/src/api/wasm.c
+++ b/src/api/wasm.c
@@ -685,8 +685,6 @@ m3ApiRawFunction(wasmtic_map)
     if (y == -1) { y = 0; }
     if (w == -1) { w = 30; }
     if (h == -1) { h = 17; }
-    if (sx == -1) { sx = 0; }
-    if (sy == -1) { sy = 0; }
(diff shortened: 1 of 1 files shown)


nesbox (approved):
I'll merge this PR but I think we have to find a better solution to detect default variables in wasm.
Because I see the same problem for `x = -1` or `y = -1` for example.
@joshgoebel what do you think?

PR #2135, opened 2023-02-20
New Pause() api function

sthilaid:
Hi! I wanted this functionality to ease debugging. This is not quite a breakpoint but is helpful nonetheless. What do you think? (I only implemented it in scheme for now to not waste time if this will be rejected).

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -35,12 +35,14 @@ typedef struct { u8 index; tic_flip flip; tic_rotate rotate; } RemapResult;
 typedef void(*RemapFunc)(void*, s32 x, s32 y, RemapResult* result);
 
 typedef void(*TraceOutput)(void*, const char*, u8 color);
+typedef void(*PauseCallback)(void*, const char*);
 typedef void(*ErrorOutput)(void*, const char*);
 typedef void(*ExitCallback)(void*);
 
 typedef struct
 {
     TraceOutput trace;
+    PauseCallback pause;
     ErrorOutput error;
     ExitCallback exit;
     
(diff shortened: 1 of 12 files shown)


joshgoebel:
Feels like something you could easily build yourself outside the API just set aside a secondary controller for controlling things like pause, next frame, etc...

nesbox:
I think you could do the same by calling 
```
trace('info')
exit()
```

sthilaid:
@nesbox your suggestion is almost the same, BUT it seems that after `exit` is called, `resume` does not work (it doesn't restart the execution, or is it just me that is missing something?

sthilaid:
@joshgoebel, how would you proceed to do this outside of the tic80 codebase? I'm not sure I understand your suggestion.

joshgoebel:
> for controlling things like pause, next frame, etc...

Simply build these concepts into your program itself.  So it can pause, and then allow you to advance frame by frame. A simple state machine is all you need.

nesbox:
Why if we modify the `exit()` api and make game able to resume after the call?

BuoYancYdabl:
> Why if we modify the `exit()` api and make game able to resume after the call?

Most of the games that are logically quits after ending and user launches them again with resume: am I joke to you?

sthilaid:
As @BuoYancYdabl mentioned, it seems like the exit method should keep it's semantic. 

@joshgoebel I see, indeed it would work, but it seems like a fair amount of overhead that each program needs to implement. It would be nice to be implemented at the API/core level. 

My implementation doesn't work very well, it was more a test and gather feedback and interest. It seems a lot of you are reluctant, so let's just forget about it :)

joshgoebel:
> but it seems like a fair amount of overhead that each program needs to implement.

I really don't think it's that much effort, but perhaps I'm not thinking of something...

> It would be nice to be implemented at the API/core level.

I don't think we're going to add a full on debugger, so I honestly don't see the value for MOST users to be able to frame step without a full debugger.

PR #2144, opened 2023-02-26
Added support for fget to return all flags

sthilaid:
A bitfield return when no flag is passed as paramater (similarly to the pico-8 api). I tested all languages except for wasm which I honestly don't know how to test.

### DIFF
--- a/src/api/wasm.c
+++ b/src/api/wasm.c
@@ -412,7 +412,7 @@ m3ApiRawFunction(wasmtic_keyp)
 
 m3ApiRawFunction(wasmtic_fget)
 {
-    m3ApiReturnType  (bool)
+    m3ApiReturnType  (int8_t)
 
     m3ApiGetArg      (int32_t, sprite_index);
     m3ApiGetArg      (int8_t, flag);
--- a/src/api/wasm.c
+++ b/src/api/wasm.c
@@ -419,7 +419,14 @@ m3ApiRawFunction(wasmtic_fget)
     
     tic_mem* tic = (tic_mem*)getWasmCore(runtime);
 
-    m3ApiReturn(tic_api_fget(tic, sprite_index, flag));
+    if (flag == -1)
+    {
+        m3ApiReturn(tic_api_fget_all(tic, sprite_index));
+    }
+    else
+    {
+        m3ApiReturn(tic_api_fget(tic, sprite_index, flag));
+    }
 
     m3ApiSuccess();
 }
--- a/src/core/draw.c
+++ b/src/core/draw.c
@@ -450,6 +450,12 @@ bool tic_api_fget(tic_mem* memory, s32 index, u8 flag)
     return *getFlag(memory, index, flag) & (1 << flag);
 }
 
+u8 tic_api_fget_all(tic_mem* memory, s32 index)
+{
+    static const u8 dummyFlag = 0;
+    return *getFlag(memory, index, dummyFlag);
+}
+
 void tic_api_fset(tic_mem* memory, s32 index, u8 flag, bool value)
 {
     if (value)
--- a/src/api.h
+++ b/src/api.h
@@ -806,11 +806,12 @@ enum
                                                                                                                         \
                                                                                                                         \
     macro(fget,                                                                                                         \
-        "fget(sprite_id flag) -> bool",                                                                                 \
+        "fget(sprite_id flag) -> bool\nfget(sprite_id) -> flags",                                                       \
                                                                                                                         \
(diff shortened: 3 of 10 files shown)


src/api/wasm.c
```diff
@@ -419,7 +419,14 @@ m3ApiRawFunction(wasmtic_fget)
     
     tic_mem* tic = (tic_mem*)getWasmCore(runtime);
 
-    m3ApiReturn(tic_api_fget(tic, sprite_index, flag));
+    if (flag == -1)
```
joshgoebel, line 422: Minimally `m3ApiReturnType` would need to change to `uint8_t` and we'd need to make sure that doesn't break the binary API of any of the existing target WASM languages.  If they typecast 0 to falsy and non-zero to truthy we might be ok here.

sthilaid: hey @joshgoebel, good catch, I missed the return type statement. Could I do something like 
```
if (flag == -1) {
    m3ApiReturnType(int);
    m3ApiReturn(tic_api_fget_all(tic, sprite_index));
    } else {
        m3ApiReturnType(bool);
        m3ApiReturn(tic_api_fget(tic, sprite_index, flag));
    }
```
maybe? What do you think?

joshgoebel: No, the WASM<=>outside API must be static.  The params (type and number) must always be the same, the return type must always be the same.  But when I say "binary compatibility" I'm referring to what any cartridges are already expecting that have been compiled.

All these params are really u32 or u64 going across the wire (the only true WASM primitives)... but how they are interpreted on the WASM VM side is what matters.  You can't change that by changing TIC-80...

That's the only reason I suggest we might be able to get away with changing `bool` to `uint_8t` here... it all depends on how the various compilers are typecasting those values.

sthilaid: ok yes, I see what you mean. How would we proceed about this then? It could also become a new api function, like `fgetall` or something like that. Like I mentioned, since I don't know how to test wasm, and apparently, it seems kindof impossible to test if there are an unknown amount of compilers that compiles to wasm, we're a bit in a pickle in terms of quality assurance...

joshgoebel: Only the ones we have libs for would need to be tested.  That's only like 2-3?

sthilaid: ok i'll update the pull request then. Do you know how to test them?

joshgoebel: Sure.  Write a simple program in each language that calls the API expecting it to return a bool - printing both cases to the screen, etc. Compile and test.

Then run that same binary cartridge on a new compiled version of TIC-80 that's changed the binary API, and see if it still works as expected.  The wiki has instructions for working with WASM cartridges.

src/core/draw.c
```diff
@@ -450,6 +450,12 @@ bool tic_api_fget(tic_mem* memory, s32 index, u8 flag)
     return *getFlag(memory, index, flag) & (1 << flag);
 }
 
+u8 tic_api_fget_all(tic_mem* memory, s32 index)
+{
+    static const u8 dummyFlag = 0;
```
nesbox, line 455: I don't understand why you use a `static` variable here.

sthilaid:
It's too explicit a hard-coded value. Just passing zero directly is
obscure, but using this named zero makes it obvious that it's just a dummy
value.

sthilaid:
Ok, to be honest, I don't know how to test the wasm languages and I don't really want to spent the time figuring it out. Is there anyone setup to test those easily that could grab this commit and do a quick test? Otherwise I might just drop this pull request...

sthilaid:
All right, I tried, I really did, to test the wasm demos but my free time is not worth wasting on this. For the C demo, just downloading the git repo required and resolving the git state exceeded my patience. For zig, there is a missing 'game.tic' file when following the README instructions, the C required SDK is way too big for me to have the patience to get, install, compile and use, and I would need a special nightly version of rust wich I don't want to install.

It's too bad because adding to the API is fairly straightforward despite having multiple languages but the WASM stuff is a serious deterrent for me.

joshgoebel:
`tic80 --fs . --cmd 'load game.tic & import binary zig-out/lib/cart.wasm & save'`

If this had said `your_game.tic` would that have been clearer?  This is a tough challenge, but at the very least we could improve the docs if that would help the next person. 

I'd be happy to help with the WASM testing (at least for Zig - which I've hacked on before) but not sure when I could commit the time... making changes to the WASM API (typically a no, no period) is definitely a real challenge.  Everyone not having to deal with binary compatible APIs should be thankful. :-)

Issue #1913, opened 2022-05-04
Flashing colors in surf menu


ginger-brave:
When you open a public game, colors start flashing

https://user-images.githubusercontent.com/104861287/166734153-e91a6234-efde-44ae-9204-be8036995ad7.mp4

BuoYancYdabl:
epileptic-friendly game XD

grigoriynikolaev:
So it isn't a bug, but a feature? Is there a way to turn it off system wide? Not an epileptic, but necessity to squint every time it happens is quit annoying.

joshgoebel:
That looks like the animation boot sequence, but if we aren't shooting the boot screen then we shouldn't be doing the animation.

nesbox:
Looks like a bug

PR #2149, opened 2023-03-02
Clamp input to animEffect()

Madadog:
Fixes https://github.com/nesbox/TIC-80/issues/1913.

animEffect() seems to expect a value between 0 and 1, and #1913 is caused by the surf coverFade animation lerping past 1, which feeds values greater than 256 into fadePalette(), causing the flashing colors. This PR clamps movie ticks for each item ensuring animEffect won't receive values > 1.

This could create issues if any animation items rely on playing past their end. I didn't notice any while visually checking the editor and surf menu animations, but I could easily have missed something.

It might be better to move this clamp directly to the start of animEffect().

Alternative fixes for #1913 include clamping the fadePalette() value between 0 and 256, or stopping the animation at the right time.

### DIFF
--- a/src/studio/studio.c
+++ b/src/studio/studio.c
@@ -914,7 +914,10 @@ static inline float animEffect(AnimEffect effect, float x)
 static void animTick(Movie* movie)
 {
     for(Anim* it = movie->items, *end = it + movie->count; it != end; ++it)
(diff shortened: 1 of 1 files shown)


nesbox (approved):

nesbox:
Thank you for pointing me to where this is going, because I've tried to investigate issue #1913 without success.

PR #2200, opened 2023-05-15
Convert `vendor/pocketpy` into a git submodule

blueloveTH:
See #2184


blueloveTH:
For submodule intergration, it requires `python3` command to run pre-build scripts.
However, `rpi`, `rpi-pro`, etc.. containers does not have `python3` command.
So the build process failed on these platforms.

blueloveTH:
Also, in android, the system python version is a bit old, and some standard functions are missing.

Maybe, we can use an environment that supports cpython 3.8+ to run the pre-build script, and then move the output to the container.

nesbox (approved):

Issue #2190, opened 2023-05-10
Bank parameter does not work for exporting music


borbware:
The `bank` parameter doesn't work for exporting music tracks. Regardless of its value, the music track with the id corresponding to the `id` parameter from the **currently active bank** is exported

Thus, it's not possible to export music from other banks when using TIC-80 from the command line.

joshgoebel:
How are you defining currently active?

borbware:
![image](https://github.com/nesbox/TIC-80/assets/19384640/6609afd7-f199-4826-85f8-7f576490968d)
the one shown on red in the music editor.

bztsrc:
> The bank parameter doesn't work for exporting music tracks. Regardless of its value, the music track with the id corresponding to the id parameter from the currently active bank is exported

According to the code, the function `tic_cart_save` does not have any reference to the "currently active bank" at all. It saves music patterns and tracks like this:
```c
        buffer = SAVE_CHUNK(CHUNK_PATTERNS, cart->banks[i].music.patterns,  i);
        buffer = SAVE_CHUNK(CHUNK_MUSIC,    cart->banks[i].music.tracks,    i);
```
Where `i` is a loop variable and given a value [here](https://github.com/nesbox/TIC-80/blob/main/src/cart.c#L307):
```c
    for(s32 i = 0; i < TIC_BANKS; i++)
```
I can't see how this code could only save the current bank, except `SAVE_CHUNK` checks and skips fully empty (filled with just zeros) banks.

Are you absolutely sure you have valid music data in the other banks too, not just in one?

Hope this helps,
bzt

borbware:
Here's an example cartridge so you can try it for yourself if I'm missing something here.
[exportbankbug.zip](https://github.com/nesbox/TIC-80/files/11466766/exportbankbug.zip)

Consider these commands:
1) `export music megalo.wav id=0 bank=0`
2) `export music fanfare.wav id=0 bank=1`

If you run both the commands, they will export the same song, the one on bank 0 (because it's active by default).
If you go to music editor, disable bank sync (the red ankh symbol), and change the active bank to 1, the song on bank 1 is exported instead.

You can also choose to not disable bank sync: you will still get the song on bank 1 with both the commands, but it will be composed of noise, because the instrument on bank 1 is not defined.

borbware:
[...]

PR #2202, opened 2023-05-17
Fix bank parameter for music export

borbware:
This commit fixes the bank parameter for exporting music (see issue #2190).

There is still some ambiguity present, because the exported music depends on the sfx bank as well. 

If bank chaining is on, the exported music uses the same sfx bank as the music bank.
If bank chaining is off, the exported music uses the sfx from the currently active sfx bank.

This could be fixed by introducing a `chained=true/false` parameter for the export.


borbware:
Note: I didn't manage to call the static `switchBank` function, so I implemented some of its functionality inside `studioExportMusic`.

nesbox (approved):

PR #2204, opened 2023-05-21
Update Rust template to remove optimisation issue

soxfox42:
Previously I configured the Rust template to disable optimisation due to issues with accessing memory directly. Instead, I think it's best to mention Rust's volatile read and write functions in the README. These allow reliable access to MMIO, so are suitable for memory access in TIC-80.


joshgoebel:
> Don't access memory directly.

This seems weird to say when we go right ahead to explain HOW to do it.  I think the wording here could be improved somehow.

soxfox42:
Fair point, I've updated it to make it clear this only refers to deref-ing raw pointers, not the `_volatile` functions.

PR #2212, opened 2023-06-04
update pocketpy to  v1.0

koltenpearson:
also adjusted cmake to handle the python3 build dependency in a more cross platform way


nesbox (approved):
thank you :)

PR #2217, opened 2023-06-11
adds editor options for indentation with spaces or tabs 

koltenpearson:
lets you set a tab size, as well as tabs vs spaces


nesbox (changes requested):
I just found some strange behavior
video47
When I reset all the settings using `config reset` command I saw some strange behavior in the code editor as in the video above. I pressed the tab key but UI wasn't updating, but if you try to select code, you will see the tabs.
Could you please fix this before merging?

koltenpearson:
Looks like I just forgot to set the right default tab size, so it defaulted to 0

nesbox (approved):

PR #2234, opened 2023-07-14
Run key processing even when ctrl key is pressed

blind:
I noticed the keyboard shortcuts in the music tracker using the ctrl key did not work. Turns out the code was never called if ctrl was pressed.

Moved the key processing in music tracker out of the `!ctrl` scope.
Added early exit from `processPianoKeyboard()` if ctrl or alt keys are pressed, which mirrors the way it's done in `processPatternKeyboard()`.


nesbox (approved):
thank you

Issue #2260, opened 2023-08-14
Rust template instructions not working with 1.1.2692-dev


villesundell:
Hello!
The instructions at https://github.com/nesbox/TIC-80/blob/main/templates/rust/README.md seem outdated. In addition to clear syntax issues ("card name is missing"), there seem to be no clear instructions on how to do continuous development (as I always get the default Hello World application despite my changes). I am running `1.1.2692-dev (02fd0eb)`.

villesundell:
Solved this by changing the command to: `import binary target/wasm32-unknown-unknown/release/cart.wasm`. This way, a new WASM cartridge will be created, and a WASM file can be imported so that it would be executed instead of the default Lua script. Will make a PR on this.

PR #2261, opened 2023-08-14
#2260: Updated Rust template build instructions

villesundell:
It turned out, that a new WASM cartridge must be created prior to saving a new one, in order to prevent the default Lua code running, instead of the WASM code.


nesbox (approved):
thank you

PR #2298, opened 2023-09-08
Update zig examples and template to latest Zig

tupini07:
The previous Zig `build` and `tic80.zig` files built fine with Zig 0.10, but no longer build with the latest _stable_ (0.11) nor the current development version (0.12). 

This PR updates the Zig examples so that they build with both 0.11 and 0.12 😄 

I've also formatted the files in question with Zig's language server formatter.

### DIFF
--- a/demos/bunny/wasmmark/build.zig
+++ b/demos/bunny/wasmmark/build.zig
@@ -1,19 +1,27 @@
 const std = @import("std");
 
-pub fn build(b: *std.build.Builder) void {
-    const mode = b.standardReleaseOptions();
+pub fn build(b: *std.Build) !void {
+    const optimize = b.standardOptimizeOption(.{});
 
-    const lib = b.addSharedLibrary("cart", "src/main.zig", .unversioned);
-    lib.setBuildMode(mode);
-    lib.setTarget(.{ .cpu_arch = .wasm32, .os_tag = .freestanding });
+    const lib = b.addSharedLibrary(.{
+        .name = "cart",
+        .root_source_file = .{ .path = "src/main.zig" },
+        .target = .{ .cpu_arch = .wasm32, .os_tag = .wasi },
+        .optimize = optimize,
+    });
+
     lib.import_memory = true;
+    lib.stack_size = 8192;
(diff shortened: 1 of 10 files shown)


demos/bunny/wasmmark/build.zig
```diff
@@ -1,19 +1,27 @@
...
     lib.export_table = true;
+
     // all the memory below 96kb is reserved for TIC and memory mapped I/O
     // so our own usage must start above the 96kb mark
     lib.global_base = 96 * 1024;
-    lib.stack_size = 8192;
-    lib.export_symbol_names = &[_][]const u8{ "TIC", "OVR", "BDR", "SCR", "BOOT" };
-    lib.install();
+
+    lib.export_symbol_names = &[_][]const u8{ "TIC", "OVR", "BDR", "BOOT" };
```
joshgoebel, line 24: Has SCR been entirely removed now or just deprecated?  If the latter we many want to leave it here for backwards compatibility...

demos/bunny/wasmmark/src/main.zig
```diff
@@ -105,81 +100,83 @@ fn addBunny() void {
...
+    // var x : u32 = 100000000;
+    // tic.FRAMEBUFFER[x] = 56;
 
-// testscreen();
+    // testscreen();
 }
 
+export fn BDR() void {}
+
+export fn OVR() void {}
```
joshgoebel, line 182: Are these necessary now to compile?

demos/bunny/wasmmark/build.zig
nesbox: I don't think someone used it, first of all, it was initially called SCN, and second, most langs don't support SCN (only Lua and JS). We can remove it without any doubts.

nesbox (approved):

demos/bunny/wasmmark/src/main.zig
tupini07: yep, if these are not present then the Zig compiler will complain that it couldn't find the symbols we're telling it to export.

Issue #2291, opened 2023-09-02
[Bug] [JavaScript] Unexpected `spr()` rendering with `undefined` colorKey


scambier:
The [`spr()`](https://github.com/nesbox/TIC-80/wiki/spr) function has 3 required parameters, and several optional parameters that have default values.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters):
> Default function parameters allow named parameters to be initialized with default values **if no value or `undefined` is passed.**

i.e.
```js
// calling
spr(1, 1, 1)
// should be equivalent to
spr(1, 1, 1, undefined)
// which should have the same result as
spr(1, 1, 1, -1)
// since -1 is the default value for colorKey
```

But that's not the case with TIC-80, _at least_ for the `colorKey`, and probably for other functions with optional params; when not specified, the default value is `-1`, but when explicitly called with `undefined`, the value actually used is `0`

```js
// calling
spr(1, 1, 1, undefined)
// is actually equivalent to
spr(1, 1, 1, 0)
```

This is related to #2268, as this is the cause of my transparent sprite. I _think_ the bug predates QuickJS.

Eulenmensch93:
I created a pull request to fix this issue.
After the fix all variants should behave the same way.

```
function TIC()
{
    cls(13)
    spr(1,80,20)
    spr(1,80,40,undefined)
    spr(1,80,60,-1)
    spr(1,80,80,[undefined])
}
```
Before the fix, the first two versions would use index 0 as the transparent color.

I tried to check the implementation of lua and python to see how they handle this case.
But both of them define an array `static u8 colors[TIC_PALETTE_SIZE];` and assign in some value if a parameter is passed `colors[i-1] = getLuaNumber(lua, -1);`. To me it looks like the array is not initialized at all (random memory), or 0. I am not sure if all the C and C++ compiler handle the array creation the same way.

I also wrote some test lua code to see how the function behaves there, and the behavior is totally different.
```
function TIC()
    cls(13)
    spr(1,80,20)
    spr(1,80,20,-1)
    spr(1,80,20,{})
end
```
Now in the first case the color 0 is visible, and in both other cases the 0 index color is not visible.
The second case might be a problem because the color table is a u8 array, so since it is unsigned the -1 will be a gigantic number.

I think the function needs to be checked for correctness in all languages.

PR #2309, opened 2023-09-21
Fix issue 2291

Eulenmensch93:
As described in issue 2291, the code to handle the colorkey parameter in the spr function did not use the correct default value in one of the two code paths.

We now pass the correct default value to getInteger2, this handles cases where the parameter is not present, and where the parameter is `undefined` in JS.


TimotheeGreg:
Let me just add a link to: issue #2291.

nesbox (approved):

Issue #2296, opened 2023-09-05
line function not working on python


rodigu:
OS: Windows
Version: `1.1.2736 Pro`
Code:
```py
def TIC():
 line(0,0,2,2,2)
```
![image](https://github.com/nesbox/TIC-80/assets/58960796/191e8892-7727-45ea-9abd-fa09528659f0)

PR #2314, opened 2023-09-26
Fix issue 2296

Eulenmensch93:
The Python API does not accept integer values for floating point parameters.
So `line(0,0,100,100,3)` will cause an error but `line(0.0,0.0,100.0,100.0,3)` will
work.

To fix this a function called  `to_float_wrapper` is implemented so that we can
first check if the parameter is an integer number and if yes implicitly cast it into
a double.

The pull request changes the behavior of all Python API functions which accept
floating point numbers.

See Issue: #2296

### DIFF
--- a/src/api/python.c
+++ b/src/api/python.c
@@ -7,6 +7,21 @@
 #include <stdio.h>
 #include <string.h>
 
+bool to_float_wrapper(pkpy_vm* vm_handle, int index, double* ret)
+{
+    // Implicit convert of int to double
+    if( pkpy_is_int(vm_handle, index) )
+    {
+        int temp;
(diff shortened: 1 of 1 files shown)


blueloveTH:
I think https://github.com/nesbox/TIC-80/pull/2315 can solve this. `v1.2.x` implements automatic float cast inside the api.

Eulenmensch93:
> I think #2315 can solve this. `v1.2.x` implements automatic float cast inside the api.

If this is the case and the issue is solved with the new version, we can close this pull request.

@nesbox 
Do you see any use in my pull request or can I close it?

nesbox:
@Eulenmensch93 thanks for the help, it seems it has already been fixed in PR #2315

PR #2318, opened 2023-09-29
Add a few more emacs bindings

non:
This commit adds the following key bindings:

 - Ctrl + '_': undo
 - Ctrl + '<': beginning of document
 - Ctrl + '>': end of document

These are all standard Emacs key bindings.


sogaiu:
Interesting, I didn't know you could do `Ctrl + '_'` to undo. TIL :)

I've been using `Ctrl + '/'`.  The former (with underscore) requires pressing the Shift key on the kinds of keyboards I've been using, while the latter does not.  Perhaps that's different for other folks' environments.

I don't think the latter two (with the inequality symbols) are standard keybindings.  `Alt + '<'` and `Alt + '>'` are though AFAIK.  `Ctrl + '<'` and `Ctrl + '>'` show up as unbound when I do `emacs -q`.

nesbox (approved):

sogaiu:
I think it would be better if there were some discussion about these keybindings...

Issue #722, opened 2018-09-07
Octave shortcut collides with cartridge save shortcut


nesbox:
@JasonNumberThirteen says
> I noticed when everytime I want to save my project while being in SFX editor using Ctrl+S, it automatically changes sound octave to C# because it's attached to the "S" key which causes collision. IMHO, it might be nicely to make some detection which changes octave only when user doesn't hold the Ctrl key.

Skeptim:
This can be marked as done thanks to https://github.com/nesbox/TIC-80/pull/2336

PR #2336, opened 2023-10-24
Increase in possible notes for playing on piano roll, new keyboard layout (like in popular DAWs)

jejikeh:
Slightly changed the keyboard in the SFX editor so that it became more similar to the popular DAWs (Ableton, Garage Band, etc...)

1. Added the ability to switch between octaves on Z and X
2. Raised all the shortcuts one row higher (A, W, S, E, D, ...)
3. Increase the number of possible notes that can be played on the keyboard

### DIFF
--- a/src/studio/editors/music.c
+++ b/src/studio/editors/music.c
@@ -1102,6 +1102,9 @@ static void processTrackerKeyboard(Music* music)
                 resetSelection(music);
         }
 
+    static const u8 newOctaveIndex = 16;
+    static const u8 keboardShift = 5;
+
     static const u8 Piano[] =
     {
         tic_key_z,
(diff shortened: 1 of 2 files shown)


TimotheeGreg:
The same hotkeys apply on the tracker mode of the music editor. They should probably be consistent.

jejikeh:
I added support for playing on the second octave in SFX piano roll as well as in Music Editor. Added shortcuts to change the octave.

1. To play on 1 octave press: Z../
2. To play on 2 octave press: Q..P
3. To change octave press: shift+z / shift+x

TimotheeGreg:
I came across this today: #722 
That could be an issue to check before merging, or in the other way, your pull request could solve it.

jejikeh:
Since in the SFC editor you can play notes by holding the keys. The note is played if you release ctrl before S. 

This can be fixed, but I'm afraid that the code will not be very good :\ (while it occurs to me to create a variable that will store the value of S while ctrl was pressed).

nesbox (approved):

PR #2347, opened 2023-11-07
Fix support for the Xcode generator

DidierMalenfant:
Here's a few small tweaks to generate an Xcode project files.

The biggest change here is updating the zip library to the latest version.

Since I'm new to contributing to this project, I welcome any feedback or suggestions on how to make sure this change doesn't break other platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -434,8 +434,8 @@ set(GIFLIB_SRC
 add_library(giflib STATIC ${GIFLIB_SRC})
 target_include_directories(giflib
     PRIVATE ${GIFLIB_DIR}
-    INTERFACE 
-        ${THIRDPARTY_DIR}/giflib 
+    INTERFACE
+        ${THIRDPARTY_DIR}/giflib
         ${THIRDPARTY_DIR}/msf_gif)
 
 ################################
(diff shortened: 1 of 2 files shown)


DidierMalenfant:
Ok it looks like the targets that are failing are because of the higher cmake version required brought by the latest zip library.

Any particular reason why those targets are stuck at an older make version?

nesbox:
We use Docker containers to build some targets, so you will need to update Cmake in those containers or just make changes for Xcode only.

DidierMalenfant:
Makes sense. I can totally look into that.

Just making sure that you're ok with updating some of the dependencies if necessary. I already have to do this on a couple of them to fix issues with iOS support.

TimotheeGreg:
That would be great to have this!

DidierMalenfant:
> That would be great to have this!

I've still got some of the docker images to update but it's looking promising.

DidierMalenfant:
Was this automatically closed? I'm still working on getting the docker images updated...

TimotheeGreg:
Cool! I guess nesbox thought you were not working on it any more.

DidierMalenfant:
Sorry it's taking me so long. Limited time to work on this but I am making progress.

TimotheeGreg:
No problem, take your time

DidierMalenfant:
So I tried my best but this may be a little bit above my pay grade unfortunately.

The changes to get the Xcode generator are daily minor but they require a more recent version of cmake and that breaks the CI.

Some of the existing docker images are easy to upgrade to a newer version like the linux ones which can just use a newer ubuntu, but unfortunately some docker images don't build any longer.

devkitpro doesn't release deb files on GitHub anymore. I managed to inline a copy of [this](https://apt.devkitpro.org/install-devkitpro-pacman) script to rebuild the docker image correctly.

SDL changed their build system to GitHub actions and do not seem to post rpi builds anymore. I tried to follow [this](https://github.com/libsdl-org/SDL/blob/main/docs/README-raspberrypi.md#cross-compiling-from-x86-linux) to update the docker images but I couldn't get it to work.

Android action doesn't build locally for me and I couldn't figure out why.

I'm going to humbly give up here. if someone with more knowledge than me want to point me in the right direction then I'l be happy to put some more hours into this but otherwise this can be closed for now.

Issue #1873, opened 2022-04-07
VSYNC makes HTML export run fast when on monitors with high refresh rates? (latest dev version)


Gate88:
I'm currently using the HTML export feature to embed my TIC-80 game for playtesting on itch.io. I noticed with the current stable version (version 0.90.1723 Pro (9c38a80)) the audio gets really choppy and slows down the game when in the background.

To try to fix this, I pulled the latest version (version 1.0.2128-dev Pro (551cf2d)) and the audio seems a lot better! However, I'm having an issue that seems specific to the HTML export (the executable export seems to work fine) that was not happening in the previous stable version.

On a high refresh monitor (144hz) the game runs super fast in the HTML export. If I press ESC, go to "options" and turn off "VSYNC" then it seems to run normally again. I think VSYNC may cause the game to run too fast in the HTML export when on a high refresh rate monitor. This does not seem to happen in the native .exe, nor when I export natively.

If you have a high-refresh rate monitor, you can see the issue reproduced here:
https://gate.itch.io/tic80-workspace?password=new-tic-80-build

Also, is there any way to get the HTML export to launch with different default settings (e.g. VSYNC off or different controls)?

nesbox:
Unfortunately, we can't vsync properly on a high refresh rate monitor, alternatively, we can make `vsync=false` the default in the HTML export, or you can use the command line argument `--vsync 0` to turn it off in your special case.

``` js
var Module = {canvas: document.getElementById('canvas'), arguments:['cart.tic --vsync 0']}
```
https://github.com/nesbox/TIC-80/blob/main/build/html/export.html#L35

nesbox:
I disabled vsync by default for html export 6718e82ed0b2fdafca5d853475c2fc1795f54a37

nesbox:
... and let's make TIC workable on a high refresh rate monitor in the next version.

joshgoebel:
I might be interested in this, where can I find `tic80.js`?

Gate88:
Nice, thanks for the help and quick response @nesbox!

Gate88:
OK -- I pulled latest release candidate and tried again.

I think it's working better, but frame timing in general seems to drift a little more in the new version than it did the previous one.

The first time I ran the game here, the music was slow and it never recovered. Subsequent runs it's running a little better, but if you alt+tab and do something else for a bit the game will get into the same state.
https://gate.itch.io/tic80-workspace?password=new-tic-80-build

[...]

PR #2354, opened 2023-11-10
fix high fps bug for >60Hz displays

borbware:
Fixes bug discussed in #1873 for non-emscripten builds (everything else except html). Tested on Windows 10 machine with a 160Hz display.

Note: This could introduce a lag when catching up with frames not drawn yet.

Thanks to @ronchaine for the code, I just verified the result & created a pull request. :D


borbware:
only tested on a single 160Hz laptop screen - waiting if @Alia-Traces can test with their multi-monitor setup.

nesbox (approved):

Issue #2335, opened 2023-10-23
X scrolling in code editor is reversed


TimotheeGreg:
Hi,

When I x-scroll to the left with my mouse the code disappears on the left, as if the "camera" was moving to the right. And vice versa. That's the reverse of other software I am used to.

Note that when I use shift + y-scroll that goes how I would expect: scrolling downward moves the "camera" to the right.

That's minor.

TimotheeGreg:
Yes, you got it right, sorry if that was not clear ^^
Indeed a mouse can have two scroll wheels, or the other option (more common I think) is that you can push the scroll wheel to the right or the left to scroll horizontally.

TimotheeGreg:
You can also scroll horizontally with trackpads. That make me think:
It's reversed on the trackpad too on TIC-80 compared to other software, even if it does not feel as bad as a with the mouse wheel.
When I move my fingers to the right for an horizontal scrolling on TIC-80, the text moves to the left, while this time in other software it's the contrary. I think we want to have the same movement than for a touch-screen, your fingers push the text to the direction you go.

sogaiu:
Thanks for the clarifications.

I had forgotten about being able to push / tilt some scroll wheels to the left / right.  Wish I had that on my trackball :)

Ah, trackpads...have been avoiding using them as much as possible for years.  Forgot about being able to scroll with them (^^;  Good point.

Scrolling behavior being consistent with other typical applications does sound better.  Seems like it would feel less surprising.

No clue what this would involve.  May be it's straight-forward?

mickeyordog:
Any fix for this yet?

nesbox:
I added the **REVERT_SCROLL** option to the config here https://github.com/nesbox/TIC-80/pull/2355

Wang-Yue:
Not same but closely related bug --- when moving horizontally in the code editor using trackpad on macOS, the code moves to the opposite direction (compared to all the other native apps).

PR #2355, opened 2023-11-11
#2335: added REVERT_SCROLL option to the config

nesbox:


TimotheeGreg:
Great! Thanks a lot!
Ideally the default value should be the contrary of what existed up to now, imo, is that the case?

Issue #2360, opened 2023-11-19
Issues in calling `pmem` from wasm


peterhellberg:
I'm trying to write some code in Zig that depends on `pmem`, and were having some unexpected results, and then I looked in `src/api/wasm.c` and I believe there might be a bug in how the `bool` parameter to `tic_api_pmem` is assigned. _(I very rarely touch any C code so I'm a bit rusty in this regard)_

Does this mean that `writeToStorage` can never be `true`?
https://github.com/nesbox/TIC-80/blob/5061bb221fb8cd2e116efbe14295d7e3af198118/src/api/wasm.c#L579-L586

Doing something like this allows me to solve my issue, but would change the signature of `wasmtic_pmem`

```diff
diff --git a/src/api/wasm.c b/src/api/wasm.c
index 5dfd20f3..111b0655 100644
--- a/src/api/wasm.c
+++ b/src/api/wasm.c
@@ -575,16 +575,11 @@ m3ApiRawFunction(wasmtic_pmem)
     m3ApiReturnType  (uint32_t)
 
     m3ApiGetArg      (int32_t, address)
-    m3ApiGetArg      (int64_t, value)
-    bool writeToStorage;
+    m3ApiGetArg      (uint32_t, value)
+    m3ApiGetArg      (bool, writeToStorage)
 
     tic_mem* tic = (tic_mem*)getWasmCore(runtime);
 
-    // read value
-    if (value == -1) { 
-        writeToStorage = false;
-    };
-
     // TODO: this should move into tic_api_pmem, should it not?
     if (address >= TIC_PERSISTENT_SIZE) {
         m3ApiReturn(0);
@@ -987,7 +982,7 @@ M3Result linkTicAPI(IM3Module module)
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "peek4",   "i(i)",          &wasmtic_peek4)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "peek2",   "i(i)",          &wasmtic_peek2)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "peek1",   "i(i)",          &wasmtic_peek1)));
-    _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "pmem",    "i(ii)",         &wasmtic_pmem)));
+    _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "pmem",    "i(iii)",        &wasmtic_pmem)));
[...]

PR #2361, opened 2023-11-19
api/wasm: Make pmem `writeToStorage` an argument

peterhellberg:
Related to #2360 

> [!Important]
> This changes the signature of `wasmtic_pmem` so might not be acceptable?


joshgoebel:
This would be a breaking change.  If we wanted to preserve the existing behavior you could use one of the high bits of the 64bit value as flags (and then mask it back out for the memory access) - or we could just add another API function entirely.

peterhellberg:
@joshgoebel Ok, thank you for your input. I'll close this PR.

PR #2362, opened 2023-11-19
api/wasm: `writeToStorage` defaults to `true`, update `wasmtic_pmem` signature from `i(ii)` to `i(iI)`

peterhellberg:
I noticed two issues in `src/api/wasm.c` when trying to call `pmem` from a small [Zig](https://ziglang.org/) :zap: program (compiling to `.wasm`)

- `writeToStorage` was never set to `true` _(unless I've misunderstood the code that is)_
- The signature of `wasmtic_pmem` indicated that the arguments would be three 32 bit values: `i(ii)`, and not two 32 bit (return value and `address`) + one 64 bit value (for `value`) as I expected: `i(iI)`

Related to issue #2360

### DIFF
--- a/src/api/wasm.c
+++ b/src/api/wasm.c
@@ -576,7 +576,7 @@ m3ApiRawFunction(wasmtic_pmem)
 
     m3ApiGetArg      (int32_t, address)
     m3ApiGetArg      (int64_t, value)
-    bool writeToStorage;
+    bool writeToStorage = true;
 
     tic_mem* tic = (tic_mem*)getWasmCore(runtime);
 
@@ -987,7 +987,7 @@ M3Result linkTicAPI(IM3Module module)
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "peek4",   "i(i)",          &wasmtic_peek4)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "peek2",   "i(i)",          &wasmtic_peek2)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "peek1",   "i(i)",          &wasmtic_peek1)));
-    _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "pmem",    "i(ii)",         &wasmtic_pmem)));
+    _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "pmem",    "i(iI)",         &wasmtic_pmem)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "poke",    "v(iii)",        &wasmtic_poke)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "poke4",   "v(ii)",         &wasmtic_poke4)));
     _   (SuppressLookupFailure (m3_LinkRawFunction (module, "env", "poke2",   "v(ii)",         &wasmtic_poke2)));


joshgoebel:
Wait, why is value a 64 bit number - pmem stores and retrieves 32 bit numbers...  it seems like the C code is correct and the Zig API needs to change, no?

peterhellberg:
Because a i32 can not represent all u32 numbers + -1, I would personally be fine to have https://github.com/nesbox/TIC-80/blob/5061bb221fb8cd2e116efbe14295d7e3af198118/src/api/wasm.c#L578 be an `int32_t`, but I suspect others might need to be able to represent entire uint32_t values.

peterhellberg:
@joshgoebel We can probably disregard that I used `Zig` when building a `.wasm`, the same problem can be reproduced using `C`, as shown above.

```C
#include "tic80.h"

WASM_IMPORT("snprintf")
int snprintf(char* s, size_t n, const char* format, ...);

WASM_EXPORT("BOOT")
void BOOT() {
    pmem(1, pmem(1,-1)+1);
}

WASM_EXPORT("TIC")
void TIC() {
    cls(13);

    char buf[20];
    snprintf(buf, 20, "Started %d times", pmem(1,-1));
    print(buf, 3, 3, 15, 0, 1, 1);
}
```

joshgoebel:
_The goal here is to always maintain backwards compatibility._ If someone is using the existing API (bugs and all), that should continue to work.  I think the behavior here should be documented and one of two things happen:

- Introduce a new function with the signature in the name: `pmem_i_iI` (or some such)... libraries can be modified to link against this without breaking backwards compatibility with `pmem`
- client libraries could also choose to reimplement `pmem` using `peek` and `poke`.

peterhellberg:
@joshgoebel This wouldn't break any compatibility though? (since the function is currently broken i TIC-80)

peterhellberg:
Looked through the repo history and these two bugs have been present since https://github.com/nesbox/TIC-80/commit/aeabb55bef63a11014107d2d0a369ac7ec324530

joshgoebel:
> since the function is currently broken in TIC-80, not sure that it has ever worked)

A valid signature itself can't be "broken". It might be buggy but if anyone has ever successfully linked against it, then changing it breaks their code.  

You're saying our shipping Zig library breaks out of the box, yes?
We'd need to confirm that every other library (every WASM target) is also broken in the same way. 

Even that would be a compromise since it's quite possible to write WASM by hand.  And also possible that individual users have patched their libraries to work with the existing buggy API - and just never reported back here.

joshgoebel:
Correct me if I'm wrong but if linked correct the existing binary API works, just limits you to 32 bit signed numbers other than -1.

peterhellberg:
Disregard Zig for now, it is broken from C as well (as I noted above)

peterhellberg:
> A valid signature itself can't be "broken". It might be buggy but if anyone has ever successfully linked against it, then changing it breaks their code.

I believe my point is that it isn't valid. But maybe I've missed something fundamental in why this isn't working.

peterhellberg:
Another way to "fix" the problem of not being able to call the exported `pmem` function from wasm in TIC-80 without changing the signature would be to reduce the size of the value argument _(but I would think that would be a worse solution than to correct the signature to match the arguments to `wasmtic_pmem`)_
 
```diff
diff --git a/src/api/wasm.c b/src/api/wasm.c
index 5dfd20f3..f6142536 100644
--- a/src/api/wasm.c
+++ b/src/api/wasm.c
@@ -575,8 +575,8 @@ m3ApiRawFunction(wasmtic_pmem)
     m3ApiReturnType  (uint32_t)
 
     m3ApiGetArg      (int32_t, address)
-    m3ApiGetArg      (int64_t, value)
-    bool writeToStorage;
+    m3ApiGetArg      (int32_t, value)
+    bool writeToStorage = true;
 
     tic_mem* tic = (tic_mem*)getWasmCore(runtime);
```

peterhellberg:
### The same issue can be seen with code written in `D`:
Screenshot from 2023-11-20 18-46-43



```d
import tic80;

extern(C):

int  snprintf(scope char* s, size_t n, scope const char* format, scope const ...);

void BOOT() {
  pmem(1,cast(int)pmem(1,-1)+1);
}

void TIC() {
  cls(13);

  char[20] buf;
  char* bufptr = cast(char*)buf;
  snprintf(bufptr, 20, "Started %d times", pmem(1,-1));
  print(bufptr, 3, 3, 15, 0, 1, 1);
}
```

With the fix in this PR, and this change to `src/tic80.d`: `uint pmem(int index, long value);`

nesbox (approved):

Issue #2364, opened 2023-11-21
Thin text cursor


bonjorno7:
The 1-character thick text cursor is fine, but I still occasionally get confused by it, because I'm used to the modern thin line cursor.

TimotheeGreg:
You can customize this in the configuration file, look at https://github.com/nesbox/TIC-80/wiki/System-Configuration-Cartridge#sprites-and-sfx

bonjorno7:
Which sprite or option specifically?
I can't seem to find it.

TimotheeGreg:
Oh, ok, I misunderstood!

bonjorno7:
Found the implementation in drawCursor in code.c and it's easy enough to change.
Still need to figure out how to add an option though, since I figure not everyone will want this.

TimotheeGreg:
Maybe that could inspire you for adding an option: https://github.com/nesbox/TIC-80/pull/2355 :shrug:

bonjorno7:
That looks like exactly what I need, thanks!

PR #2373, opened 2023-11-21
Alt caret

bonjorno7:
The alt caret is one or two pixels wide depending on the font, and does not draw a gray copy of the character underneath.
It will be more familiar to people who are used to modern text editors.
I went with "caret" instead of "cursor" to avoid confusion with the mouse pointer, but can change this if you want.

tic80_alt_caret

Fixes: #2364

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -14,6 +14,7 @@ THEME=
 		CURSOR =2,
 		SHADOW =true,
 		ALT_FONT=false,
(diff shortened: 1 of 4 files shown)


bonjorno7:
Building updates `build/assets/config.tic.dat`, but I'm not sure whether I should commit that.

nesbox (approved):

Issue #2370, opened 2023-11-21
Trim excess whitespace


bonjorno7:
I'd be nice to have either a button, or option to do it on save, to remove excess whitespace from the end of lines, and excess empty lines from the end of the file.
I think it makes sense especially because whitespace counts against the size limit.

PR #2377, opened 2023-11-25
Trim whitespace on save

bonjorno7:
When you hit Ctrl + S with this option enabled, it will trim spaces from the end of lines, and newlines from the end of the file.
I think this not only looks nice, but actually matters, because spaces count against the file size limit.

It adds a trailing newline at the end of the file if there isn't one.
It respects the file size limit (64kb or 512kb for the pro version).
It updates the cursor position and selection appropriately.

Note: if your selection was entirely trimmed, you're left with a zero length selection.
I decided not to fix this, as the user may have created such a selection intentionally.
Shouldn't be difficult though, if you'd like me to change it.

Another note: trimming only happens with Ctrl + S, not with typing save in the console.
I could try fixing this if you want, but decided not to do it yet, because I feel it's better to only trim if you can see it happen, which you can't from the console (might make sense to only run this when you're actually in the code editor then, but haven't looked into that yet).
Another reason I didn't fix it is because it was convenient being able to save without running the trimming code while I was still developing it.

Unlike my last PR, this time I've included changes to `config.tic.dat`, because I now know what it's for.

Fixes: #2370


bonjorno7:
Like my other PR, building makes changes to `build/assets/config.tic.dat`.
Since that one got merged without those changes, I'm not committing them here either.

bonjorno7:
I'm hoping that updating `config.tic.dat` will fix the build errors.

bonjorno7:
I'm not sure why some of the builds failed.

bonjorno7:
I replaced it with a ternary, now it should be fine.
Used a force push so these back and forth changes don't need to live in the repo history forever.

bonjorno7:
It might make sense to create an undo step, though I'm not sure how to go about this.

bonjorno7:
It calls `history` before calling `update`; hopefully those are the right functions in the right order.
I did find one new issue before this (which was unfortunately not fixed by the last change): if you remove text at the end of the file by selecting it with your mouse and hitting backspace/delete, then save, it will bring back the deleted text from the dead, but only if a trailing newline was added (if there was already a trailing newline, the issue doesn't happen).
I'm not sure what exactly is going on with that, but it's worth noting I guess.

bonjorno7:
Okay I know what's going on.
I was under the assumption that I could rely on the rest of the file being nulled, but deleting your selection at the end of the file just sets one null and leaves the garbage data.
When my code adds a trailing newline, it can overwrite that null, meaning it's then followed by previously deleted characters until the next null.
Since apparently the file doesn't need to be zero'd to be valid, I'll just change the code to zero only one byte, and leave any garbage data after it alone.

bonjorno7:
There was even more to it: the loop didn't exit because the null check didn't happen, because the null was overwritten by a newline.
Now it stores whether we found a null at the start of an iteration, and checks it at the end of that iteration.
This way it actually exits when encountering a terminator, even if there's garbage data after it.

nesbox (approved):

Issue #1875, opened 2022-04-09
memcpy glitches out in the web build


verysoftwares:
i'm using memcpy to move screen memory around. see the glitches here, choose 2-4 players and move around: http://tic80.com/play?cart=2740

![image](https://user-images.githubusercontent.com/54413321/162589926-ad17ebea-5c60-454a-86ca-76acc4278b7a.png)

a downloaded cartridge works just fine.

nesbox:
Just reproduced it on the latest html build and I'm wondering what could be causing it, maybe compiler optimizations?

nesbox:
No, I see the same thing in TIC-80 for desktop, are you sure your code is working correctly?

verysoftwares:
i can't reproduce on desktop (0.90.1723 Pro)

nesbox:
I see glitches on the latest dev build
![image](https://user-images.githubusercontent.com/1101448/162694564-3d012878-98bb-484e-b705-7e559b61cf10.png)

mbikovitsky:
AFAICT, there are two related issues here:

1. The LUALA code uses `memcpy` to copy potentially overlapping buffers.
2. The TIC-80 `memcpy` implementation [uses the C `memcpy` function](https://github.com/nesbox/TIC-80/blob/e1dbcd001c7333b9b196ab9c04dffabc963f5c08/src/core/core.c#L129) to do the copy.

It is undefined behaviour to use `memcpy` on overlapping buffers: <https://en.cppreference.com/w/c/string/byte/memcpy>. But, [the documentation for the TIC-80 `memcpy`](https://github.com/nesbox/TIC-80/wiki/memcpy) doesn't specify this as a constraint, so users have no way to know of this pitfall.

The immediate workaround for @verysoftwares is to copy memory manually. Something like:

```lua
function memmove(dst, src, length)
    if dst <= src then
        for i=0,length-1 do
            poke(dst+i, peek(src+i))
        end
    else
        for i=length-1,0,-1 do
            poke(dst+i, peek(src+i))
        end
    end
end
```

It's slower, but in my testing the bug is no longer present (AFAICT).

@nesbox, I think the correct fix is to use [`memmove`](https://en.cppreference.com/w/c/string/byte/memmove) in `tic_api_memcpy`, since that's the behaviour of the TIC-80 `memcpy` users expect.

PR #2417, opened 2023-12-24
Use memmove in tic_api_memcpy

mbikovitsky:
Calling `memcpy` with overlapping source and destination buffers is undefined behaviour. For example, see #1875. But, users of the _TIC-80 `memcpy` function_ probably expect it to work fine even with overlapping buffers. Thus: `memmove`.


joshgoebel (approved):
https://stackoverflow.com/a/53486461

Looks like a positive change to me.

nesbox (approved):

Issue #2257, opened 2023-08-08
Make all the languages work as shared plugins


nesbox:
There will be TIC-80 binary with only Lua support (and other lua based langs like MoonScript)
all the other languages will work dynamically through shared libs.

joshgoebel:
Interesting. :)

Wang-Yue:
I think this can be marked as done.

nesbox:
I think not yet, I meant to move the languages into dynamic libraries that would be automatically loaded at the startup

PR #2418, opened 2023-12-25
Make WREN a configurable build option.

Wang-Yue:
Currently TIC-80 supports so many programming languages. While this gives game developers a lot of flexibilities, it has a few shortcomings not limited to larger binary size and longer build time. Most of the languages are not actively used. Some of those languages' developers are not actively developing / maintaining their languages anymore. TIC-80 is also not supporting all languages equally so features are missing here and there. From security perspective bundling in all those languages is also not ideal. 

I decide to make most languages optional so user can configure the build process. Not sure if you agree on this proposal so for the first PR I am only doing it for a single language. If this PR gets accepted I can continue to work on other languages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,23 +229,31 @@ target_include_directories(lpeg PRIVATE ${LUA_DIR})
 # WREN
 ################################
 
-set(WREN_DIR ${THIRDPARTY_DIR}/wren/src)
-set(WREN_SRC
-    ${WREN_DIR}/optional/wren_opt_meta.c
-    ${WREN_DIR}/optional/wren_opt_random.c
-    ${WREN_DIR}/vm/wren_compiler.c
-    ${WREN_DIR}/vm/wren_core.c
-    ${WREN_DIR}/vm/wren_debug.c
-    ${WREN_DIR}/vm/wren_primitive.c
-    ${WREN_DIR}/vm/wren_utils.c
-    ${WREN_DIR}/vm/wren_value.c
-    ${WREN_DIR}/vm/wren_vm.c
-)
+set(BUILD_WITH_WREN_DEFAULT TRUE)
 
-add_library(wren STATIC ${WREN_SRC})
-target_include_directories(wren PUBLIC ${THIRDPARTY_DIR}/wren/src/include)
-target_include_directories(wren PRIVATE ${THIRDPARTY_DIR}/wren/src/optional)
-target_include_directories(wren PRIVATE ${THIRDPARTY_DIR}/wren/src/vm)
+option(BUILD_WITH_WREN "WREN Enabled" ${BUILD_WITH_WREN_DEFAULT})
+message("BUILD_WITH_WREN: ${BUILD_WITH_WREN}")
+
+if(BUILD_WITH_WREN)
+
+    set(WREN_DIR ${THIRDPARTY_DIR}/wren/src)
+    set(WREN_SRC
+        ${WREN_DIR}/optional/wren_opt_meta.c
+        ${WREN_DIR}/optional/wren_opt_random.c
+        ${WREN_DIR}/vm/wren_compiler.c
+        ${WREN_DIR}/vm/wren_core.c
+        ${WREN_DIR}/vm/wren_debug.c
+        ${WREN_DIR}/vm/wren_primitive.c
+        ${WREN_DIR}/vm/wren_utils.c
+        ${WREN_DIR}/vm/wren_value.c
+        ${WREN_DIR}/vm/wren_vm.c
+    )
+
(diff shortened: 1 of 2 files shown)


include/tic80_config.h
```diff
@@ -36,7 +35,6 @@
 #   define TIC_BUILD_WITH_MOON     1
 #   define TIC_BUILD_WITH_FENNEL   1
 #   define TIC_BUILD_WITH_JS       1
-#   define TIC_BUILD_WITH_WREN     1
```
joshgoebel, line 39: Please do not change the defaults.  By default we build with all languges.

joshgoebel:
> Most of the languages are not actively used. 

Taken to the logical conclusion that argument would boil down to "only include 1-3 languages", which I don't think is a direction we prefer _by default_ - that ship sailed long ago.  Anyone who wants is free to build a smaller build (and all for the compiled defines to allow them to do so).  I think this may matter more for embedded uses... I don't find the full build at all "large" on an actual PC.

> Some of those languages' developers are not actively developing / maintaining their languages anymore. 

The latest version of Wren is still 0.4.0, and that's what we bundle - so I hope this isn't targeted at Wren... If there are any bugs we should be creating issues so they can be looked into, not removing languages from the default build.  Whether a language might *ever* become too old, too unmaintained that it needs to be removed is probably a larger discussion.

> TIC-80 is also not supporting all languages equally so features are missing here and there.

This is certainly true though - and one issue with supporting so many languages.  It'd be great if we had a language champion for each language that knew both that language and C really well to keep things maintained... but again if something is missing, lets make sure we have an issue for it so at least we know where the gaps are.

TimotheeGreg:
Related to issue #2257

Wang-Yue:
>  I don't think is a direction we prefer by default 

The PR doesn't change the default --- all languages are still built by default.

> Anyone who wants is free to build a smaller build (and all for the compiled defines to allow them to do so).

This PR gives them the option to build a smaller build. Without which user are not able to do so easily.

In current HEAD there are two languages already being done like so, (Janet and Ruby). I don't understand why other languages are not following these rules. The inconsistent behavior should be fixed. 

> If there are any bugs we should be creating issues so they can be looked into, not removing languages from the default build. 

Again you are not losing anything in the default build at all. 

>The latest version of Wren is still 0.4.0, and that's what we bundle - so I hope this isn't targeted at Wren...

Wren is just an example and certainly I am not targeting it. I made it clear  in the PR introduction that this is just a start, Ideally all languages should be opt in. If this is accepted I can make it happen for other languages too. 

I am not saying you are not using the latest version. What I refer to is upstream. Some of the languages are not actively developed by authors for various reasons (people lost interest, want to spend more time with family, or simply because of the war going on, refer to wasm3 for example). There may be security bugs or crashes that are not guaranteed to be actively fixed.

include/tic80_config.h
Wang-Yue: I now understand why you misunderstood the purpose of this PR. 
THings defined in tic80_config are not the default, but brutal forced value (imho). 
Default should and is defined in CMakeList.txt. see line 232 of the changed file. 
`set(BUILD_WITH_WREN_DEFAULT TRUE)`

If user don't intentionally disable it, it builds the same binary as current HEAD.

nesbox: I agree, pls don't change these definitions. Alternatively, you can move all of them to the cmake.

nesbox (changes requested):

include/tic80_config.h
Wang-Yue: @nesbox what do you mean by it? I did move the TIC_BUILD_WITH_WREN to Cmake and it is now controlled by  BUILD_WITH_WREN_DEFAULT whose default value is true. This is align with Janet and ruby implementation.

The PR only move one (wren) to cmake. I do have plan to move all of them but before that I need to see if you guys agree with this approach or not, thus I sent out this change for feedback.

Wang-Yue:
@nesbox  I replied your comment and please take a look at it with more patience. This PR does everything following what Janet and Ruby is already doing. If you want TIC_BUILD_WITH_WREN to be configurable in cmake, it should not be in include/tic80_config.h (ruby and janet are not defined there). Define the wren macro there (which force it to be true no matter what definition is given from CMake) will break the build as well if user do not opt into it.

CMakeLists.txt
```diff
@@ -628,6 +643,10 @@ if(BUILD_STUB)
 
 endif()
 
+if(BUILD_WITH_WREN)
```
Wang-Yue, line 646: @nesbox take a look at here. the TIC_BUILD_WITH_WREN is not removed, but moved here, exactly following what you said in the other comment.. You also see mruby and janet flags defined in the same way below.

nesbox (approved):

PR #2420, opened 2023-12-26
add `--target=wasm32-unknown-unknown` to README.md

phocks:
Without `--target=wasm32-unknown-unknown` there is no `wasm32-unknown-unknown` folder and we get a file not found.


phocks:
I notice this game has this toolchain file which specifies it https://github.com/Dzejkop/super-space-smugglers-jam/blob/main/rust-toolchain.toml

nesbox (approved):

PR #2435, opened 2024-01-08
Fix load behavior around the website

### LINKED ISSUE #2433: load command behaves oddly around the website
The implementation of the website integration in TIC-80 creates the illusion of a folder structure; however, the `load` command behaves oddly around this boundary.

**Steps to reproduce**
1. Create a folder called `foo`. (The name is irrelevant.)
2. `cd foo`
3. Save a cart called `bar.tic` (Again, name is irrelevant and only used for instructions.)
4. `cd ..`
5. `load foo/bar.tic`
6. Observe how the file is loaded.
7. Navigate into `tic80.com` and locate a cart. Remember the path of this cart.
8. `cd` back to the root directory
9. Attempt to `load` the cart using the path

**Expected behavior**
The cart loads.

**Actual behavior (as of 1.1.2837)**
The load errors with `file not found`.

MineRobber9000:
Fixes #2433 (or at least, it will when I'm done with it).

- [x] Allow loading a file from the base `tic80.com` directory from the root as if it were normal
- [ ] Allow loading a file from a subdirectory of the base `tic80.com` directory from the root as if it were normal

### DIFF
--- a/src/studio/fs.c
+++ b/src/studio/fs.c
@@ -541,6 +541,46 @@ void tic_fs_changedir(tic_fs* fs, const char* dir)
 
 typedef struct
 {
+    char *dir;
+    char *file;
+} tic_fs_splitdir_result;
+
+#if !defined(__TIC_WINDOWS__)
+#include <libgen.h>
+#endif
+
+tic_fs_splitdir_result* tic_fs_splitdir(tic_fs *fs, const char *path)
+{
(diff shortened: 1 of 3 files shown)


MineRobber9000:
With college starting back up, I don't have time to work on this anymore. Besides, I never really knew what I was doing with regards to splitdir and all that. :(

PR #2453, opened 2024-02-07
add remap function

TheDrawingCoder-Gamer:
Changes:
- update wasm3 to latest commit
- changes the remap argument in wasm's map to a pointer holding a struct with data
- this data is used to call the remap function
- signature is the same so it's compatible


TheDrawingCoder-Gamer:
can't figure out how to call WASM functions that return a value. if you can do that you could remove the res_ptr arg. the user_data arg could also possibly be removed if we made a holder struct that held a pointer and function pointer

nesbox (approved):

nesbox:
@TheDrawingCoder-Gamer we have conflicts in the PR, could you pls resove?

TheDrawingCoder-Gamer:
@nesbox done!

TheDrawingCoder-Gamer:
@nesbox there is no longer a diff between pocketpy on this branch and main + it builds, so i think it's ready now

PR #2505, opened 2024-04-14
Byte Battle / FFT support

aliceisjustplaying:
Hi,

I've taken the prior patches and ported them to the latest version. Would there be interest in upstreaming and merging this (after making appropriate adjustments)?

This adds support for Byte Battles and FFT. It also sets devmode to on by default, but that should be a CLI argument, and it also replaces the default cart to a more demoscene-ey one.

This PR is mostly to gauge interest, I can split things up to separate smaller ones later.


Skeptim:
I'll let more competent people answer (nesbox has time to work on TIC-80 mostly on the weekends so you might need to wait for next weekend).
But that looks cool!
Having the dev mode as a CLI option is a good idea, but I personally don't like it (Esc+F1 is easy to do) so I don't think it should be 'on' by default.
Also I don't know what you want the default cart to be but it should be the same in every language and understandable by people who know little about coding. TIC-80 is also used to learn how to code.

aliceisjustplaying:
@nesbox Great! I'll chop up the changes into three (or more) separate pull requests:

1. the FFT support
2. ByteBattles mode support, which I agree should be hidden by default. It's already hidden behind CLI options, and I think that's a sensible setup? I don't think it's worth adding to the options screen itself
3. a CLI parameter that sets Developer Mode on by default instead of off

Already got started on the first :)

PR #2528, opened 2024-04-23
Compile for MacOS arm64 in CI as well

aliceisjustplaying:
GitHub has runners now for macOS 14 / arm64 so we can have native builds from the CI.


aliceisjustplaying:
Of course the new macos build fails when I create the PR but the error log shows that it's some issue with the runner.

nesbox (approved):
thank you

Issue #2448, opened 2024-01-29
Crash when hitting 'tab' after typing 'load'


AMcBain:
Version 1.1.2837 Pro (be42d6f)

**Steps to reproduce:**
1. Launch TIC-80
2. Type 'load '
3. Hit tab

**What happens?**
The TIC-80 window closes abruptly.

**What should happen?**
Not crash.

If there's more characters typed before hitting 'tab' then TIC-80 lists all matching ones, but when there's no characters to match on I guess just don't list anything?

nesbox:
can't reproduce it on windows binaries, what os do you use?

AMcBain:
> can't reproduce it on windows binaries, what os do you use?

`Windows 10 Pro, 22H2 build 19045.4046`

I launch it by double-click in Explorer, nothing special.

The space after is important, as if you type 'load' (no space) and hit tab, it adds a space. Then hitting tab again causes a quick hang then crash.

aliceisjustplaying:
i've spent the last god knows how many hours debugging this because it was driving me crazy and i'm 90% sure it only happens when a folder has a lot of files, enough that it fills the screen and then some. happens with load and save. i have a branch with a bazillion print statements on top of debugging with lldb, so far no luck, but my search continues.

aliceisjustplaying:
it's some use after free/heap corruption bug. lldb backtraces vary, it's something about tab completion but sometimes the backtrace is about GPU stuff which is weird. sometimes it isn't.

aliceisjustplaying:
okay, i've found the bug, and started working on a fix, though it's a bit complicated. I've written it up here:
https://github.com/aliceisjustplaying/TIC-80/pull/10

aliceisjustplaying:
Bugfix in https://github.com/nesbox/TIC-80/pull/2533.

PR #2533, opened 2024-04-26
Fix tab-completion crash with load/save/etc.

aliceisjustplaying:
# Issue

When using tab completion with `load`/`save` in a folder with a lot of files, or really anything that'd result in an output longer than `CONSOLE_BUFFER_SCREEN`, TIC-80 would crash. This is due to `data->options` of TabCompleteData overflowing and not being null-terminated.

## Minimal repro:

Create 100 blank `.tic` files:
```
#!/bin/bash
mkdir repro && cd repro || exit 1
for i in {001..100}; do
  touch "test-$i.tic"
done
```

Start TIC-80 with the `repro` folder as its base: `./build/bin/tic80 --fs ~/repro`

Type "load<tab>" and it'll likely crash. maybe not at first, but after a few tries at most, definitely.

## Notes

This was a hell of a bug to track down, but also a fun one that made me learn `lldb` and `ASAN`. I have some other PRs coming soon about that.

## Future improvements

I have a few suggestions for future improvements and PRs:
* Upping `CONSOLE_BUFFER_SCREEN` (=760) to something higher, so more can be displayed, and suggestions will be cut off less at `load`/`save`
* Having the suggestions sorted alphabetically, which they are currently not

## String handling

This is a larger discussion, but it's worth considering either:
* Migrating string operations in the codebase to [`strlcat`](https://linux.die.net/man/3/strlcat) which is safer and guarantees null-termination, or
* Adopting a real string library, such as [`cstr`](https://github.com/stclib/STC/blob/master/docs/cstr_api.md) from [`STC`](https://github.com/stclib/STC) or [`SDS`](https://github.com/antirez/sds)

Curious of your thoughts about the potential improvements I've listed in the two sections above.

----

Fixes https://github.com/nesbox/TIC-80/issues/2448.

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -1186,6 +1186,9 @@ static void addTabCompleteOption(TabCompleteData* data, const char* option)
         {
             // This is the first option to be added. Initialize the prefix.
             strncpy(data->commonPrefix, option, CONSOLE_BUFFER_SCREEN);
+            // strncpy does not null-terminate if the source string is too long so
+            // null terminate in the end just to be sure
+            data->commonPrefix[CONSOLE_BUFFER_SCREEN - 1] = 0;
         }
         else
         {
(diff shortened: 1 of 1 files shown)


nesbox:
I've created a separate issue for your suggested future improvements https://github.com/nesbox/TIC-80/issues/2540.
Regarding string handling, I'm sure `strlcat/strncat` is the easiest way to make string handling safer :)
Thank you 👍

Issue #1844, opened 2022-02-17
Locally compiled version won't start, claims a folder doesn't exist even though it does


jennybrown8:
I built tic80 locally.  It built fine, no errors.  When I try to run it, it gives the following error about a folder not existing.  I created the folder manually, then reran, and it still complains about it, even though the folder exists.  Seems to be a misleading error message when something else is the true cause, but I have no idea what.

    Windows 10 Pro 20H2, build 19042.1526
    Cloned from this repo on Feb 17 2022 from main. 
    Same error whether I build with or without pro flag.

    C:\Users\jenny\dev\demoart\TIC-80\build\bin>tic80
    error: folder `C:\Users\jenny\AppData\Roaming\com.nesbox.tic\TIC-80\` doesn't exist

    C:\Users\jenny\dev\demoart\TIC-80\build\bin>dir 
    C:\Users\jenny\AppData\Roaming\com.nesbox.tic\TIC-80\
    Volume in drive C is Windows
    Volume Serial Number is AC37-9F02

    Directory of C:\Users\jenny\AppData\Roaming\com.nesbox.tic\TIC-80

    Thu 2022-02-17  01:19 AM    <DIR>          .
    Thu 2022-02-17  01:19 AM    <DIR>          ..
               0 File(s)              0 bytes
               2 Dir(s)  574,803,382,272 bytes free

nesbox:
Hmm, I've never seen this before, does the 0.90 version work?

aliceisjustplaying:
I can repro this, strangely enough, on a Windows XP build (built on Windows 11) but executed on Windows XP. Works on Windows 7.

![Screenshot 2024-04-27 at 4 28 08 PM](https://github.com/nesbox/TIC-80/assets/81575558/01187995-241b-4a10-a70d-f715e1559a43)

aliceisjustplaying:
double-checked with the CI build; same issue.

aliceisjustplaying:
This is a bug in the runtime/toolchain and the tl;dr is that you need to use `GetFileAttributesEx` on Windows (or at least on Windows XP at least) instead of `stat`.

aliceisjustplaying:
Not sure why the error happens for the OP, though.

Issue #2357, opened 2023-11-11
TIC-80 does not start on Windows XP


SIRprise:
If I try to start the XP-version of TIC-80 on my german XP, I got:
"error: folder 'C:\Dokumente und Einstellungen\SIRprise\Anwendungsdaten\com.nesbox.tic\TIC-80\' doesn't exist"
The problem is: It actually does exist (it is empty)

The problem seems to be similar to #1495

nesbox:
As a temporary solution, you can run it with `tic80 --fs .` command

SIRprise:
Didn't work: "error: folder `.` doesn't exist.
It seems to be I can enter any value and it won't be found, even if I copy everything to C:\temp and use --fs C:\temp

aliceisjustplaying:
Fixed in #2558

PR #2558, opened 2024-05-02
Fix the error 'folder <folder> doesn't exist' under Windows XP (win32 build)

aliceisjustplaying:
[There is a bug](https://stackoverflow.com/questions/32452777/visual-c-2015-express-stat-not-working-on-windows-xp) in the toolchain we use for the Windows XP builds that makes stat not work and give false errors. The recommended solution is to use `GetFileAttributesEx` instead. The cleanest way I see for it is to create a shim that maps to stat and conditionally #include it.

Tested on a Windows XP VM, freshly compiled on Windows 11 with the `v141_xp` toolchain and works.

Fixes https://github.com/nesbox/TIC-80/issues/1844, https://github.com/nesbox/TIC-80/issues/2357, https://github.com/nesbox/TIC-80/issues/1924.

### DIFF
--- a/src/studio/fs.c
+++ b/src/studio/fs.c
@@ -158,6 +158,50 @@ static const char* stringToUtf8(const FsString* wstr)
     return str;
 }
 
+time_t FileTimeToTimeT(FILETIME* ft) {
+    ULARGE_INTEGER ull;
+    ull.LowPart = ft->dwLowDateTime;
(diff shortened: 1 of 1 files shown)


nesbox (approved):
I couldn't reproduce this bug, were you able to reproduce it?

aliceisjustplaying:
@nesbox yes repro'd on XP

PR #2567, opened 2024-05-05
Fix `surf` on win32 builds

aliceisjustplaying:
`surf` defaults to HTTPS however the TIC-80 website uses a modern certificate that's not available (by default at least) on old systems, thus we default to plain `http` on the 32-bit build.

Tested on a Thinkpad x220 running Windows XP (32-bit) / Windows 7 (32-bit).

Old build on the left, fix on the right.

surf-winxp

surf-win7-fix


nesbox (approved):
thank you

Issue #2561, opened 2024-05-03
The "X button to enter surf" needs a better input check


Weeppiko:
Currently, there is a strange behavior with it: When the game throws an error and you're still holding the X button on a gamepad, it warps you directly to Surf. It should check for the X button only when it is pressed, not if it is held.

Skeptim:
@LettucePie Could you have a look at that?
Sorry to add work ^^'

LettucePie:
Hello! Thank you for the report.

I'll start looking into more precise ways to distinguish the button input. If there already are flags designed for `pressed`, `released`, and `held` then it will be painlessly easy to fix.

I am busy for the next couple of days however, so it may be a bit until I get to it.

LettucePie:
Hello again, did a quick search into the button press command before going to work. It's located in the `io.c` file at 43 ( https://github.com/nesbox/TIC-80/blob/49111447e40e97eabd66d3b167770315f5cf693b/src/core/io.c#L43 )

I then searched for different ways that function is being used since I don't yet understand what it's doing. I found a bunch of results where along with the input command is passed a Hold and Period value. ( https://github.com/nesbox/TIC-80/blob/49111447e40e97eabd66d3b167770315f5cf693b/src/studio/screens/menu.c#L227 )

I went to do a quick build, but unfortunately my system isn't ready for the newest build instructions and my internet is very very slow, so my quick morning investigation ends here. I have no way of testing it right now, but in theory you can edit my line of `tic_api_btnp(tic, 6, -1, -1)` into `tic_api_btnp(tic, 6, 20, 3)` in `console.c` ( https://github.com/nesbox/TIC-80/blob/49111447e40e97eabd66d3b167770315f5cf693b/src/studio/screens/console.c#L4183 )

Skeptim:
I tried this but it did not work, the issue is still there.
You can find documentation on how this API should work on the wiki: https://github.com/nesbox/TIC-80/wiki/btnp but I would have expected `6, -1, -1` to be the best option.

LettucePie:
Making some progress, I successfully distinguished against a freshly pressed down button vs on being used or held. It didn't fix the issue though which I thought confusing. It was like the inputs weren't being dropped between changing states, and that was exactly it. `tic_api_reset` in `core.c` seems to handle dropping inputs when a game crashes, or a screenstate changes entirely. Turns out, gamepads were not yet included lmao.

[...]

PR #2571, opened 2024-05-07
Added functions for distinguishing gamepad button presses, also fix gamepad data persisting

LettucePie:
Hello again!

This fixes issue 2561 (https://github.com/nesbox/TIC-80/issues/2561) which is an issue I inadvertently created, familiar territory for most of us here I'm sure lol.

I am still not very experienced with C and memory management, and while fixing this I found some security messages and other questions within `core.c`. While it seems like I didn't violate any security issues, it might be best if someone who understands this better checks in. I separated my changes to `core.c` in two commits. My attempt to solve 2561 requires those changes.

- `817e09c3299306c9ce88ac7fa6ad1607dfe0367f` before touching core.c
- `c3328075aa3d2bd1697f362a7cc340f473d4615f` after

To easily test the issue, I made a new lua cart called "breaktest"

```
x = 1
function TIC()
	if btn(6) then
		trace("breaking") 
		x = x + y 
	end
	cls(0)
end
```

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -280,6 +280,33 @@ enum
         tic_mem*, s32 id, s32 hold, s32 period)                                                                         \
                                                                                                                         \
                                                                                                                         \
+    macro(btnd,                                                                                                         \
+        "btnd(id) -> just pressed",                                                                                     \
+                                                                                                                        \
+        "This function allows you to read the status of one of the buttons attached to TIC.\n"                          \
+        "The function returns true if the key with the supplied id was just pressed.\n"                                 \
+        "It only returns true for the single tick the event occurred in.\n"                                             \
+        1,                                                                                                              \
+        1,                                                                                                              \
+        0,                                                                                                              \
+        u32,                                                                                                            \
+        tic_mem*, s32 id)                                                                                               \
+                                                                                                                        \
+                                                                                                                        \
+                                                                                                                        \
+    macro(btnu,                                                                                                         \
+        "btnu(id) -> just released",                                                                                    \
+                                                                                                                        \
+        "This function allows you to read the status of one of the buttons attached to TIC.\n"                          \
+        "The function returns true if the key with the supplied id was just released.\n"                                \
+        "It only returns true for the single tick the event occurred in.\n"                                             \
+        1,                                                                                                              \
+        1,                                                                                                              \
+        0,                                                                                                              \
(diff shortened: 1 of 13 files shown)


LettucePie:
Some builds failed... Looks like I need to add my new functions to a header file I'm guessing? I can't seem to find where the other functions on `io.c` are hanging out. There isn't even an `io.h` file...

LettucePie:
_emerging from the trees and brush, covered in dirt and wounds while hyperventilating_ 

**Okay! So it turns out you don't just simply add functions to io and api!**

I was confused why some platforms weren't building properly and noticed the only place where `tic_api_btn` and `btnp` were addressed was the `api.h` file in a big weird list of other api functions...

Anyways, here's `btnd` and `btnu` working in all available langs except wasm. I just didn't test it because I don't entirely understand what it wants me to do.
[btnd-btnu-tests.zip](https://github.com/nesbox/TIC-80/files/15240254/btnd-btnu-tests.zip)

nesbox:
I have serious doubts that we should bloat the API for these functions when user can always read the current state of the gamepad buttons using `peek(0x0FF80)`

LettucePie:
It's ultimately up to you. I personally feel like the `btnu` call has a bit of value as being generally accessible, and doesn't require users to form variables to compare against.

If you read issue 2561 you'll find I originally made these functions to solve a bug where users who crash their game from pressing button 6 get warped to Surf. It was my changes to `core.c` that ultimately corrects this. `btnd` and `btnu` are sort of convenient byproducts. The original api call using `btnp` along with the changes to `core.c` should suffice (I haven't tested that scenario).

LettucePie:
@nesbox I tested my hypothesis from earlier now that I'm back home. The changes to `core.c` were all that are required for fixing issue 2561. I already setup the fix at Pull Request 2579 (https://github.com/nesbox/TIC-80/pull/2579).

I don't know if you plan on asking for more opinions on `btnd` and `btnu` so I'll leave this open. It's no disrespect to me if you chose to close it. Now that I have an idea of what it takes to expand and adjust the api though, I'm excited to check out the issues and todo for some challenges to tackle lol.

nesbox:
@LettucePie thank you for your help, I merged the pull request regarding the bug in the core, but as for the new API, I suggest closing this part, sorry.

PR #2592, opened 2024-05-20
libretro: Update the list of static language libraries

RobLoach:
I'm unsure if this is the correct approach, but I've updated the cmake definition for libretro to construct a list of language libraries to static link against. Found that on some platforms, since JS was disabled, it wasn't able to find libquickjs.a, so hopefully this fixes it.


nesbox (approved):

nesbox:
ahh, thank you for the support 😄

Issue #2292, opened 2023-09-02
Keyboard Arrows not working in the Android Version


NoctisLXV:
I downloaded the version 1.1.27.37 and noticed that in the android version, my bluetooth keyboard arrows doesn't work. For some reason, the editor recognizes all the other keys, but the arrow keys doesn't work (and I tested It in other apps and works normally). Even when testing a game, TIC-80 doesn't recognize the arrow keys, and only the arrow keys.

I use a Galaxy Tab S6 Lite, Android 13. In the last version of the App, it worked normally.

Thanks.

nathancmcrae:
Same issue, same hardware. The latest release that the arrow keys work on for me is 1.0.2164. Hope that helps.

PR #2593, opened 2024-05-22
Fix for Arrow keys not working on android.

Pedrocasf:
Update SDLActivity.java so that the arrow keys are processed first then the joystick.


Pedrocasf:
Should close  #2292.

nesbox (approved):

PR #2604, opened 2024-05-26
Don't run CI jobs twice on PRs

aliceisjustplaying:
Because of `on: [push, pull_request]`, CI jobs get triggered twice on PRs, this fixes that.

This is a simple fix; may be able to refactor the CI jobs into something nicer in the future that doesn't need it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ jobs:
   # === Windows XP ===
   winxp:
     runs-on: windows-2019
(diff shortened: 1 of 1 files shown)


aliceisjustplaying:
Hmm, maybe this doesn't happen anymore? It definitely happened a few weeks ago when I created this patch. Did Github fix something around this?

nesbox:
I'm unsure about this PR, if it doesn't reproduce I will close it for now.

PR #2607, opened 2024-05-26
Feature: FFT support

aliceisjustplaying:
Oh my god we finally got here.

A few comments:
* The feature is hidden behind the `--fft` flag. If it's not enabled or we're on unsupported platforms (3DS, baremetal PI, Android) the functions `fft()` and `ffts()` simply return 0 and nothing gets even initialized.
* Took the liberty to add the authors of the past Byte Battle / FFT patches to the contributors as well as the prior work in @TheNuSan's [Bonzomatic fork](https://github.com/TheNuSan/Bonzomatic). Let me know if you'd like any changes regarding that
* Had to squash most of the FFT feature into one commit because rebasing became nigh-impossible after some recent changes
* Passing around `fftdata.h` is kind of gnarly but it's the least bad solution I've found. If you have ideas for improvements they're extremely welcome
* This feature will definitely keep evolving, but this is a good moment to merge it to upstream and iterate from there


nesbox (approved):

nesbox:
@aliceisjustplaying  Awesome, thank you 👍

PR #2619, opened 2024-06-21
Add a check to draw.c:getPalette so it does not write out of bounds past the mapping static array into whatever static happens to come after: Fixes #2611 segfault

### LINKED ISSUE #2611: Segfault when saving after running
Version: 1.2.2948-dev Pro (d6ef398)

To reproduce, create a new cart with a name, and fill the code with the following:
```
function TIC()

	spr(256, 96, 24, -1)
	
end
```
Running this will show the first sprite in roughly the middle of the screen. If you then save out the cart, you'll achieve a segfault with the following stacktrace:
```
(gdb) bt
#0  0x00005555555bf361 in project_ext ()
#1  0x00005555555cd818 in saveCartName ()
#2  0x00005555555cd92d in saveCartName ()
#3  0x00005555555cd96c in saveCart ()
#4  0x00005555555bf648 in saveProject ()
#5  0x00005555555bffdb in processShortcuts ()
#6  0x00005555555c0913 in renderStudio ()
#7  0x00005555555c1754 in studio_tick ()
#8  0x00005555555b8a9c in gpuTick ()
#9  0x00005555555b917f in start ()
#10 0x00005555555b928c in main ()
```

I'm unsure as to what could even be causing this, as I've reduced it down to just that one single line in an blank cart!

OvermindDL1:
Fixes #2611 segfault

Please see that related issues for full details about this static memory corruption issue.

This is not a complete fix, but it at least no-ops the memory corruption by performing a bounds check on user supplied input.

### DIFF
--- a/src/core/draw.c
+++ b/src/core/draw.c
@@ -51,7 +51,12 @@ static u8* getPalette(tic_mem* tic, u8* colors, u8 count)
 {
     static u8 mapping[TIC_PALETTE_SIZE];
     for (s32 i = 0; i < TIC_PALETTE_SIZE; i++) mapping[i] = tic_tool_peek4(tic->ram->vram.mapping, i);
-    for (s32 i = 0; i < count; i++) mapping[colors[i]] = TRANSPARENT_COLOR;
+    for (s32 i = 0; i < count; i++) {
+        if (colors[i] < TIC_PALETTE_SIZE)
+        {
+            mapping[colors[i]] = TRANSPARENT_COLOR;
+        }
+    }
     return mapping;
 }
 


OvermindDL1:
As an aside, I'm not sure this `getPalette` function is correct at all, but I'm not sure how it is supposed to be structured to know better at the current time.  The lack of documentation as to its purpose of a `get*` function writing to memory is also especially confusing to me.

OvermindDL1:
For note, the macos-arm64 build failure appears unrelated, it seems the resources on that macos build runner were exhausted and it was unable to spawn more clang processes to compile, thus not related to this PR.

nesbox (approved):

Issue #2573, opened 2024-05-07
Command line not fully visible when using past commands


Skeptim:
If I use a command line that is longer than one line it will go to line so we can see it entirely but then if I press the Up arrow of the keyboard to use this same "too long" command line, we will see only the first line.
That's minor.

Here is an example, after using the command line once, I press Up arrow and I can't see the command line I am using entirely.

![video20](https://github.com/nesbox/TIC-80/assets/26139286/f70a13d2-620b-4e0a-8531-3b62df0f038b)

Miguel-hrvs:
That's true, also happens in the master branch, is pretty annoying. I'll see if I can do something.

Miguel-hrvs:
One fast fix would be to apply the cls command automatically while pressing the up arrow key to show history:

https://github.com/nesbox/TIC-80/assets/81558772/ea2d90e6-bb7d-44d0-a1b8-54a26a279b23

PR #2627, opened 2024-07-10
Avoid command from history not fully visible at the bottom

Miguel-hrvs:
Fixes: https://github.com/nesbox/TIC-80/issues/2573

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -4270,7 +4270,10 @@ static void processKeyboard(Console* console)
         }
         else
         {
-            if(keyWasPressed(console->studio, tic_key_up)) onHistoryUp(console);
+            if(keyWasPressed(console->studio, tic_key_up)) {
+			    onClsCommand(console);
+			    onHistoryUp(console);
+            }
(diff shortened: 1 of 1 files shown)


src/studio/screens/console.c
```diff
@@ -4270,7 +4270,10 @@ static void processKeyboard(Console* console)
         }
         else
         {
-            if(keyWasPressed(console->studio, tic_key_up)) onHistoryUp(console);
+            if(keyWasPressed(console->studio, tic_key_up)) {
+			    onClsCommand(console);
```
nesbox, line 4274: hmm, I don't understand why we need to clear console everytime we press UP button, 
I think we need to scroll screen to make text visible instead cleaning

nesbox (changes requested):

Miguel-hrvs:
It was a workaround until the proper scrolling is implemented.

src/studio/screens/console.c
Miguel-hrvs: In the end, the proper solution was very easy, I'm going to apply it, cheers :)

nesbox (approved):

Issue #1992, opened 2022-07-22
Bug: 256 sprite apears on ESC


BuoYancYdabl:
**What is wrong**
When tapping fast escape key the 256 sprite shows on the cursor for one frame!

**How to recreate**
1) Fill 256 sprite with anything noticable;
2) Start going back and forth to editor and console;
3) You'll notice sprite appearing on one frame below cursor.

**Additional information**
![GIF](https://user-images.githubusercontent.com/36867670/180575818-8be942df-c7fc-47ef-bb68-30ac180f4360.gif)
Platform: Windows 7, PC
Version of TIC: v1.0 Pro winxp
Is fullscreen: yes

BuoYancYdabl:
Little update: for some reason it is imposible to recreate, in an empty cart this isnt happening. But its happening exactly in my game. I tried restarting tic, using different modes.

BuoYancYdabl:
Please confirm if this also happens to you. I need to know is this my PC, ot the TIC-80. It might be visual artifact, because im specifically using wrong videodriver.

pythonmcpi:
Can reproduce using the Windows (regular) build from Github Actions Build #954. I believe my video drivers are up to date.

joshgoebel:
Perhaps another issue with the new cursor code?  Tagging as bug.

pyanmcbride:
> Perhaps another issue with the new cursor code? Tagging as bug.

I have noticed it before the new cursors.

BuoYancYdabl:
Ok, so somehow that bug moved from 258 to 259. This happens on a quarter of a second if you interfere in to cursor memory. Have this bug in the other project now.

BuoYancYdabl:
Well I have found some more weird stuff. For example, this bug affects the moving between editors or console while sprite in 257, 258 and 259 are appearing when game is lanching/closing. It have nothing to do with manual cursor changing, because I disabled that in my game to test it.

Miguel-hrvs:
If you set tic_point hot to something high, like tic_point hot = {100000}; the sprite doesn't show up anymore.

Issue #2583, opened 2024-05-11
[BUG(?)] Mouse sprite and ram aren't linked


SRB2er:
<div type='discussions-op-text'>

<sup>Originally posted [here]( https://github.com/nesbox/TIC-80/discussions/2577) until i just remembered issues exist. sorry. 

I found out a weird bug with the sprite the mouse uses
rather than always use the data in ram,  the mouse seems to only use the sprite in ram during startup, after which it can't be changed (not linked between each other)

here's some lua code to explain what I mean 
**(you'll need to put a sprite in spr 257):**

```
--mouse sprite "bug"
--put something in sprite 257 for this
--preferably something that isnt just 1 color
t=0
function TIC()
 cls()
 mx,my=mouse()
 --self explanatory
 print("sprite 257:",mx-64,my-10,12)
 --self explanatory
 print("mouse:",mx-64,my,12)
 --draws sprite 257 8 pixels under mouse
 spr(257,mx,my-10) 
 --should set mouse sprite to 257
 poke(0x3ffb,1) 
 -- writes a semi checkboard of 
 -- white and red to sprite 257
 -- after 5 seconds
 if t>300 then
 memset(0x6020,0xc2,32)
 end
 t=t+1
end
```

</div>

Miguel-hrvs:
Fixed by: https://github.com/nesbox/TIC-80/pull/2629

Miguel-hrvs:
Fixed by: https://github.com/nesbox/TIC-80/pull/2629

PR #2629, opened 2024-07-11
Fix mouse showing sprite 258 pressing esc

Miguel-hrvs:
Fixes: https://github.com/nesbox/TIC-80/issues/1992 and https://github.com/nesbox/TIC-80/issues/2583


src/studio/studio.c
```diff
@@ -2313,19 +2313,13 @@ static void blitCursor(Studio* studio)
         if(tic->ram->vram.vars.cursor.system)
         {
             bank = &getConfig(studio)->cart->bank0;
-            hot = (tic_point[])
```
nesbox, line 2316: Why are you deleting hardcoded "hot" points of the cursor here?

Miguel-hrvs: Didn't know why the hot code was responsible for duplicating the mouse sprites. Worked great without it and made it simpler. So I left it that way.

nesbox: We use hot point in the code editor for example for IBeam cursor

Miguel-hrvs: The IBeam cursor works even without the hot point.

nesbox: it works without hotpoint, but the click point should be the center of the I Beam cursor

Miguel-hrvs: I see, I'll look into that

Miguel-hrvs: Found the proper fix, ready to be merged

src/studio/studio.c
```diff
@@ -2334,20 +2334,16 @@ static void blitCursor(Studio* studio)
 
         tic_point hot = {0};
 
-        if(tic->ram->vram.vars.cursor.system)
+        bank = &getConfig(studio)->cart->bank0;
```
nesbox, line 2337: You removed using of `vram.vars.cursor.system` flag which user could use to switch between system/custom cursors

src/studio/studio.c
```diff
@@ -2332,22 +2332,21 @@ static void blitCursor(Studio* studio)
...
+			}[CLAMP(sprite, 0, 2)];
+		
+		if(tic->ram->vram.vars.cursor.system)
+		{
+			bank = &getConfig(studio)->cart->bank0;
+		}
+		
         const tic_palette* pal = &bank->palette.vbank0;
-        const tic_tile* tile = &bank->sprites.data[sprite];
+        const tic_tile* tile = &studio->config->cart->bank0.sprites.data[sprite];
```
nesbox, line 2349: you always use the cursor from the config cart, this code will not work if I want to change the cursor from the game

nesbox (changes requested):

src/studio/studio.c
Miguel-hrvs: That's true. I'll look into that.

src/studio/studio.c
```diff
@@ -2329,37 +2329,34 @@ static void blitCursor(Studio* studio)
...
+			for(s32 y = s.y, endy = MIN(y + TIC_SPRITESIZE, TIC80_FULLHEIGHT), i = 0; y != endy; ++y, dst += TIC80_FULLWIDTH - TIC_SPRITESIZE)
+				for(s32 x = s.x, endx = x + TIC_SPRITESIZE; x != endx; ++x, ++i, ++dst)
+					if(x < TIC80_FULLWIDTH)
+					{
+						u8 c = tic_tool_peek4(tile->data, i);
+						if(c)
+							*dst = tic_rgba(&pal->colors[c]);
+					}
+		}
         else if(sprite == 0) return;
```
nesbox, line 2359: :) we don't draw anything if `tic->ram->vram.vars.cursor.system == false`

nesbox (changes requested):

Miguel-hrvs:
This is fixed in sokol-wip.

Issue #1986, opened 2022-07-20
Bug: Bookmarks "sticks" the code position


BuoYancYdabl:
**Issue**
Sometimes when you have bookmarks, you can use the other part of code not marked by bookmarks. And for some reason after you open sprite or map editor and then go back your code editor position will be on one of this bookmarks. It sets which bookmark it will stick to by random. I had 2 bookmarks and after going back and forth between editors my code was randomly jumping between this two bookmarks.

**How to recreate**
1) Put bookmark anywhere you want;
2) Scroll somewhere else, write some code so the pointer will define its new position;
3) Open map or sprite editor, choose diferent sprites by pointer or do something;
4) Go back.
You'll see pointer is now on the bookmark and not where it was supposed to be.

TimotheeGreg:
That's due to using F1 to switch to code editor. That's true also when you switch from game to code editor with Esc+F1.
And when you have several bookmarks it will move to the next bookmark each time.

PR #2630, opened 2024-07-11
fix f1 hotkey conflict

Miguel-hrvs:
Fixes: https://github.com/nesbox/TIC-80/issues/1986

### DIFF
--- a/src/studio/editors/code.c
+++ b/src/studio/editors/code.c
@@ -2959,7 +2959,7 @@ static void processKeyboard(Code* code)
     bool usedKeybinding = true;
 
     // handle bookmarks
-    if(keyWasPressed(code->studio, tic_key_f10))
+    if(keyWasPressed(code->studio, tic_key_f7))
     {
         if(ctrl && shift)
         {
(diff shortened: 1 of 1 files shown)


Miguel-hrvs:
Wait, I have to find something different than f10

src/studio/editors/code.c
```diff
@@ -2959,7 +2959,7 @@ static void processKeyboard(Code* code)
     bool usedKeybinding = true;
 
     // handle bookmarks
-    if(keyWasPressed(code->studio, tic_key_f1))
+    if(keyWasPressed(code->studio, tic_key_f7))
```
nesbox, line 2962: We already use the `F7` hotkey to save the cover image for the game but only in game mode. 
To fix the #1986 issue I think we need to filter `F1` presses when you switch to the code editor using hotkeys,
I mean don't allow bookmarking when you just switched from other editor using `F1`.

Miguel-hrvs: It would also be a good solution. Useful too to fix the android problem when pressing ctrl + s writes s. Don't know why that one doesn't happen on desktop.

Miguel-hrvs: Just made a proper fix for this, I'll submit it

Issue #2515, opened 2024-04-19
Changing tile in map editor changes tool


SolarLune:
If you are in the map editor and change the selected tile, the tool will change back to Draw [1]. That's fine, but it would probably be simpler if it stayed with whatever tool you initially selected (so that you can, for example, replace multiple tiles with the Fill tool easily).

PR #2631, opened 2024-07-11
Fix changing tile in map editor changes tool + not needing to hold shift to show tiles

Miguel-hrvs:
Fixes: https://github.com/nesbox/TIC-80/issues/2515
Also makes shift behave the same as clicking with the mouse the tiles menu, this way there's no need to keep holding the shift key to select tiles.


nesbox (approved):

Miguel-hrvs:
Seems that I need to make more changes for the unintended tool change when selecting tile.

PR #2645, opened 2024-07-31
add Nim Wasm template

Archargelod:
Nim is a language that compiles to C, so thank you to the person who implemented the C template header.
I've only had to translate it to Nim with c2nim tool and add some niceties.

Nimscript config config.nims, that's used for building, is mostly from WASM-4 template with TIC-80 specific memory constraints/addresses. To build a project - the only dependency (other than the Nim compiler) is WASI-SDK.

Included are two ported examples: default cart and bunnymark program, which I used for testing.


joshgoebel:
Super nice.  What's the thinking with `internal` vs `tic80`?

Archargelod:
> What's the thinking with `internal` vs `tic80`?

1. I've wanted `btn` and `btnp` to work as in lua api and return `bool` type  (in C they return `unsigned int` for some reason), but I also didn't want to make original procedures to be unavailable (in case that uint32 has some usages).
So right now it is possible to access these symbols by importing them this way:
```Nim
import cart/tic80 except btn, btnp
from cart/internal import btn, btnp

assert btn(P1_Up) is uint32
```
or with namespaces:
```Nim
import cart/tic80
from cart/internal import nil

assert internal.btn(P1_Up) is uint32
assert btn(P1_Up) is bool
```
2. "cart/internal" is a bindings file, where "cart/tic80" is for nicer functions absent or abstracted from tic-80 api.

nesbox (approved):

PR #2649, opened 2024-08-08
Fix gcc truncate and overflow warnings

Miguel-hrvs:


src/studio/editors/code.c
```diff
@@ -140,11 +140,11 @@ static void drawStatus(Code* code)
         tic_api_rect(code->tic, 0, TIC80_HEIGHT - Height, TIC80_WIDTH, Height, getCodeColor(code));
         if(!bb->battle.hidetime)
         {
-            sprintf(code->status.size, "%i/%i", (u32)strlen(code->src), bb->limit.current);
+            snprintf(code->status.size, sizeof(code->status.size), "%i/%i", (u32)strlen(code->src), bb->limit.current);
 
-            char buf[sizeof "00:00"];
+            char buf[32];
```
nesbox, line 145: Why do you replaced `sizeof "00:00"` with 32, we don't need 32 bytes here

src/studio/config.c
```diff
@@ -248,8 +248,8 @@ static void saveOptions(Config* config)
 {
     const struct StudioOptions* options = &config->data.options;
 
-    string buf;
-    sprintf(buf.data, JSON(
+    char buf[2048];
```
nesbox, line 251: why do you removed using of `string` struct and increased it from 1024 to 2048

src/studio/editors/map.c
```diff
@@ -237,7 +237,7 @@ static void drawTileIndex(Map* map, s32 x, s32 y)
 
     if(index >= 0)
     {
-        char buf[sizeof "#9999"];
+        char buf[16];
```
nesbox, line 240: same here, we don't need 16 bytes here

nesbox (changes requested):

Miguel-hrvs:
I'll try to improve it

PR #2662, opened 2024-08-26
Proposal: remove support for `sokol`

aliceisjustplaying:
As far as I can tell this is not used anywhere anymore. It's not part of [the latest release](https://github.com/nesbox/TIC-80/releases/tag/v1.1.2837), doesn't get auto-built in the CI, and there aren't even instructions to build in the README either.

There may be use cases I'm not aware of, though.

If it's truly not in use, my proposal would be to simply remove it.


nesbox:
Interesting, I had plans to use **Sokol** library as the default renderer for **win/linux/mac/wasm/android** and I started implementing it on this branch https://github.com/nesbox/TIC-80/tree/sokol

nesbox:
We can remove Sokol support for now, but in the future, I'm going to bring it back by default

nesbox (approved):

aliceisjustplaying:
Yeah I didn't see that one.

I stumbled into this bc based on my research it doesn't handle keyboard layout changes which I kind of needed for the AZERTY fix.

aliceisjustplaying:
There are ways to do keyboard shortcuts that are keyboard-layout independent except here we do need that, for example.

Issue #2520, opened 2024-04-21
Keyboard shortcut `alt-5` conflicts with `{` character on macOS with a French keyboard


Ecco:
TIC-80 is currently unusable for Mac users in France. Indeed, to type a brace character `{` on a French Mac keyboard you're supposed to hit `Alt-(`.

However, when a user with a French Mac keyboard layout presses `Alt-(`, TIC-80 switches to the Music editor instead of yielding a `{` character.

Note: On those keyboards, the `(` key is shared with the `5` key. To get a `5` you're supposed to press `Shift-(`. I guess this somehow trips TIC-80 that confuses `Alt-(` with the `Alt-5` shortcut (that, on this keyboard, should therefore be `Alt-Shift-(`.

![French Mac Keyboard and parenthesis key](https://github.com/nesbox/TIC-80/assets/55989/a87de444-c957-4494-a588-9691cea21cbf)

I tested this on the current macOS release as well as on the latest HTML build of the main branch. The bug is present in both builds.

aliceisjustplaying:
There is a way to detect they keyboard layout in SDL2: https://gist.github.com/g2p/8597984

KTamas:
@Ecco PR with fix at https://github.com/nesbox/TIC-80/pull/2657.

aliceisjustplaying:
@Ecco PR with fix at https://github.com/nesbox/TIC-80/pull/2657.

Issue #2661, opened 2024-08-25
If`alt+1` produces a key that's valid in TIC-80, it gets inserted in the code editor instead of executing the shortcut


aliceisjustplaying:
Per the title, for some reason the alt+1 shortcut will insert `1` but only under Linux. 

Found while working on https://github.com/nesbox/TIC-80/pull/2657.

aliceisjustplaying:
The Linux thing was a red herring: this happens with every keyboard layout where `alt+1` produces a valid key. 

On macOS with the default US keyboard, `alt+1` produces `¡`, which is not valid in TIC-80. On Windows 11 with the default US keyboard, alt+1 does nothing. Likely the reason this was not caught before.

aliceisjustplaying:
This is a problem specifically when switching away the editor and switching back.-.

Miguel-hrvs:
Is very similar to the f1 bookmark problem that was fixed recently.

Issue #2660, opened 2024-08-25
TIC-80 doesn't work when forced to run under Wayland


aliceisjustplaying:
SDL2 apps run under X11/Xwayland by default. If I force TIC-80 to run under Wayland with `SDL_VIDEODRIVER=wayland` (which should work (?)), it launches but I see no UI whatsoever (do get audio though).

Found while working on https://github.com/nesbox/TIC-80/pull/2657.

nesbox:
Maybe we need to play with these options in SDL2 cmake
https://github.com/libsdl-org/SDL/blob/f070c83a6059c604cbd098680ddaee391b0a7341/CMakeLists.txt#L490
![image](https://github.com/user-attachments/assets/5302aabd-b6cd-4df2-af58-b0f3f705e1bb)

aliceisjustplaying:
Ok, `sdl-gpu` breaks under Wayland, haven't gone too deep why, it's an old library the first place. if I compile TIC-80 without it, I can run it on Wayland but it does not handle scaling well (this is on a retina display with scaling set to 200%) which is the next thing I'll look into.

![Screenshot 2024-08-26 at 11 43 36 PM](https://github.com/user-attachments/assets/ee35a327-b956-45bf-840e-92899bd60b51)

aliceisjustplaying:
I do wonder how much work would it be to remove the `sdl-gpu` dependency and code the CRT feature without it, may look into that as well.

ndrew222:
Is there any fix for this? I am finding an issue running TIC-80 on Sway window manager

bakuhatu:
> Is there any fix for this? I am finding an issue running TIC-80 on Sway window manager
> 
> ```
> INFO: Unable to initialize SDL Video: -1, wayland not available
> ```

I'm on labwc, but this worked for me:

`SDL_VIDEODRIVER=x11 tic80`

Issue #1761, opened 2021-12-22
Tic80 integrated editor not fully usable with Neo keyboard layout


dvdkhlng:
Hi,

for German keyboards there exists an ergonomically optimized keyboard layout called Neo [1]. It's something like Dvorak but optimized for german language and programming.  E.g. on X-Windows you can activate it via

``setxkbmap de neo``

Neo defines additional shift-keys named Mod3 and Mod4 to make special characters and command keys more reachable from the home row.  Unfortunately keys on "Layer 4" do not seem to work in the Tic-80 editor.  E.g. pressing Right_Alt+W should produce a Backpace, but for some reason that does not work in Tic-80.

``xev`` shows the following event sequence for the Neo version of Backspace:

```
KeyPress event, serial 38, synthetic NO, window 0x4a00001,
    root 0x22f, subw 0x0, time 22043892, (79,64), root:(84,262),
    state 0x0, keycode 108 (keysym 0xfe11, ISO_Level5_Shift), same_screen YES,
    XKeysymToKeycode returns keycode: 94
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyPress event, serial 38, synthetic NO, window 0x4a00001,
    root 0x22f, subw 0x0, time 22045139, (79,64), root:(84,262),
    state 0x20, keycode 25 (keysym 0xff08, BackSpace), same_screen YES,
    XKeysymToKeycode returns keycode: 22
    XLookupString gives 1 bytes: (08) "
    XmbLookupString gives 1 bytes: (08) "
    XFilterEvent returns: False

KeyRelease event, serial 38, synthetic NO, window 0x4a00001,
    root 0x22f, subw 0x0, time 22045628, (79,64), root:(84,262),
    state 0x20, keycode 25 (keysym 0xff08, BackSpace), same_screen YES,
    XKeysymToKeycode returns keycode: 22
    XLookupString gives 1 bytes: (08) "
    XFilterEvent returns: False

KeyRelease event, serial 38, synthetic NO, window 0x4a00001,
    root 0x22f, subw 0x0, time 22046346, (79,64), root:(84,262),
    state 0x20, keycode 108 (keysym 0xfe13, ISO_Level5_Lock), same_screen YES,
    XKeysymToKeycode returns keycode: 23
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False

```

[...]

PR #2663, opened 2024-08-26
Fix TIC-80 when using a French AZERTY layout

aliceisjustplaying:
Fixes https://github.com/nesbox/TIC-80/issues/2520. `alt+[grave]12345` is not triggered if an AZERTY layout is set (the user can still use esc + function keys). This allows AZERTY keyboards to fully (as far as I can tell) work with TIC-80.

There is also some groundwork laid for a potential fix of https://github.com/nesbox/TIC-80/issues/1761 in the future, as well as detecting other layouts; this is more work though and definitely a separate PR. Let me know if you'd rather not have this bit and I can dumb it down to strictly detecting AZERTY.

Tested on MacOS 14, Windows 11 and Linux (EndeavourOS/Arch + Wayland + KDE Plasma 6).

Notes for Linux:
* All SDL2 apps, therefore TIC-80 as well run under Xwayland by default. Keyboard layout change detection will only work if done via `setxkbmap` (or something that calls that). If I change the keyboard from KDE Plasma 6 directly, the layout change itself it is not passed down/detected. There isn't anything we can do about this.
* TIC-80 doesn't work when forced to Wayland (`SDL_VIDEODRIVER=wayland`), only Xwayland. It appears to start and has audio but no window is shown. That's a separate bug to track down. Created #2660.
* alt+1 will insert a `1` for some reason when already on the code tab.  That's a separate bug to track down. Created #2661.

### DIFF
--- a/src/system/sdl/main.c
+++ b/src/system/sdl/main.c
@@ -145,6 +145,7 @@ static struct
     {
         bool state[tic_keys_count];
         bool pressed[tic_keys_count];
+        bool isFrenchKeyboard;
         char text;
 
 #if defined(TOUCH_INPUT_SUPPORT)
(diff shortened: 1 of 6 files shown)


src/system/sdl/main.c
```diff
@@ -1062,6 +1062,26 @@ static void handleKeydown(SDL_Keycode keycode, bool down, bool* state, bool* pre
...
 
+char* detect_keyboard_layout() 
+{
+    char q = SDL_GetKeyFromScancode(SDL_SCANCODE_Q);
+    char w = SDL_GetKeyFromScancode(SDL_SCANCODE_W);
+    char y = SDL_GetKeyFromScancode(SDL_SCANCODE_Y);
+
+    char* layout = "unknown";
+
+    if (q == 'q' && w == 'w' && y == 'y') layout = "qwerty"; // US etc.
```
nesbox, line 1073: can we use enum here instead of strings?
smth like
```c
typedef enum
{
tic_layout_qwerty,
...
} tic_layout;
```

nesbox (changes requested):

nesbox:
@aliceisjustplaying now we have a conflict, could you pls resolve it?\
Thanks

src/system/sdl/main.c
aliceisjustplaying: Done in 4f54237

aliceisjustplaying:
Merge conflict fixed

nesbox (approved):

PR #2665, opened 2024-08-26
Fix issue with alt+1 on certain layouts

### LINKED ISSUE #2661: If`alt+1` produces a key that's valid in TIC-80, it gets inserted in the code editor instead of executing the shortcut

aliceisjustplaying:
Fixes #2661. Draft because might fold it into #2663.


Miguel-hrvs:
Fixes the problem with es layout

aliceisjustplaying:
I'm working on a more comprehensive fix around keyboard layouts because this whole thing is more complex than I first thought; that being said, this is still an improvement and worth merging.

nesbox (approved):

PR #2675, opened 2024-09-05
Update to Zig 0.14.0-dev, run through `$ zig fmt`

Des-Nerger:


aliceisjustplaying:
@Des-Nerger Is there a reason for guiding people towards a - presumably unstable - dev release as opposed to the current 0.13.0 release?

nesbox (approved):

Des-Nerger:
@aliceisjustplaying, it may work with older versions (like the aforementioned 0.13.0), but that's not guaranteed. Would require extra testing.

PR #2676, opened 2024-09-05
use modulus to loop map

anescient:
Large values for map offsets can cause slowdown due to being unwound with increment loops on each tile drawn. This change unwinds the coordinates with modulus instead.

### DIFF
--- a/src/core/draw.c
+++ b/src/core/draw.c
@@ -279,13 +279,12 @@ static void drawMap(tic_core* core, const tic_map* src, s32 x, s32 y, s32 width,
     for (s32 j = y, jj = sy; j < y + height; j++, jj += size)
         for (s32 i = x, ii = sx; i < x + width; i++, ii += size)
         {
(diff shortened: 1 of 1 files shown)


joshgoebel (approved):
Wow.  Looks pretty awesome to me.

anescient:
I just found there's an existing function tic_modulo. This would be better:
`            s32 mi = tic_modulo(i, TIC_MAP_WIDTH);
            s32 mj = tic_modulo(j, TIC_MAP_HEIGHT);`
Should I create another request?

anescient:
I've changed this to a draft because tic_modulo may be broken; it produces out of range values if x is a negative multiple of m. I'm going to examine things and see if correcting this will break things or if things are currently broken.

anescient:
From what I see nothing depends on broken behavior from tic_modulo, and I did find a related bug and created an issue for it.

PR #2680, opened 2024-09-09
break out fs_isdir and use it to test fs argument

### LINKED ISSUE #2572: `--fs` accepts files, not just folders; errors follow
`--fs` correctly throws an error if a directory ***or a file*** doesn't exist, however TIC-80 is more than happy to start anyways with a file as an argument. Once started, `save` (predictably) fails.

Present in latest upstream as of commit https://github.com/nesbox/TIC-80/commit/bf5b95cf31bc40cc85ab400d723f95cb841436bb.

image

anescient:
This resolves #2572 using the existing isdir function. If the fs argument specifies a file rather than a folder it's treated as an error.

At least it will when it doesn't break the rpi build.


anescient:
It's clear what was wrong with the rpi build, and I've addressed that. I don't know what's wrong with macos-arm64, now, but I don't think it's this code.

Miguel-hrvs:
The macos build likes to fail randomly due to some github actions weirdness.

nesbox (approved):

Issue #2022, opened 2022-10-14
Request: flood fill function - `fillc(x, y, color)`


eggbanana173:
SPANISH:
mi sugerencia es una función llamada "fillc(x, y, color)"
esta función rellenaría la pantalla de un color definido en una posición especifica como si fuera la herramienta de Paint "rellenar"
sin eliminar las figuras como lo hace "cls(color)"
Ejemplo:
```lua
-- este hipotetico ejemplo hace que rellene a un triangulo que es un "trib()"

function TIC()
 cls()
 trib(25,25,45,80,80,50,4)
 fillc(50,50,4)
end`
```

darltrash:
Translation:

My suggestion is a new function like "fillc(x, y, color)"
Which would fill the entire continuous space of a certain color, like the "fill" feature in MS Paint.
Ejemplo:

```lua
-- This hypothetical example fills a triangle created by "trib()"

function TIC()
 cls()
 trib(25,25,45,80,80,50,4)
 fillc(50,50,4)
end
```

joshgoebel:
Personally I think this would make a great function for an external library. :-)

nesbox:
Hmm, I think it's out of scope of the current api,  and @joshgoebel is right that this would make a great function for an external library

anescient:
I've implemented this in #2683

PR #2683, opened 2024-09-13
flood fill paint function

anescient:
This is an implementation of Paul Heckbert's flood fill, providing the functionality of the QBasic PAINT method. This fills a contiguous area of solid color starting from a seed point.

All languages are supported, but not all languages are tested.

There's a closed/wishlist for this:  #2022

Edit: Actually this isn't quite like QB's PAINT. Continued below.

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -633,14 +633,15 @@ enum
                                                                                                                         \
                                                                                                                         \
     macro(paint,                                                                                                        \
-        "paint(x y color)",                                                                                             \
+        "paint(x y color bordercolor=-1)",                                                                              \
                                                                                                                         \
-        "This function fills an area.",                                                                                 \
-        3,                                                                                                              \
+        "This function fills a contiguous area with a new color.\n"                                                     \
+        "If bordercolor is given fill will extend to color boundary.",                                                  \
+        4,                                                                                                              \
         3,                                                                                                              \
         0,                                                                                                              \
         void,                                                                                                           \
-        tic_mem*, s32 x, s32 y, u8 color)                                                                               \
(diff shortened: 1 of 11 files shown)


anescient:
I was able to generate some torture tests that hit the stack limit around 3000, so I've changed the stack size to 4000. I'm not sure how to generate an absolute worst case, so I'm not sure how large is too large.

sinsinpub:
just a random thought since qb mentioned, 2nd param bordercolor also frequently used and worth to be implemented?
image

anescient:
> just a random thought since qb mentioned, 2nd param bordercolor also frequently used and worth to be implemented?

Definitely worth it but might need another implementation; I think this painter depends on stopping at its own color.

anescient:
I went back to check the behavior of QB paint when it's given a border color, and it turns out there's _always_ a border color that just defaults to the paint color. For reference:

qbpaints1

I'm thinking tic paint should replicate this behavior if a border color is provided, but then unlike QB paint if no border color is given it should fill whatever contiguous space starts at the seed point, like I've implemented so far.

anescient:
I'm finding QB paint is silly and shouldn't be replicated exactly. When a boundary color is given it's hard to predict whether the paint will leak through its own color:

qbpaints2

Rather than using QB as reference the behavior I'm going to pursue is: if a border color is given then paint to within the border color or the paint color itself, and as before, when no border color is given then effectively every color is border except the color at the seed point.

anescient:
I found a worse worst case for memory usage which suggested the stack limit should be about 10,000 to be safe, so I've replaced the stack with a much smaller queue.

nesbox (approved):

PR #2688, opened 2024-09-24
Embed R in TIC-80 as an additional scripting language

bryce-carson:
R is my favourite langauge, and it isn't embedded in many (if any) applications that don't focus on simply providing a front-end to R _for R_. This pull request is intended to fully embed R as a new scripting language in TIC-80.

I have worked on this for a couple days now.

At the moment I am working on getting GitHub Actions setup so that my fork will have its own GitHub Pages website containing the exported HTML document of the literate Org source file I'm doing my development work in. I'll also work, eventually, on getting GitHub Actions to automatically tangle the source file `r.c` from the Org file and check that into the repository's src/api folder.

Code review and advice on embedding is greatly appreciated, and eventually I'd love for this to become an official language and not simply exist in my experimental fork.


bryce-carson:
https://www.r-project.org/

bryce-carson:
It makes sense to use https://github.com/wch/r-source as a Git submodule.

bryce-carson:
At the moment my HTML file which should be relocated to the root directory as index.html exists at src/c/TIC-80/src/api/r.html

bryce-carson:
The literate document will help the maintainer of TIC-80 review the code, and offer guidance on its development. :shrug: I think.

bryce-carson:
I am at the point of conducting my first informal integration tests. I was able to resolve all of my compile-time issues except the issue quoted in the last commit. I will reproduce that quotation in this comment for easier viewing.

```
/home/bryce/src/c/TIC-80/src/api/r.c:175:10: fatal error: ../build/assets/rdemo.tic.dat: No such file or directory
  175 | #include "../build/assets/rdemo.tic.dat"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/r.dir/build.make:76: CMakeFiles/r.dir/src/api/r.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:283: CMakeFiles/r.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

bryce-carson:
https://github.com/nesbox/TIC-80/tree/0e34eaaa50bfdefa546988edc0c8efa1bf280130/build/tools

I had seen these tools before, but had no idea what they were used for, and I couldn't compile the bin2txt.

```
/usr/bin/ld: /tmp/ccYrGMuK.o: in function `main':
bin2txt.c:(.text+0x139): undefined reference to `compress2'
collect2: error: ld returned 1 exit status
```

I have _zlib-devel_ installed on Fedora, so I should have `zlib.h` and compress2 should be defined.

I only realized what the tools were used for [after seeing this file](https://github.com/nesbox/TIC-80/blob/main/assets.bat), which given it is a BAT file won't run except on Windows. Another PR _I may or may not open_ could address that by writing operating-system agnostic CMake scripts to identify all demo and benchmark files and run the necessary tools on them at build time, automatically (mark in place of benchmark is a bad naming convention in my opinion). That will make updating these files much easier.

bryce-carson:
> https://github.com/nesbox/TIC-80/tree/0e34eaaa50bfdefa546988edc0c8efa1bf280130/build/tools
> 
> I had seen these tools before, but had no idea what they were used for, and I couldn't compile the bin2txt.
> 
> ```
> /usr/bin/ld: /tmp/ccYrGMuK.o: in function `main':
> bin2txt.c:(.text+0x139): undefined reference to `compress2'
> collect2: error: ld returned 1 exit status
> ```
> 
> I have _zlib-devel_ installed on Fedora, so I should have `zlib.h` and compress2 should be defined.
> 
> I only realized what the tools were used for [after seeing this file](https://github.com/nesbox/TIC-80/blob/main/assets.bat), which given it is a BAT file won't run except on Windows. Another PR _I may or may not open_ could address that by writing operating-system agnostic CMake scripts to identify all demo and benchmark files and run the necessary tools on them at build time, automatically (mark in place of benchmark is a bad naming convention in my opinion). That will make updating these files much easier.

Ideally, actually, none of these DAT files would be present in the repository and the required DAT files would be built at compile-time _only_.

bryce-carson:
[r.html.tar.gz](https://github.com/user-attachments/files/17143336/r.html.tar.gz)

This compressed HTML file is the exported literate document in which I wrote my work.

This was my first real C project, I'm still learning the language after studying Narain Gehani's _C: an Advanced Introduction_.

I was able to sort out the DAT files, and I was able to compile and link my code into TIC-80.

The application doesn't crash on start, which is good, but anything related to R simply doesn't work. When I run `new r` an R cartridge isn't created, a Lua cartridge is created, and I cannot find rdemo.tic or rmark.tic when browsing the directories from within TIC-80.

I think at this point I will need some hand-holding to complete this PR, learn what I need to, and advance in my understanding of complex C software.

Regards and thanks!

bryce-carson:
A couple of notes:
- When no languages are defined the _new_ command will inform the user of that, but if the user tries to create a cartridge of a language not properly integrated---or not integrated at all---then TIC-80 will simply create a Lua cartridge.
- Sometimes the build will include the Lua langauge despite -DBUILD_WITH_LUA=OFF.

bryce-carson:
@sthilaid, would you mind looking at this pull request and telling me if I've missed a number of big steps? I'm beginning to really struggle with this language integration and I'd love to hear about your experience integrating Scheme. I'll update this PR with r.c so you can look at that directly.

cmake/r.cmake
```diff
@@ -0,0 +1,44 @@
...
+option(BUILD_WITH_R "R Enabled" ${BUILD_WITH_ALL})
+message("BUILD_WITH_R: ${BUILD_WITH_R}")
+
+if(BUILD_WITH_R AND NOT N3DS AND NOT BAREMETALAPI)
+  set(TIC_BUILD_WITH_R TRUE)
+
+  if(NOT BUILD_STATIC)
+    include("cmake/renv.cmake")
+
+    set(R_LIB_DIR "/usr/include/R")
```
aitap, line 13: Instead of hard-coding the paths, it's better to run the particular R you would like to link with and ask it about its include path, something like `Rscript -e 'cat(R.home("include"))'`.

src/api/r.c
```diff
@@ -0,0 +1,331 @@
...
+	/* TODO: I think it makes sense to evaluate the code, then check if this
+	 * exists, then end the init function with evaluating the TIC-80 function. */
+	Rf_eval(Rf_mkString("if (exists(\"BOOT\") && is.function(BOOT) BOOT() else BOOT <- function() NULL;;"),
+					R_GlobalEnv);
+
+	tic_script *config = (tic_get_script(tic));
+	for (int kiwi = 0; kiwi <= config->api_keywordsCount; kiwi++) {
+#define defineRFn_(keyword) Rf_eval(Rf_mkString("t80."#keyword" <- function(...) .C(r_"#keyword", ...);"), R_GlobalEnv);
+#define doDefineRFunction(keyword) defineRFn_(keyword)
+		doDefineRFunction(config->api_keywords[kiwi]);
```
aitap, line 129: Unfortunately, `#` is a preprocessor operation, not a runtime operation. This expands at compile time into

```c
Rf_eval(Rf_mkString("t80.""config->api_keywords[kiwi]"" <- function(...) .C(r_""config->api_keywords[kiwi]"", ...);"), R_GlobalEnv);;
```

I'm afraid you might have to use a temporary buffer with C string functions in order to achieve proper string concatenation.

src/api/r.c
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	/* TODO: I think it makes sense to evaluate the code, then check if this
+	 * exists, then end the init function with evaluating the TIC-80 function. */
+	Rf_eval(Rf_mkString("if (exists(\"BOOT\") && is.function(BOOT) BOOT() else BOOT <- function() NULL;;"),
+					R_GlobalEnv);
+
+	tic_script *config = (tic_get_script(tic));
+	for (int kiwi = 0; kiwi <= config->api_keywordsCount; kiwi++) {
+#define defineRFn_(keyword) Rf_eval(Rf_mkString("t80."#keyword" <- function(...) .C(r_"#keyword", ...);"), R_GlobalEnv);
```
aitap, line 127: Every object that is not protected from garbage collection or explicitly referenced by another protected object [can be destroyed during a call to almost any R API](https://cran.r-project.org/doc/manuals/R-exts.html#Garbage-Collection) (except those that protect or unprotect the values). During `Rf_eval()`, the argument given to the function is not protected, so could be potentially collected.

Moreover, the values evaluated by `Rf_eval` are usually [special call objects of type `LANGSXP`](https://cran.r-project.org/doc/manuals/R-exts.html#Evaluating-R-expressions-from-C), not strings. Evaluating a string, like in R itself, gives the string back (`eval("hello")` ⇒ `"hello"`).

Perhaps you could write yourself a special helper that would call `Rf_mkString`, protect the return value, [parse the string](https://cran.r-project.org/doc/manuals/R-exts.html#Parsing-R-code-from-C), protect the return value, `Rf_eval()` the produced expressions, then unprotect the temporary variables and return?

cmake/r.cmake
bryce-carson: Yes, the paths I determined with `R CMD config --cppflags` and `R CMD config --ldflags`. I had much difficulty with CMake, despite reading some of the manual and beginning the tutorial until I felt comfortable to try again (after linking against a library in one of the introductory tutorial steps).

I have written an additional CMake file to generate a header which I can include to set the environment variables appropriately during runtime when they aren't set otherwise. At the moment I want to focus on getting a working integration running on my own system before I worry further about making it portable; it was a struggle to get where I am now.

I do know about the need for environment variables, libraries, etc. that R is dependent upon for its operation. I'll focus on them again later.

bryce-carson: > Yes, the paths I determined with `R CMD config --cppflags` and `R CMD config --ldflags`.

By this I mean that I'm aware of the commands and will use them later. For now, hard-coding was a troubleshooting methodology to simply ensure that they're there.

src/api/r.c
bryce-carson: > I'm afraid you might have to use a temporary buffer with C string functions in order to achieve proper string concatenation.

Indeed. I realized that the last time I was working on this and I am nearly finished working on that. Thank you for pointing this out.

src/api/r.c
bryce-carson: Those are all good points. Thank you for your review.

PR #2698, opened 2024-10-07
Fix ruby mouse coordinates

### LINKED ISSUE #2695: Discrepancy Between Returned Mouse Coordinates and Actual Cursor Position in Ruby
There is a discrepancy between the returned coordinates from the mouse method and the actual mouse cursor position in Ruby. When the cursor is visually placed at (0,0) on the screen, the mouse method returns coordinates of (8,4). The returned value seems to be consistently offset by a certain amount from the screen coordinates.
```
# script:  ruby

def TIC
	cls 13
	rect mouse[:x],mouse[:y],8,8,3
	print "#{mouse[:x]},#{mouse[:y]}"
end
```
video1

anescient:
This change makes the Ruby mouse() function take coordinates from the core api rather than directly from memory.

Resolves #2695

### DIFF
--- a/src/api/mruby.c
+++ b/src/api/mruby.c
@@ -1012,12 +1012,13 @@ static mrb_value mrb_mouse(mrb_state *mrb, mrb_value self)
 
     tic_core* core = getMRubyMachine(mrb); tic_mem* tic = (tic_mem*)core;
 
+    tic_point pos = core->api.mouse(tic);
     const tic80_mouse* mouse = &tic->ram->input.mouse;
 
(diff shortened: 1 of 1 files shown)


MineRobber9000:
If the memory has the wrong values, I would imagine it'd make more sense to fix whatever's putting the wrong values in memory rather than just taking the mouse coordinates from a different place.

anescient:
I don't think anything is wrong; I think it's by design. There's a border area so raw mouse coordinates on the window won't match the drawable area.

nesbox (approved):

PR #2718, opened 2024-11-21
Update jsdemo.js

bettywolfy:


bettywolfy:
why is this here? i don't remember doing that :v

PR #2745, opened 2025-02-02
use -O2 and add arm64 build

Miguel-hrvs:
Works great on 64bit raspberry's


Miguel-hrvs:
Hold until we check if O3 breaks arm version

nesbox (approved):

PR #2763, opened 2025-03-17
Baremetal Pi fix

josei:
Raspberry Pi baremetal versions have been broken for a while.

I've been doing extensive testing on a recently-purchased Raspberry PI 3 (which wasn't supported by *any* of the existing releases, not even v0.90, unlike *older* Raspberry PI 3 boards, which I know should be the same hardware but it looks they're not).

I finally found that simply including the reference to `fopen`/`fclose` commands in the bytebattle code in `studio.c` breaks the build and makes it not go past the 4 color gradient firmware screen. Commenting out this code fixes it.

There was an additional bug about `tic80` folder check always returning `false` under baremetal. This bug was easier to find thanks to logs working again after the previous change.

Additionally, the build only boots with just LUA support, so the next `cmake` command is required:
```
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=baremetalpi/toolchain.cmake -DBUILD_WITH_LUA=ON -DBUILD_WITH_MOON=OFF -DBUILD_WITH_FENNEL=OFF -DBUILD_WITH_WREN=OFF -DBUILD_WITH_RUBY=OFF -DBUILD_WITH_WASM=OFF -DBUILD_WITH_SCHEME=OFF -DBUILD_WITH_SQUIRREL=OFF -DBUILD_WITH_PYTHON=OFF -DBUILD_WITH_JS=OFF -DBUILD_WITH_JANET=OFF ..
```

More research is needed to fix the rest of languages.

Sample build [rpi3-baremetal.zip](https://github.com/user-attachments/files/19292978/rpi3-baremetal.zip). The compiler version I used is `arm-gnu-toolchain-13.2.Rel1-aarch64-arm-none-eabi`.

Additional bugs that are fixed in this PR for the Baremetal version:
- Removed pixel cursor (not needed anymore since a cursor is rendered by TIC-80).
- `del` command now supports deleting files and folders.
- `.local` folder is hidden from console (simply moved to parent folder outside `tic80`).
- `tic80.com` option is not shown, as baremetal has currently no network support.
- Support USB storage if plugged in at boot (fall back to SD card if no external storage is found).
- Added locale support for ES type layout (pass `-DKEYBOARD_LAYOUT=ES` to `cmake` to use it). It also requires disabling ALT+2 and ALT+3 shortcuts, as otherwise there's no way to write `@` and `#` symbols. Layouts for more languages can be added in the future (although they require recompiling).
- Fixed potential buffer overflow if pressing too many keys at the same time (use `TIC80_KEY_BUFFER` limit).
- Fixed missing mouse-up event.


src/studio/fs.c
```diff
@@ -452,14 +456,12 @@ bool tic_fs_deldir(tic_fs* fs, const char* name)
...
-    return false;
-#else
     const char* path = tic_fs_path(fs, name);
 
     const FsString* pathString = utf8ToString(path);
+#if defined(BAREMETALPI)
+    return f_unlink(pathString);
+#else
     bool result = tic_remove(pathString);
     freeString(pathString);
```
joshgoebel, line 466: Don't we need to free the string in BOTH cases?

src/studio/fs.c
```diff
@@ -427,9 +430,10 @@ void tic_fs_enum(tic_fs* fs, fs_list_callback onItem, fs_done_callback onDone, v
 bool tic_fs_deldir(tic_fs* fs, const char* name)
 {
 #if defined(BAREMETALPI)
-    // TODO BAREMETALPI
-    dbg("tic_fs_deldir %s", name);
-    return 0;
+    const char* path = tic_fs_path(fs, name);
+    const FsString* pathString = utf8ToString(path);
+
```
joshgoebel, line 435: Pretty sure you also need to `free` here.

src/studio/fs.c
```diff
@@ -548,8 +550,7 @@ static void onEnumPublicDirsDone(void* data)
 bool fs_isdir(const char* path)
 {
 #if defined(BAREMETALPI)
-    // This function isn't applicable to baremetal.
-    return false;
+    return fs_exists(path);
```
joshgoebel, line 553: This isn't exactly the same as "is directoy" since it's only checking for existence, correct?

src/studio/studio.c
```diff
@@ -413,9 +413,20 @@ char getKeyboardText(Studio* studio)
...
 
+#ifdef KEYBOARD_LAYOUT_ES
+        // US KEYS:                     " abcdefghijklmnopqrstuvwxyz0123456789-=[]\\;'`,./< ";
+        static const char Symbols[] =   " abcdefghijklmnopqrstuvwxyz0123456789'!`+cn'o,.-< ";
+        static const char Shift[] =     " ABCDEFGHIJKLMNOPQRSTUVWXYZ=!\" $%&/()??^*CN\"a;:_> ";
+        static const char Alt[] =       "                            |@#        []} {\\     ";
+#else
         static const char Symbols[] =   " abcdefghijklmnopqrstuvwxyz0123456789-=[]\\;'`,./ ";
         static const char Shift[] =     " ABCDEFGHIJKLMNOPQRSTUVWXYZ)!@#$%^&*(_+{}|:\"~<>? ";
+        static const char Alt[] =       "                                                  ";
```
joshgoebel, line 424: Why does the "default" (other than ES) not match the lines removed?

src/studio/studio.c
```diff
@@ -1827,6 +1841,7 @@ static void processShortcuts(Studio* studio)
 #if defined(BUILD_EDITORS)
         else if(studio->mode != TIC_RUN_MODE && studio->config->data.keyboardLayout != tic_layout_azerty)
         {
+#ifndef KEYBOARD_LAYOUT_ES
```
joshgoebel, line 1844: Why would this only apply to ES layout?

joshgoebel:
Breaking work like this up into smaller commits could make it easier to review in the future.

josei:
> Breaking work like this up into smaller commits could make it easier to review in the future.

Thanks for the advice. I rebuilt the commit history to make it easier to understand.

src/studio/studio.c
josei: In ES layout, ALT+1 outputs `|`, ALT+2 outputs `@`, and ALT+3 outputs `#`. With those shortcuts in place (which can be still accessed using ESC, F1, F2, F3, etc.), there's no way to type symbols `|`, `@`, or `#`.

src/studio/studio.c
josei: It's been refactored for clarity, since the `caps` boolean only reverses `shift` condition, and the ternary conditional operator was already too complex even without the `alt` modifier.

src/studio/fs.c
josei: Yes, fixed in e8ac15d488e3fee7143aa5dd5c870ddbe7d3026e.

src/studio/fs.c
josei: Fixed in e8ac15d488e3fee7143aa5dd5c870ddbe7d3026e.

nesbox (approved):

PR #2790, opened 2025-06-20
Fix saving options to JSON

Softkandi:
Remove the JSON macro, which doesn't work as intended because it literally pastes preprocessor code into the options file, corrupting it.


nesbox (approved):
thank you, and sorry for the delay with the review 🤦

Issue #2147, opened 2023-02-28
libretro: no cursor despite `mouse_cursor = "arrow"`


Des-Nerger:
I've done some tests and now suspect it's the whole drawing mechanism inside the `tic80_libretro.c` that is not working:
```c
	tic_mem* tic = (tic_mem*)state->tic;
	...
	switch (cursortype) {
		...
		case MOUSE_CURSOR_ARROW:
		        // seems to draw into where nothing is seen
			tic_api_tri(tic, state->mouseX, state->mouseY, state->mouseX + 3, state->mouseY, state->mouseX, state->mouseY + 3, state->mouseCursorColor);
			tic_api_line(tic, state->mouseX + 3, state->mouseY, state->mouseX, state->mouseY + 3, tic_color_black);
```

Des-Nerger:
Ok, I found a fix, even though it's not very efficient:
```diff
diff --git a/src/system/libretro/tic80_libretro.c b/src/system/libretro/tic80_libretro.c
index c830d60..2dd417d 100644
--- a/src/system/libretro/tic80_libretro.c
+++ b/src/system/libretro/tic80_libretro.c
@@ -737,7 +737,7 @@ void tic80_libretro_mousecursor(tic80* game, tic80_mouse* mouse, enum mouse_curs
        // Determine which cursor to draw.
        switch (cursortype) {
                case MOUSE_CURSOR_NONE:
-                       // Nothing.
+                       return;
                break;
                case MOUSE_CURSOR_DOT:
                        tic_api_pix(tic, state->mouseX, state->mouseY, state->mouseCursorColor, false);
@@ -753,6 +753,7 @@ void tic80_libretro_mousecursor(tic80* game, tic80_mouse* mouse, enum mouse_curs
                        tic_api_line(tic, state->mouseX + 3, state->mouseY, state->mouseX, state->mouseY + 3, tic_color_black);
                break;
        }
+       tic_core_blit(tic);
 }
 
 /**
```

nesbox:
I'm not sure about using `tic_core_blit()` here, I think you should draw your cursor directly in the screen buffer `game->screen`

Des-Nerger:
Yes, that's why I said "it's not very efficient".
But then one had to reimplement the functionality of `tic_api_tri`, `tic_api_line` and `tic_api_pix`, basically reduplicating the code for the case of drawing directly to `game->screen`, which imo isn't a very neat solution either.
[...]

PR #2822, opened 2025-09-01
Libretro cursorfix

imsys:
Should solve the cursor not showing in some games #2147 and the cursor trailing bug (also mentioned in that issue).

I have also set the cursor to start hidden before the first move. Otherwise we have to wait the first 5 seconds to it to hide in case we want to take screenshots of the game.


imsys:
@RobLoach

PR #2826, opened 2025-09-08
Informative screenshot filenames

imsys:
I was taking screenshots of many games to add them to screenscraper.fr but then I did not really know from what game each screenshot was.
For me, the screenshots from retroarch are way more useful, they contain the ROM name + datetime, that's much more helpful and informative, so I thought to just copy their pattern.

But in retroarch they save as .png for screenshots and .mkv for video. As TIC save everything as gif, I was unsure if that distinction would go to the filename. I basically save everything with the same pattern, but if requested, I could make a change to show the distinction between screenshot and video.

The pattern is:
ROMname-YYMMDD-hhmmss-ms.gif

Retroarch doesn't goes as far as the milliseconds, but I thought it would be nice in case of rapid consecutive screenshot taking.

By the way, I kind of dislike it opening up the picture viewer, as it may disrupt during intense gameplay while taking the screenshot. But I also understand the reason: so the user knows where the screenshots are being saved.

### DIFF
--- a/src/studio/studio.c
+++ b/src/studio/studio.c
@@ -24,6 +24,8 @@
 
 #if defined(BUILD_EDITORS)
 
+#include <time.h>
+#include <sys/time.h>
 #include "editors/code.h"
 #include "editors/sprite.h"
 #include "editors/map.h"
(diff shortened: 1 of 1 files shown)


nesbox (approved):
ok, lets try :)
thank you

PR #2831, opened 2025-09-11
[Android standalone] update SDL files to v2.32.10

imsys:
The SDL2 update (86a4cc3a92c8cb95241d3e157143d14364c32402) worked great on all plataforms, but on Android that got crashes.


imsys:
This is now working on my device.
I can install the apk, and the virtual keyboard works.

nesbox (approved):
👍

PR #2832, opened 2025-09-11
Informs retroarch that PNG extension is also supported

imsys:


imsys:
I tested it locally on my Linux, on Android standalone and retroarch. All seems to be working as expected, to load .tic and .png files.

nesbox:
If you want to run a tic80 binary built with the `BUILD_EDITORS=OFF` option, you need to embed the cart binary into it.
You can find here https://github.com/nesbox/TIC-80/blob/main/src/studio/screens/start.c#L189 how we load the embedded cart. And here https://github.com/nesbox/TIC-80/blob/main/src/studio/screens/console.c#L2070, is how we embed it.

nesbox (approved):

PR #2833, opened 2025-09-11
Replace msf_gif submodule with single header (–711 MiB) + macOS build fix

imsys:
Saves 711 MiB of disk space by adding the msf_gif header file instead of the whole repo that contains mostly test cases.
It saves 354 MiB of bandwidth, but it is stored twice, in `.git/modules/vendor/msf_gif` and `vendor/msf_gif`, so that's why it takes double the size.

The full TIC80 repo after getting all the submodules, before making any build takes 1203.7 MiB, so this lib alone was taking 59% of disk space. I verified and it was due to many test cases that are unnecessary to the TIC80 project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -370,7 +370,7 @@ jobs:
       - name: Build
         run: |
           cd build
-          cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_STATIC=ON -DBUILD_SDLGPU=On -DBUILD_WITH_ALL=ON ..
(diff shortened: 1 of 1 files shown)


imsys:
Sure, but I’ll send a few more contributions before adding to the credits. :)
I’m kind of on vacation, and I plan to dedicate about two weeks to the TIC-80 project.
My goal is to fix most of the games that were broken by regressions. I think I’ve identified about three or four different cases (or causes).

After that, there’s a very difficult task that I’m not completely sure I can solve. Some games run too slowly on cheap retro handhelds, like the Emuurom and others. I’m not sure if the issue comes from the libretro core specifically, or if the standalone version on those devices has the same slowdown. Either way, I plan to investigate and try to improve performance.

Issue #2811, opened 2025-08-15
Android need to automatically open  .tic files with  TIC-80


Popolon:
It's very add to make the link between .tic TIC-80 cardridges files and TIC-80 on Android today.

the   intent-filter section in file build/android/app/src/main/AndroidManifest.xml

need the missing parameters ``<data android:pathPattern=".*\\.tic">``

Like this:
```
            <!-- Drop file event -->
            <!--
            <intent-filter>
                <action android:name="android.intent.action.VIEW" />
                <category android:name="android.intent.category.DEFAULT" />
                <data android:mimeType="*/*" />
                <data android:pathPattern=".*\\.tic" />
            </intent-filter>
```

nesbox:
Please check if it works https://github.com/nesbox/TIC-80/pull/2834

imsys:
Apk is at https://github.com/nesbox/TIC-80/actions/runs/17691922422/artifacts/4001827125

imsys:
Strange, I'm quite sure I got it working in a previous test.
Anyway, I will rest a bit and check this later.

imsys:
Oh, maybe it's the tooling. I'm using more up to date building tools locally.

imsys:
![Image](https://github.com/user-attachments/assets/7f1060d6-40c5-4cea-aa27-94d27e639a6a)
![Image](https://github.com/user-attachments/assets/1f974cf8-0d84-495b-a65e-553a9697a4a8)

imsys:
> This version of TIC-80 when launching the icon (not a cart) just make black screen or display the icon in middle of black screen in waydroid, didn't tested on my phone. Previous one worked.

Sorry, try this one that is from a the [latest build](https://github.com/nesbox/TIC-80/actions/runs/17740019940/job/50411418202): [artifact apk](https://github.com/nesbox/TIC-80/actions/runs/17740019940/artifacts/4015077543)

PR #2837, opened 2025-09-14
[Android standalone] fix association of .tic and .png to TIC-80 app

imsys:
Solves #2811

### DIFF
--- a/build/android/app/src/main/AndroidManifest.xml
+++ b/build/android/app/src/main/AndroidManifest.xml
@@ -89,11 +89,24 @@
             </intent-filter>
 
             <!-- Drop file event -->
+            <!-- Handle .tic files -->
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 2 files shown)


imsys:
The option does show up now, but when I actually try to load the game in that way, it just opens the app, without the game..

imsys:
This works, but there is still a problem that can be fixed in a later pull request:
When pressing the back button, for 98% of the games it will quit the game by showing the code, while in only 2% of the games it shows the menu.

imsys:
I just tried to build it locally with BUILD_EDITORS=OFF, but this makes the games not load, it just opens a blackscreen... it would require a bit of debugging to use BUILD_EDITORS=OFF

nesbox:
Thanks for the PR.
Regarding the menu that appears when you press Esc, that's how it used to be. You'd press Esc and always go to the menu. But people started pressuring me to remove that feature. For some reason, it really annoyed them, maybe because when they're quickly browsing through a bunch of games, they find it tiresome to constantly exit through the menu. 
So, I gave in, even though I don't like it. 
My original idea was always to show that menu.

PR #2839, opened 2025-09-14
[WIP] Add htmllocal export to use locally built wasm

CarlQLange:
Add htmllocal export to use locally built wasm, so you can use the latest (or a custom) TIC-80 build in the browser. This is a WIP, it works for my use case but I'm not that sure if it is a good fit (if there are reasons for you not to want to do it this way). 

Basically, we just bake the local build/html files (wasm included) into the TIC-80 executable at build time. Then when you do `export htmllocal mygame`, it'll use those files instead of doing a network call and getting them from the server instead (also means you can export for web while offline for what it's worth). 

It works well enough for what I wanted, which was to test out the safari keyboard fix I added a PR for earlier!

Claude code wrote practically all the code here. I will not be hurt if you don't merge this, very little human effort went into it but I do find it useful and functional :)


CarlQLange:
Going to close this as it's too complex for me to personally understand. If someone would like to resurrect it feel free.

PR #2884, opened 2026-02-06
Adds savestate support to the libretro core

imsys:
This is for Lua cartridges only.

When trying to save on non-Lua, the user gets a message:
"Savestate currently not supported in non-Lua TIC-80 cartridges"

---
So, I basically was playing [Multiple Maze](https://tic80.com/play?cart=2866), the game is a great puzzle, but there is no save implemented. I was getting quite far, then I got soft-locked and the only option was starting over.

So then I thought, let's try to get an AI to implement a savestate, and it did it successfully.

So I then started to check game by game to improve the savestate functionality.

It is working quite nice on most simple games, but there are some still with some inconsistent behavior during loading saves, I will keep working on this on the next few days. I'm already sharing this in case someone wants to give some input/opinion.

Although I do not think it will be possible to save the state of EMUUROM, as that has basically all the memory space for the Lua code already filled. I'm leaving EMUUROM to test later in the hope that fixing smaller games could make more complex games works. But If I still can't get that to work, I could just add an `If EMUUROM` or maybe `If code memory almost full` give some error message saying unsupported, because it would be better than the current long freeze that I get when trying to save it.


imsys:
I noticed that PICO-8 has savestate support, so I looked how they are doing it.

They basically use a Lua fork ([samhocevar/z8lua](https://github.com/samhocevar/z8lua)) with serialization support:
https://github.com/jtothebell/fake-08/blob/f6bab5a7ba521ce440e45d1aeef6122674be6ee9/source/vm.cpp#L1493-L1522
https://github.com/jtothebell/fake-08/blob/f6bab5a7ba521ce440e45d1aeef6122674be6ee9/platform/libretro/libretro.cpp#L560-L585

I think the main serialization code that would need to be imported would be found at [samhocevar/z8lua/blob/zepto8/eris.c](https://github.com/samhocevar/z8lua/blob/zepto8/eris.c)

nesbox:
There is a call here we would rather leave to you: continue, or close.

The shape is the problem: a save assembled by walking `_G` is partial by construction — one empirical rule per cartridge, and closures and coroutines cannot be captured at all. A restore that is nearly right is worse than no save. If you continue, we would want a written contract for what a savestate contains, and a refusal with a message wherever that contract cannot be met. If that is more than you want to take on, closing it yourself is a fine outcome.

PR #2886, opened 2026-02-18
Fix crash when pressing ESC during code editor popup hide animation (#2881)

### LINKED ISSUE #2881: TIC-80 crashing when you leave the code editor while the UI is animating
self-explanatory title but i managed to replicate this crash:

press any icon that triggers a animation, the bookmark icon, the goto icon, the find icon and the outline menu in the topright of the code editor and then press whatever icon you clicked again. while the animation is happening and it's not fully invisible yet, press ESC to leave the code editor and TIC-80 will crash

TetherIO:
### Problem
TIC-80 crashes when you close a code editor popup (Find/Goto/Bookmarks/Outline) and press ESC again while the popup is still animating closed.

Studio routes ESC to `code->escape()` whenever `code->mode != TEXT_EDIT_MODE` therefore ESC can call `escape()` multiple times during the popup hide animation. On the first call, `escape()` restores `popup.prevPos/prevSel` and clears them; the second call would attempt to restore NULL into the pointers and crash during the editor update.

### Fix
Make the popup-close path in `Code::escape()` safe against being invoked again mid-animation:
- If `popup.prevPos` is already NULL, don't restart the hide animation, don't touch cursor pointers.

This keeps the behavior the same for a normal single ESC, but prevents the crash from repeated ESC during the hide animation.
No behavioral changes outside the timing window (repeated ESC while the popup is already closing).

Fixes #2881

### DIFF
--- a/THIRD_PARTY_LICENSES.md
+++ b/THIRD_PARTY_LICENSES.md
@@ -36,7 +36,6 @@ THE SOFTWARE.
 
 
 
-
 ### blip_buf $vers
 Band-Limited Audio Buffer
 Shay Green <gblargg@gmail.com>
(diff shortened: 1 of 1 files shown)


TetherIO:
Closing this PR because it was opened from main and picked up additional commits.
Bug fix moved to #2887 (tied to the issue).

Issue #2821, opened 2025-08-31
Android: Arrow keys not recognized in the editors/console when using a wireless keyboard.


diemartindie:
Hello!

Sorry in advance if this has already been reported. I couldn't find anything from some searches.

When using an *Xtrike Me KB-303* wireless keyboard on Android, all keys work as expected except for the arrow keys which don't work, but just in the editors and console. The arrow keys work in-game as expected.

The arrow keys also work as expected everywhere when connecting a cheap PC keyboard (*Weibo* brand, unknown model) via an USB to USB-C adapter.

* **Device:** Lenovo Tab M10 FHD Plus
* **OS:** Android 10
* **TIC-80 Version:** 1.1.2837 from GitHub releases

I can give more info if needed.

Thank you in advance!

IgorGRBR:
I can also confirm this behavior with my no-name bluetooth keyboads.

As a quick and dirty workaround - one can use [the keymapper app](https://f-droid.org/en/packages/io.github.sds100.keymapper/) to map physical arrow keys and esc to virtual ones, when the TIC-80 app is open.

ky-tt:
I had the same issue before and it seems to be gone with 80926eba362315809925d9df3528662874d6b3ae . Thanks @RDeckard 🙏

Issue #623, opened 2018-06-15
External keyboard support is broken in the Android version


briangordon:
Version: Android

- [ ] The alt+1, alt+2, etc key combinations to tab through the editor work a few times and then stop working until you restart the app.
- [ ] After switching apps using the standard Android soft button, the TIC-80 code editor stops responding to letter keys until you press escape to go back to the command line.
- [ ] The gamepad can't be configured to use external keyboard keys- you're stuck with the on-screen keys.

aliceisjustplaying:
Using the latest release pro version, on a Pixel 2 with LineageOS 19 (Android 13):
* Cannot repro the first issue
* Cannot repro the second issue
* Need a gamepad to test this

@briangordon is this still happening to you?

PR #2889, opened 2026-02-27
Creates the BUILD_SURF build option

imsys:
I'm trying to make TIC-80 a bit more friendly on handhelds.

For handhelds, I don't need all that is provided by BUILD_EDITORS, but I still want the SURF. So I tried to separate those options as much as I could. To prevent problems when trying to decouple it, I have set:
```cmake
if(BUILD_EDITORS)
    set(BUILD_SURF ON)
endif()
```
So BUILD_SURF is kind of an essential part of BUILD_EDITORS.

The trickiest problem is that `surf.c` is quite dependent on `console.c`, for that I just asked for help with the Antigravity AI editor, and its suggestion was to create `console_minimal.c`, Idk what you will think of that. In my opinion this is not ideal but it gets the job done, haha.

After succeeding, I just fixed up the menu.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,9 @@ option(BUILD_WITH_ZLIB "Build with zlib linked" ON)
 option(TIC80_TARGET "Target binary suffix")
 option(PREFER_SYSTEM_LIBRARIES "Prefer link with system libraries" OFF)
 
+if(BUILD_EDITORS)
+    set(BUILD_SURF ON)
+endif()
 
 if(NOT TIC80_TARGET)
     set(TIC80_TARGET tic80)
--- a/build/android/app/build.gradle
+++ b/build/android/app/build.gradle
@@ -27,7 +27,12 @@ android {
                 arguments "APP_PLATFORM=android-31"
             }
             cmake {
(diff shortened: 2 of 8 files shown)


imsys:
When no game loaded:
Screenshot from 2026-02-27 05-43-47

When a game is loaded: Screenshot from 2026-02-27 05-44-09

nesbox (approved):
Let’s go ahead and merge this. However, I think we should look into removing the surf-to-console dependency in the future, perhaps by moving those shared functions into a dedicated common utility file.

sandio:
Hi @imsys, thank you for your work.

Setting `BUILD_EDITORS=OFF` for Android makes it handheld-friendly, but it also removes the code/sprite/map/SFX/music editors and console, reducing TIC-80 to a game browser. Two things are lost:

- Game creation on Android — anyone whose primary device is a phone/tablet
- Inspecting games from Surf — opening any game's code, sprites, and maps is one of TIC-80's best educational features

This also affects several recent and open Android editor-related efforts: PRs #2896, #2897 (merged same day), #2593, and open issues #2821, #2292 (still on v1.2 board), #623

Instead of removing editors entirely, would any of these work?

1. Keep `BUILD_EDITORS=ON` and let users choose surf-only via runtime settings
2. Two APK variants — _full_ and _player_

imsys:
I agree with you, and thanks for commenting this. Please see #2937.
In my opinion it would be better to keep the studio by setting `BUILD_EDITORS=ON`, and just set the `back action` (`SDLK_AC_BACK`) to open the menu instead of the console.

Issue #2855, opened 2025-10-21
Horizontal scroll direction is wrong on macOS


DagAgren:
When scrolling with a trackpad on macOS, the vertical scroll direction is correct and consistent with the OS, but the horizontal scroll direction is inverted.

PR #2903, opened 2026-03-04
[Fix] [macOS] correct horizontal scroll direction

RDeckard:
## Why

Original issue: https://github.com/nesbox/TIC-80/issues/2855

Horizontal trackpad scrolling is inverted on macOS while vertical scrolling is already consistent.

## What

- Scoped `scrollx` inversion in `src/studio/studio.c` so it is not applied on macOS.
- Kept behavior unchanged on non-macOS platforms.

## Impact

- macOS horizontal scroll direction now follows expected OS behavior.
- No intended behavior change on non-macOS builds.

### DIFF
--- a/src/studio/studio.c
+++ b/src/studio/studio.c
@@ -2296,7 +2296,9 @@ static void processMouseStates(Studio* studio)
 
         state->dbl.ticks++;
     }
+#if !defined(__TIC_MACOSX__)
     tic->ram->input.mouse.scrollx *= -1;
+#endif
 }
 
 #if defined(BUILD_EDITORS)


nesbox (approved):

src/studio/studio.c
```diff
@@ -2296,7 +2296,9 @@ static void processMouseStates(Studio* studio)
 
         state->dbl.ticks++;
     }
+#if !defined(__TIC_MACOSX__)
     tic->ram->input.mouse.scrollx *= -1;
```
joshgoebel, line 2300: Curious does this depend on whether someone has inverted scrolling set or not?  Mac still allows up to be up or up to be down, does it not?

PR #2906, opened 2026-03-09
Fix WASM templates after lld --stack-first default change

RayenChalouati:
This updates the WASM templates to stop relying on --global-base=98304 with post-data stack placement, which no longer works reliably since newer versions of LLVM’s WebAssembly linker enable --stack-first by default.

TIC-80 reserves the first 96 KiB of memory (0x00000–0x18000) for internal systems. Previously, templates depended on the linker placing the stack after the data segment, allowing the data segment to start at --global-base=98304.

However, with --stack-first now enabled by default in lld, the stack is placed at the start of memory, which breaks this assumption and risks overlapping with TIC-80’s reserved memory.

Changes

The templates were updated to use the stack-reservation workaround instead:

Explicitly enable --stack-first

Increase stack size to 106496 bytes (96 KiB reserved + 8 KiB usable stack)

This effectively reserves the TIC-80 internal memory region while keeping a small usable stack.

Updated files

templates/rust/.cargo/config.toml

templates/nim/config.nims

templates/c/Makefile

templates/d/dub.json

templates/zig/build.zig

demos/wasm/build.zig

demos/bunny/wasmmark/build.zig

Short README notes were also added in each template to document the 96 KiB + 8 KiB stack layout.

Verification

Verification was static only:

Rust TOML parses correctly

dub.json parses correctly

git diff --check reports no issues

I was not able to run full builds because this environment does not have the required toolchains installed (cargo, rustc, zig, nim, ldc2, clang).

Remaining risk

For Zig templates, the oversized stack reservation pattern is now used, but --stack-first is not explicitly set and therefore still depends on Zig’s linker behavior.

References

llvm#151015

llvm#166998

WebAssembly linker documentation for LLVM

### DIFF
--- a/demos/bunny/wasmmark/build.zig
+++ b/demos/bunny/wasmmark/build.zig
@@ -1,5 +1,8 @@
 const std = @import("std");
 
+const tic80_reserved_memory = 96 * 1024;
+const tic80_stack_size = 8 * 1024;
+
 pub fn build(b: *std.Build) !void {
     const optimize = b.standardOptimizeOption(.{});
     const target = b.resolveTargetQuery(.{ .cpu_arch = .wasm32, .os_tag = .freestanding });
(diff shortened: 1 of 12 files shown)


nesbox:
@joshgoebel could you pls review?

imsys:
> @joshgoebel could you pls review?

If the comment is related to the baremetal failed build. That is the same error I already fixed in #2907.

nesbox (approved):

joshgoebel:
This is a neat trick. :-) Cool.

PR #2908, opened 2026-03-12
Persist config across updates

### LINKED ISSUE #2531: Keep settings when updating TIC-80
I don't know if that's possible but it would be great to keep menu settings and config file settings when updating to a more recent TIC-80 version.

Currently, it's difficult to use themes and I have to change the gamepad settings each time I update TIC-80 because I use a special (bépo) keyboard.

imsys:
This closes #2531.

I understand that this was probably done in a way to avoid breaking changes between versions, but this also seems to be a source of complain of what the default settings should be, specially regarding if the ESC key should go to the menu or console.

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -764,7 +764,7 @@ static void loadCartSection(Console* console, const tic_cartridge* cart, const c
 
 static char* getDemoCartPath(char* path, const tic_script* script)
 {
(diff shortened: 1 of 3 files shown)


imsys:
Actually, this doesn't seem to be a problem. I think the key problem is the persistence of config in the web version, at least regarding the ESC key.

Issue #1950, opened 2022-06-06
Start with "dev mode" on when booting into the console


blinry:
**Problem**

When opening https://tic80.com/create or running any of the official releases, "dev mode" is off in the settings. This means that after running a game, upon hitting Esc, people have to select "close game" from the menu every time they want to return to the console. Also, the shortcuts F1 to F5 don't work to return to specific editors.

This greatly hinders the development flow, especially for new people who don't know that they can disable this behavior in the settings. I think that both advanced programmers as well as newcomers would much prefer having dev mode on when they want to develop a game. Also personally, disabling dev mode would be the first thing I'd always do when opening TIC-80. So I think this should be the default behavior.

When I gave a workshop on TIC-80 last month, it was unfortunate to have to explain how to enable dev mode to each of the over 20 participants. I would rather want a smooth first-time experience for them, where they can switch between making and trying the game quickly, without having to find the correct menu entry before.

**Proposed solution**

When booting the console without immediately starting a game, enable dev mode. In this case, it is very likely that this is the desired behavior.

@nesbox [pointed out](https://github.com/nesbox/TIC-80/discussions/1894#discussioncomment-2889705) that the menu would also be good to have when using `surf`. I'd propose to disable dev mode again when running `surf`, in that case.

borbware:
Big agree!

Also, while we're at it, make ESC go to console in devmode instead of code editor, like it used to be.

These minor changes in workflow have basically made me stay in 0.90 for the time being...

darltrash:
In my opinion, A thing I like a lot about TIC-80 is how much it involves the average user into making things and dividing developers from regular users is not helpful in general.
I think it'd be better if we could just jump back into the editor with a shortcut of some sorts, kind of like if we could use CTRL+R to go back to the editor and NOT disable the menu or whatever at all.

borbware:
> I simply don't think the dev mode is either useful or necessary

Dev mode is necessary! For so many TIC-80 users, "press ESC to go to console" is deeply ingrained to their muscle memory, so I can't be the only one to want that functionality back.

borbware:
> At least make dev mode easy to toggle
imo, it IS easy to toggle. `menu` -> options > dev mode

blinry:
Maybe as a clarifying question to @nesbox: why does the menu (and the "non-dev mode") exist at all? I always imagined that it served two purposes:

[...]

PR #2909, opened 2026-03-12
Fix saving config in web version (wasm)

imsys:
From what I could understand, the settings in the options menu were being saved on `freeConfig()`, called only by `studio_delete()`. I suspect this all happens when exiting the TIC80.

This was not being called in the web version. Even by going in the menu and Exit TIC80, it wasn't being saved. The way I was testing it was by checking the timestamp in the place shown in the screenshot bellow.

Screenshot from 2026-03-14 12-04-12

Gemini first suggested to do a change in `src/system/sdl/main.c`, that did not fix the issue, and I did not notice any change in saving behavior.

Then I mainly had to do some changes in the mainmenu.c, and I thought it would be a good idea to rename the (config.c)`save` to `saveConfigCart` so it would be a bit more descriptive and it would be a bit more distinctive than `saveOptions`.

Having that fixed, I started to wonder if I should have the initial changes in `src/system/sdl/main.c` kept or not.

Gemini initially game this explanation for that:
> The Emscripten logic in `src/system/sdl/main.c` was using an uninitialized internal variable `FS.syncFSRequests` to decide when to sync the in-memory filesystem to the browser's persistent storage (IDBFS). Since this variable was always undefined, the sync was never triggered, causing all changes (including settings) to be lost upon refreshing or closing the page.

Then I questioned it again using a clean context and I got:
> The changes the AI suggested are essentially a "guard" or "mutex" to prevent multiple filesystem sync operations from happening at the same time.
> 
> Should you keep it?
> Yes, you should keep it.
> 
> Even if it didn't solve your specific bug, the new version is technically more robust than the original. Here is why:
> 
> **Prevents Race Conditions**: `FS.syncfs` is asynchronous (it happens in the background). If your app requests a save, and then requests another save before the first one finishes, it can cause data corruption or browser errors in IndexedDB.
> **State Management**: The original code checked `FS.syncFSRequests == 0`, but `FS.syncFSRequests` is not a standard Emscripten property. It was likely a custom variable that wasn't being updated correctly. The new `Module.syncing` flag explicitly tracks the state.
> **Clean Callback**: The new version ensures that `Module.syncing` is set back to false only once the browser confirms the write is finished.

### DIFF
--- a/src/studio/config.c
+++ b/src/studio/config.c
@@ -140,7 +140,7 @@ static void reset(Config* config)
     saveConfig(config, true);
 }
 
-static void save(Config* config)
+static void saveConfigCart(Config* config)
 {
     *config->cart = config->tic->cart;
     readConfig(config);
@@ -199,7 +199,8 @@ void initConfig(Config* config, Studio* studio, tic_fs* fs)
         .studio = studio,
         .tic = getMemory(studio),
         .cart = realloc(config->cart, sizeof(tic_cartridge)),
-        .save = save,
+        .saveConfigCart = saveConfigCart,
+        .saveOptions = saveOptions,
         .reset = reset,
         .fs = fs,
     };
--- a/src/studio/config.c
+++ b/src/studio/config.c
@@ -192,45 +192,6 @@ static void loadOptions(Config* config)
     }
 }
 
-void initConfig(Config* config, Studio* studio, tic_fs* fs)
-{
-    *config = (Config)
-    {
-        .studio = studio,
-        .tic = getMemory(studio),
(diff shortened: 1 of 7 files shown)


imsys:
macos build failed for this reason:
`clang: error: unable to execute command: posix_spawn failed: Resource temporarily unavailable`
I will just have to trigger a rebuild

imsys:
I edited the post to reflect all the changes in this PR.

imsys:
This would somewhat have helped a concern mentioned at #1950 about the having to change to dev mode in the web [tic80.com/create](https://tic80.com/create) as those settings would not persist.

nesbox (approved):

PR #2910, opened 2026-03-12
Upgrade GitHub Actions

imsys:
Trying to fix these warnings:
> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/upload-artifact@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
Screenshot from 2026-03-12 20-28-58

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,17 +2,13 @@ name: Build
 
 on: [push, pull_request]
 
-# as recommended by: https://github.com/actions/checkout/issues/1590
-env:
-  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
-
 jobs:
   # === Windows ===
   windows:
     runs-on: windows-2022
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           submodules: recursive
           fetch-depth: 0
@@ -29,7 +25,7 @@ jobs:
           cmake --build . --config MinSizeRel --parallel
 
       - name: Deploy
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v6
         with:
           name: "tic80-windows"
           path: |
@@ -61,7 +57,7 @@ jobs:
             git
             gcc
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           submodules: recursive
           fetch-depth: 0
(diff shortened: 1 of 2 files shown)


imsys:
Still getting the warning for: `windows-mingw` and `html` at https://github.com/nesbox/TIC-80/actions/runs/23028853744?pr=2910

For the html task, it depends on https://github.com/mymindstorm/setup-emsdk/issues/50

For the mingw, it uses `msys2/setup-msys` or `ruby/setup-ruby`. Ruby is used in other builds that are not showing the warning, so the problem seems to be from `msys2/setup-msys`, but I can't say for sure.

I'd say to approve this first, and later we try to fix the missing ones.

RobLoach:
Emscripten is knowing slow to update to the latest Node.js. If you find their sources, you'll see many issues related to it:
https://github.com/emscripten-core/emsdk/issues?q=is%3Aissue%20state%3Aopen%20node

imsys:
```yml
#.github/workflows/build.yml
      - name: Build js
        run: |
          cd build
          emcmake cmake -DBUILD_SDLGPU=On -DCMAKE_BUILD_TYPE=Release -DBUILD_WITH_LUA=OFF -DBUILD_WITH_JS=ON -DTIC80_TARGET=tic80js .. --fresh
          cmake --build . --parallel
```

```c
// vendor/quickjs/quickjs.c

/* default memory allocation functions with memory limitation */
static size_t js_def_malloc_usable_size(const void *ptr)
{
#if defined(__APPLE__)
    return malloc_size(ptr);
#elif defined(_WIN32)
    return _msize((void *)ptr);
#elif defined(EMSCRIPTEN) || defined(POOR_CLIB)
    return 0;
#elif defined(__linux__) || defined(__GLIBC__)
    return malloc_usable_size((void *)ptr);
#else
    /* change this to `return 0;` if compilation fails */
    return malloc_usable_size((void *)ptr); // <-- error line
#endif
}
```

The `html` build is not getting the `EMSCRIPTEN` set.

imsys:
@nesbox, this is ready!

nesbox (approved):

PR #2911, opened 2026-03-15
Sort files alphabetically when surfing.

imsys:
When I have 1000+ titles in a folder, it gets very difficult to find what I want.

This only sort for local files, as we probably want the web surf to be in a different order. (most voted and newest).


imsys:
This also tests by first using the same case, otherwise the uppercase comes first. and then all the lowercase.

nesbox (approved):

PR #2912, opened 2026-03-17
Adds support for Linux ARM GLESv2

imsys:
This basically adds supports to many devices of Linux handhelds firmwares, like Rocknix, Knulli and others.

```Dockerfile
# ========================================================
# Dockerfile to build TIC-80 for TrimUI (AArch64)
# ========================================================

FROM ubuntu:22.04

# Install build dependencies
RUN apt-get update && apt-get install -y \
    cmake git make xz-utils wget curl python3 ruby rake pkg-config \
    && rm -rf /var/lib/apt/lists/*

# Install Arm GNU Toolchain (GCC 10.3.1, AArch64)
WORKDIR /opt
RUN wget https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-aarch64-none-linux-gnu.tar.xz \
    && tar -xf gcc-arm-10.3-2021.07-x86_64-aarch64-none-linux-gnu.tar.xz \
    && rm gcc-arm-10.3-2021.07-x86_64-aarch64-none-linux-gnu.tar.xz

# Define Toolchain Paths
ENV CROSS_ROOT=/opt/gcc-arm-10.3-2021.07-x86_64-aarch64-none-linux-gnu
ENV PATH="${CROSS_ROOT}/bin:${PATH}"
ENV CROSS_COMPILE=aarch64-none-linux-gnu-

# Install symbol extraction dependencies
RUN apt-get update && apt-get install -y binutils libsdl2-dev libgles2-mesa-dev libegl1-mesa-dev

# Create stub libraries for linking (satisfied by device firmware at runtime)
# We extract the symbol names from the builder's libraries and generate empty C stubs.
# This prevents glibc versioning issues while satisfying the linker.
RUN for lib in /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0 /usr/lib/x86_64-linux-gnu/libGLESv2.so.2 /usr/lib/x86_64-linux-gnu/libEGL.so.1; do \
        outname=$(basename $lib | sed 's/-2.0//; s/\.so\..*/.so/') && \
        echo "Generating stubs for $outname from $lib" && \
        nm -D $lib | grep " T " | awk '{print "void " $3 "(){}"}' > stubs.c && \
        ${CROSS_COMPILE}gcc -shared -o ${CROSS_ROOT}/aarch64-none-linux-gnu/libc/usr/lib/$outname stubs.c; \
    done

# --- Fetch TIC-80 ---
WORKDIR /wrk
#RUN git clone https://github.com/nesbox/tic-80.git && \
#    cd tic-80 && git submodule update --init --recursive
COPY . tic-80

# --- Build TIC-80 ---
WORKDIR /wrk/tic-80/build/trimui

RUN cmake ../.. \
      -DCMAKE_SYSTEM_NAME=Linux \
      -DCMAKE_SYSTEM_PROCESSOR=aarch64 \
      -DCMAKE_C_COMPILER=${CROSS_COMPILE}gcc \
      -DCMAKE_CXX_COMPILER=${CROSS_COMPILE}g++ \
      -DCMAKE_SYSROOT=${CROSS_ROOT}/aarch64-none-linux-gnu/libc \
      -DCMAKE_FIND_ROOT_PATH=${CROSS_ROOT}/aarch64-none-linux-gnu/libc \
      -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER \
      -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
      -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
      -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ONLY \
      -DCMAKE_BUILD_TYPE=Release \
      -DBUILD_PRO=OFF \
      -DBUILD_WITH_ALL=ON \
      -DCMAKE_C_FLAGS_RELEASE="-mcpu=cortex-a53 -Ofast -flto" \
      -DCMAKE_CXX_FLAGS_RELEASE="-mcpu=cortex-a53 -Ofast -flto " \
      -DBUILD_SDL=ON \
      -DBUILD_SDLGPU=ON \
      -DBUILD_EDITORS=OFF \
      -DBUILD_SURF=ON \
      -DPREFER_SYSTEM_SDL2=ON \
      -DBUILD_STATIC=ON \
      -DTIC_DATA_PATH="/userdata/roms/tic80/"

RUN cmake --build . --target tic80 --config Release --parallel

RUN mkdir /output

# then get the file out with:

# docker run --rm -v "$PWD":/output tic-trimui:latest cp /wrk/tic-80/build/trimui/bin/tic80 /output 

RUN ["/bin/bash"]
```


imsys:
I plan to add the `linux-arm64-gles2` to the Github CI/CD, but I will wait for the PR #2910 to be closed, as to avoid the git conflict.

There is also some other things I need to test and review.

imsys:
The failed Ubuntu build is likely due to some instability on the Ubuntu servers—they’ve been targeted by DDoS attacks lately.

tic80-trimui

Anyway, this GitHub workflow build works nicely on my TrimUI under the Knulli firmware.

There are a few points that need to be addressed:

* The cursor stays at the top-left. → I think this could be handled in a different PR. (#2936) One option would be to start it hidden until the first movement and move it with the right stick. Additionally, there could be a hide timeout under a build flag when there is no movement—maybe the timeout would only be desirable during gameplay, not when using the studio.
* There is quite a border compared to the libretro version, which is stripped. I know this is the normal behavior of TIC-80, and this was probably discussed before. Not sure if a menu option or build flag to remove the border would be acceptable.
* I added the build var `TIC_DATA_PATH` to set the folder it uses when browsing, but setting the folder this way makes the built binary not compatible across different OS/firmware. For example, for Knulli it should be `/userdata/roms/tic80/`, but on Rocknix I believe it is `/storage/roms/tic80`. This won’t be a problem since they have their own build workflows with device-specific optimizations, but it would still be nice for anyone to just grab a nightly build here to test things out.

Aside from this, I got Gemini to write a library stub linker that prevents errors from version-specific glibc issues, so I think I’ll later try to unify the two linux-x86_64 versions (`linux-gcc12-glibc235` and `linux-gcc14-glibc239`).

.github/workflows/build.yml
```diff
@@ -155,6 +155,70 @@ jobs:
...
+            -DCMAKE_BUILD_TYPE=Release \
+            -DBUILD_PRO=OFF \
+            -DBUILD_WITH_ALL=ON \
+            -DCMAKE_C_FLAGS_RELEASE="-Ofast -flto" \
+            -DCMAKE_CXX_FLAGS_RELEASE="-Ofast -flto" \
+            -DBUILD_SDL=ON \
+            -DBUILD_SDLGPU=ON \
+            -DUSE_GLES2=ON \
+            -DUSE_EGL=ON \
+            -DBUILD_EDITORS=OFF \
```
computermouth, line 196: `-DBUILD_EDITORS=OFF`, I think this is a bit of an assumption here. Just because they're on arm64 linux doesn't mean the user only intends to play the games.

I'm doing edits on my armhf board.

computermouth:
Left a comment about editors, but just wanted to say I built this on my armhf board and it worked great :+1: 

(I didn't realize my build was using glx/gl swrast via mesa, tic80 is hitting 60fps no problem now on my old board)

.github/workflows/build.yml
imsys: True, it's a bit opinionated. I'm sending new PRs to enable a easier integration with gamepads without requiring to disable the editor tools. I also wonder how much BUILD_EDITORS affects performance, `EMUUROM` is the one that has a bit of slow down in some sections. I will test the difference sometime later.

Ideally we could make many builds, but then it takes forever the Github CI, maybe if we enable caching the compilation of the vendor src would help quite a bit.

computermouth: Similar to my suggestion on the other PR, this sounds like something better suited for a configuration file, rather than a parallel build.

If this shipped with the editor, desktop linux arm64 users would be able to grab a prebuilt binary, and if it were a config, that would also make for a great feature for those handhelds.

imsys: After @nesbox approves #2952, I can then change this to `BUILD_EDITORS=ON`

nesbox:
Merged `main` in to clear the conflict, so the branch carries a current base. One file needed a hand: `build.yml`, where this branch adds the arm64 GLES2 job and `main` has since changed the gcc-14 job's cmake line (ccache, Ninja, PRO) — kept the new job and `main`'s line. The rest merged cleanly.

nesbox:
Correction on my merge: the conflict resolution gave the new `linux-gcc14-glibc239` job main's cmake line (ccache, Ninja, PRO), and that job installs no ccache, so its run failed with `ccache: not found`. Restored the line the branch shipped (`a910faec` then `e4e3155b`): the job installs gcc-14 and builds Release, as its June run did. Every other job kept main's setup.

nesbox:
Rebuilt `build.yml` from `main` and added back only what this branch introduces: the arm64 GLES2 job and the gcc14 job, each with its own steps. The previous resolution had mixed the gcc14 job with `main`'s export-stub steps and left it calling ccache it never installs, which is what the failed runs were. The diff against `main` is additions only now.

nesbox (approved):
Read the flag wiring: GLES2 and EGL are opt-in at the CMake level with the defaults off, and the new job is a self-contained arm64 GLES2/EGL build for a target like a handheld. The conflict was CI only, and the rebuilt `build.yml` is `main`'s plus this branch's two jobs, additions only.

Issue #2615, opened 2024-06-10
PCM channel not working on exported builds


Weeppiko:
Executables built with the `export` command have their PCM channel completely silent

nesbox:
Yes, because we have old binaries for the export on the server, we need to set up our pipeline properly to update them on every build.

PR #2916, opened 2026-03-26
[Fix] [Export] prefer local native template for same-platform export

RDeckard:
## Why

Original issue: https://github.com/nesbox/TIC-80/issues/2615

Native export templates fetched from the server can become stale and may ship outdated runtime behavior.  
This PR does not change the server export pipeline itself; it mitigates the issue on the client side for same-platform native exports.

## What

- Added a same-platform local-template path in `src/studio/screens/console.c`:
  - `win` on Windows
  - `linux` on Linux
  - `mac` on macOS
- Same-platform native export now tries embedding into the local executable template (`fs_apppath()`) first.
- If local template use is not applicable or unavailable, behavior falls back to the existing server export flow.
- Added explicit console messages to show whether export used:
  - local native template
  - server fallback template

## Impact

Same-platform native exports are less sensitive to stale server artifacts, while cross-platform export behavior remains unchanged.

**Reviewer question:** should we also enable local-template fallback for `rpi` on Raspberry Pi builds, or keep `rpi` strictly server-backed for now?

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -2200,8 +2200,67 @@ static void exportGame(Console* console, const char* name, const char* system, n
     tic_net_get(console->net, url, callback, MOVE(data));
 }
 
+static bool canExportNativeFromLocalTemplate(const char* system)
+{
+#if defined(__TIC_WINDOWS__)
+    return strcmp(system, "win") == 0;
+#elif defined(__TIC_LINUX__)
+    return strcmp(system, "linux") == 0;
+#elif defined(__TIC_MACOSX__)
+    return strcmp(system, "mac") == 0;
+#else
+    return false;
+#endif
+}
+
+// Same-platform native export prefers the local executable template.
+// If the local template is not applicable or unavailable, caller falls back to server export.
+static bool tryExportNativeFromLocalTemplate(Console* console, const char* name, const char* system)
+{
+    if(!canExportNativeFromLocalTemplate(system))
+        return false;
+
+    const char* appPath = fs_apppath();
+
+    if(!appPath)
+        return false;
+
+    s32 appSize = 0;
+    u8* app = fs_read(appPath, &appSize);
+
+    if(!app || appSize <= 0)
+        return false;
+
+    if(app) SCOPE(free(app))
+    {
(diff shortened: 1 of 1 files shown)


computermouth:
Tested good for me on armhf linux :+1:

imsys:
I have not tested this, but I think @nesbox uses templates compiled with `BUILD_EDITORS=OFF`, so the exported final game doesn't include the console, code editor and other studio tools. So I think the solution would have to be a bit different to point to a different binary template path.

computermouth:
Maybe that toggle could instead be something that's packed into the binary also, rather than pulling the compiled builds. The binaries would be larger with dead code, but it would make the feature much more portable with way less infrastructure required.

computermouth:
I just tested desktop linux export, and it still ships the editor. I think no-editor-export is actually reserved for the pro version's exports.

imsys:
Another option could be to have the export feature look for template binaries in the TIC-80 data directory, perhaps under a template-export subfolder. The app could optionally online fetch and drop the Windows, macOS, and Linux templates on that folder. (ideally covering multiple architectures, and possibly Android as well)

Users who want custom builds could simply replace those template binaries with their own compiled versions.

I would probably prefer the online download mechanism to be kept, since the infrastructure already exists and avoids increasing the application size, that I personally would prefer to not increase further, specially when targeting lower end devices.

RDeckard:
Updated the PR to make the local-template path best-effort only.

If embedding or writing the local executable template fails, export now falls back to the existing server template flow instead of reporting a local export failure. The fallback message is also only shown for same-platform exports, so cross-platform exports keep the previous behavior/noise level.

This is still intended as a narrow mitigation for stale same-platform server templates, not as a replacement for the server/template pipeline.

nesbox (approved):
The mechanism checks out: `embedCart()` allocates a fresh buffer and `start.c:177-215` already looks for exactly that header in `fs_apppath()`, so the local template produces the shape the engine reads, and the two `SCOPE(free(...))` blocks free different allocations. Like for like as well — the server's universal stub keeps the editors, as your `check-export-stubs.sh` says in its own header. Thanks, merging.

PR #2921, opened 2026-04-09
Fix ttri in low bpp, use segment offsets in tilesheet accessors

### LINKED ISSUE #2325: Blit doesnt work propperly with ttri()
mouse scroll to change blit pages.
[blit_ttri_bug.zip](https://github.com/nesbox/TIC-80/files/12907430/blit_ttri_bug.zip)
only first page of blit works with ttri.

ddelemeny:
This PR allows `ttri` to honor blit_segment correctly. Fixes #2325 

## This is a long standing limitation
For reference, [textri was implemented with total-coordinates system](https://github.com/nesbox/TIC-80/pull/1099#issuecomment-636648340), and this persisted in ttri. My comment at the time suggests that {bank=0,page=0} segments are the only one supported and incorrectly states that a full-width tilesheet can be considered (which is not the case, only full-height seems to be accessible in v0.80 and the bug fell through versions). 

### Modifications of `tic_thilesheet` pixel accessors  in `tilesheet.c` read as follows :
- calculate the bank and page offsets based on xy coordinates and segment spec
- mask xy coordinates to segment-local (0-127) and calculate the tile id using local xy and bank/page offsets

### Modifications in `studio/editors/sprite.c` :
- use the correct segment calculation in `initTileSheet` (the discarded formula did not point to the right segment, it only seemed to work because applied offsets cancelled out)
- use segment-local coordinates in `getIndexPos*`

### Modifications in `core/draw.c`:
- extend the coordinate system to the full width of the sheet in segments where nb_pages > 1

## Resulting behavior
Blit segment can now point to any bank/page, which will be considered as the origin of the (looping) coordinate system. This allows access to any zone of the spritesheet at every bit depth.

## This is a BREAKING change fixing erroneous or underspecified behavior
`ttri` is the only user of the pixel accessors aside from the editor, and it's behavior towards `blit_segment` is incorrect.
As of now, the only usable portion of the tilesheet is {bank_orig=0, page_orig_=0,...} for each bpp, which corresponds to segments 2,4 and 8. With any other segment, it'll draw from the respective 0,0 portion.

This means that other segments are not likely to be in use, because of their non-support. This also means that any code that calls ttri on a segment other than 2,4 and 8 relies on underspecified behavior and will break after this fix.

## Relevant documentation :
- https://github.com/nesbox/TIC-80/wiki/ttri
- https://github.com/nesbox/TIC-80/wiki/Bits-Per-Pixel
- https://github.com/nesbox/TIC-80/wiki/blit-segment

### DIFF
--- a/src/core/draw.c
+++ b/src/core/draw.c
@@ -909,10 +909,10 @@ static tic_color triTexTileShader(const ShaderAttr* a, s32 pixel)
     if(!shaderStart(a, &vars, pixel))
         return TRANSPARENT_COLOR;
 
-    enum { WMask = TIC_SPRITESHEET_SIZE - 1, HMask = TIC_SPRITESHEET_SIZE * TIC_SPRITE_BANKS - 1 };
+    enum { HMask = TIC_SPRITESHEET_SIZE * TIC_SPRITE_BANKS - 1 };
 
     return shaderEnd(a, &vars, pixel, data->mapping[tic_tilesheet_getpix(&data->sheet,
-                     (s32)floor(vars.x) & WMask, (s32)floor(vars.y) & HMask)]);
+                     (s32)floor(vars.x) & (TIC_SPRITESHEET_SIZE * data->sheet.segment->nb_pages - 1) , (s32)floor(vars.y) & HMask)]);
 }
 
 static tic_color triTexVbankShader(const ShaderAttr* a, s32 pixel)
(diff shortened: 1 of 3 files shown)


ddelemeny:
Test cart:
[blit_ttri_test.zip](https://github.com/user-attachments/files/26642584/blit_ttri_test.zip)

Before : 
image
After :
image
[blit_ttri_test.zip](https://github.com/user-attachments/files/26642584/blit_ttri_test.zip)


---
PS : if you read my now deleted comment about an error in the pixel shader, you can safely ignore that. I just had trouble wrapping my head around the pixel-center math again.

nesbox (approved):

PR #2924, opened 2026-04-10
Added mobile touch controls for html export from @matstc

RobinScaletta:
Added mobile touch controls for HTML export. This is based on the work of @matstc. It allows playing TIC-80 games on mobile browsers with a virtual D-pad and buttons.

### DIFF
--- a/build/html/export.html
+++ b/build/html/export.html
@@ -11,59 +11,153 @@
         #game-frame > div { font-size: 44px; font-family: monospace; font-weight: bold;}
 
         /* You can set custom dimensions here */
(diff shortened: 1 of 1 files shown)


matstc:
Also wanted to mention that an 8-way wheelpad would allow for diagonal input as documented in [this post](https://mxtxc.itch.io/ghostwrap/devlog/1468035/using-a-wheelpad-to-move-the-character).

Not to distract from the merging of this PR, which adds a lot of value in itself.

nesbox (approved):

PR #2936, opened 2026-05-05
[SDL] Uses right stick to move cursor + only shows after first move

imsys:
This is mainly addressing issues I mentioned when running it on handheld devices. #2912

At least for now, I have not add any cursor auto-hide, as hiding may be more contentious as to time length before hiding or if hiding is appropriated at all, so maybe it would be better in a separated PR.

### DIFF
--- a/src/system/sdl/main.c
+++ b/src/system/sdl/main.c
@@ -181,6 +181,9 @@ static struct
     struct
     {
         bool focus;
+        bool moved;
+        float x;
+        float y;
     } mouse;
 
     struct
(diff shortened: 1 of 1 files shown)


nesbox:
The `moved` flag and the fractional stick accumulator both read right. One thing worth naming out loud: the right stick stops mirroring the d-pad, and the dead zone is the d-pad's `AXIS_THRESHOLD` (50%), so the cursor only moves past half travel. Branch refreshed onto `main`.

nesbox (approved):
The `moved` flag and the fractional stick accumulator both read right. Naming it out loud: the right stick stops mirroring the d-pad, and the dead zone is the d-pad's `AXIS_THRESHOLD` (50%), so the cursor only moves past half travel.

PR #2939, opened 2026-05-10
Allows to run TIC-80 commands from the system console

imsys:
I noticed that the output of the TIC-80 console was going to my system console, this is quite nice specially when there are a lot of output.

But then I thought: what if we could run commands from the system console too? just mirror both consoles?

This PR allows to do just that, I can type commands using the system console, or through the SDL console. It even allows to run commands while the SDL screen is on a different screen, like on the studio or running a game.

Tested on: Linux, Windows, and Wine.

Screenshot from 2026-06-02 20-57-40

### DIFF
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -4003,7 +4003,14 @@ static void processConsoleCommand(Console* console)
 
     if(commandSize)
     {
+#ifdef BAREMETALPI
         printf("%s", console->input.text);
+#else
+        if (!console->args.cli)
+            printf("\n");
+        else
+            printf("%s", console->input.text);
+#endif
         appendHistory(console, console->input.text);
         processCommand(console, console->input.text);
     }
@@ -4448,6 +4455,31 @@ static void tick(Console* console)
                 exitStudio(console->studio);
         }
     }
+
+#ifndef BAREMETALPI
+    if (!console->args.cli)
+    {
+        static char last_text[CONSOLE_BUFFER_SCREEN] = {0};
+        static s32 last_pos = -1;
+        if (strcmp(last_text, console->input.text) != 0 || last_pos != console->input.pos || console->tickCounter == 1)
+        {
+            strcpy(last_text, console->input.text);
+            last_pos = console->input.pos;
+
+            char dir[TICNAME_MAX];
+            tic_fs_dir(console->fs, dir);
+            printf("\r\033[K");
+            if(strlen(dir))
+                printf("%s", dir);
+            printf(">%s", console->input.text);
+
+            s32 tailLen = strlen(console->input.text) - console->input.pos;
+            if (tailLen > 0)
+                printf("\033[%dD", tailLen);
+            fflush(stdout);
+        }
+    }
+#endif
 
     console->tickCounter++;
 }
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -4006,9 +4006,7 @@ static void processConsoleCommand(Console* console)
 #ifdef BAREMETALPI
         printf("%s", console->input.text);
 #else
-        if (!console->args.cli)
-            printf("\n");
-        else
+        if (console->args.cli)
             printf("%s", console->input.text);
 #endif
         appendHistory(console, console->input.text);
@@ -4548,6 +4546,8 @@ static int apicmp(const void* a, const void* b)
     return strcmp(((const ApiItem*)a)->name, ((const ApiItem*)b)->name);
 }
 
+// Handle terminal input events directly, bypassing the regular SDL event loop for some keys.
+// This allows commands to be entered and executed even when the console is in the background.
 void console_terminal_input(Console* console, tic_key key, char text)
 {
     if(key == tic_key_return)
--- a/src/studio/screens/console.c
+++ b/src/studio/screens/console.c
@@ -4456,31 +4456,6 @@ static void tick(Console* console)
         }
     }
 
-#ifndef BAREMETALPI
-    if (!console->args.cli)
-    {
-        static char last_text[CONSOLE_BUFFER_SCREEN] = {0};
-        static s32 last_pos = -1;
-        if (strcmp(last_text, console->input.text) != 0 || last_pos != console->input.pos || console->tickCounter == 1)
-        {
-            strcpy(last_text, console->input.text);
-            last_pos = console->input.pos;
-
-            char dir[TICNAME_MAX];
-            tic_fs_dir(console->fs, dir);
-            printf("\r\033[K");
-            if(strlen(dir))
-                printf("%s", dir);
-            printf(">%s", console->input.text);
-
-            s32 tailLen = strlen(console->input.text) - console->input.pos;
-            if (tailLen > 0)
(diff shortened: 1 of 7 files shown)


imsys:
I got it to work in a way that we can send commands through the system terminal, even if the SDL window is not in the console screen.

I tried to test the windows binary using wine, and this feature was not working there.

src/system/sdl/main.c
```diff
@@ -34,8 +35,15 @@
...
-#if defined(__TIC_LINUX__)
+#if defined(__TIC_WINDOWS__)
+#include <conio.h>
+#include <io.h>
+#include <fcntl.h>
+#elif defined(__TIC_LINUX__) || defined(__APPLE__) || defined(__TIC_MACOSX__)
 #include <signal.h>
+#include <unistd.h>
+#include <sys/select.h>
+#include <termios.h>
```
imsys, line 46: Windows (`__TIC_WINDOWS__`)
*   `<conio.h>`: Needed for `_kbhit()` (to check if a key is waiting without blocking) and `_getch()` (to read a single character without waiting for Enter).
*   `<io.h>`: Needed for `_open_osfhandle()` and `_dup2()`, which are used to reconnect the standard streams (`stdin`, `stdout`, `stderr`) when the console is hidden or redirected.
*   `<fcntl.h>`: Needed for the `_O_TEXT` constant used during the stream redirection process.

 POSIX (`__TIC_LINUX__`, `__APPLE__`, `__TIC_MACOSX__`)
*   `<unistd.h>`: Provides `read()` for reading from the terminal and the `STDIN_FILENO` constant (the standard input file descriptor).
*   `<sys/select.h>`: Required for the `select()` function and the associated macros (`fd_set`, `FD_ZERO`, `FD_SET`). This is used to poll the terminal for input without blocking the entire game loop.
*   `<termios.h>`: Provides the `tcgetattr`, `tcsetattr`, and terminal mode constants (`ICANON`, `ECHO`, `TCSANOW`). These are used to switch the terminal into "Raw Mode" so TIC-80 can capture keystrokes immediately (like arrow keys) without waiting for the user to press Enter.

src/studio/screens/console.c
```diff
@@ -4003,7 +4003,9 @@ static void processConsoleCommand(Console* console)
 
     if(commandSize)
     {
+#ifdef BAREMETALPI
         printf("%s", console->input.text);
+#endif
```
imsys, line 4008: As this proposed request makes a sync between tic80 console and the system console, we have to remove the output here, otherwise we get a duplicated command output.

src/studio/screens/console.c
```diff
@@ -4402,6 +4424,8 @@ static void tick(Console* console)
         else printBack(console, "\n loading cart...");
     }
 
+    processKeyboard(console);
```
imsys, line 4427: I had to move it down here because it was causing an extra `>` at the startup:

```

 TIC-80 tiny computer
 version 1.2.3101-dev (240d0a1)
 https://tic80.com (C) 2017-2026
>
 hello! type help for help

>
```

src/system/sdl/main.c
```diff
@@ -339,7 +347,7 @@ static const u8* getSpritePtr(const tic_tile* tiles, s32 x, s32 y)
     return tiles[x / TIC_SPRITESIZE + y / TIC_SPRITESIZE * SheetCols].data;
 }
 
-static u8 getSpritePixel(const tic_tile* tiles, s32 x, s32 y)
+static u8 getSpritePixelLoc(const tic_tile* tiles, s32 x, s32 y)
```
imsys, line 350: I had to rename it because it started to cause a conflict with the one in `studio/studio.c`

```
src/system/sdl/main.c:348:11: error: conflicting types for ‘getSpritePixel’; have ‘u8(const tic_tile *, s32,  s32)’ {aka ‘unsigned char(const tic_tile *, int,  int)’}
  348 | static u8 getSpritePixel(const tic_tile* tiles, s32 x, s32 y)
      |           ^~~~~~~~~~~~~~
In file included from src/system/sdl/main.c:24:
src/studio/studio.h:258:4: note: previous declaration of ‘getSpritePixel’ with type ‘u8(tic_tile *, s32,  s32)’ {aka ‘unsigned char(tic_tile *, int,  int)’}
  258 | u8 getSpritePixel(tic_tile* tiles, s32 x, s32 y);
      |    ^~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/tic80.dir/build.make:79: CMakeFiles/tic80.dir/src/system/sdl/main.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1064: CMakeFiles/tic80.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

src/system/sdl/main.c
```diff
@@ -410,7 +418,7 @@ static void drawKeyboardLabels(tic_mem* tic, s32 shift)
     }
 }
 
-static void map2ram(tic_mem* tic)
+static void map2ramLoc(tic_mem* tic)
```
imsys, line 421: I had to rename it because it started to cause a conflict with the one in `studio/studio.c` on the Android and Nintendo Switch builds.

```
/__w/TIC-80/TIC-80/src/system/sdl/main.c:419:13: error: conflicting types for 'map2ram'; have 'void(tic_mem *)'
  419 | static void map2ram(tic_mem* tic)
      |             ^~~~~~~
In file included from /__w/TIC-80/TIC-80/src/system/sdl/main.c:24:
[ 98%] Building C object CMakeFiles/tic80.dir/src/system/nswitch/runtime.o
/__w/TIC-80/TIC-80/src/studio/studio.h:302:6: note: previous declaration of 'map2ram' with type 'void(tic_ram *, const tic_map *)'
  302 | void map2ram(tic_ram* ram, const tic_map* src);
      |      ^~~~~~~
```

imsys:
The Windows version was working quite nice, but I added extra code to get it properly work on Wine. It is desirable to it work nice on Wine for consistency and to be easier to test while on Linux.

imsys:
I also noticed that on Windows, all versions of TIC-80 seems to popup the terminal for a few milliseconds, and then it disappears when double clicking on the binary. It seems this is due to the way it is compiled.
It seems that it is being compiled as `-mconsole` (gcc) / `/SUBSYSTEM:CONSOLE` (MSVC), default for `main()` and `wmain()`. The alternative would be: `-mwindows` / `/SUBSYSTEM:WINDOWS`, default for `WINAPI WinMain()` and `WINAPI wWinMain()`.

I don't really work on windows and I don't really want to set a windows dev environment for now. And it is a bit out of scope for this PR.

Anyway, to test this PR on windows, we can just open CMD and start tic80 from there. (drag and drop to the terminal windows.)

nesbox:
Read through the whole thing; the shape works. Two things before it can merge, both in the terminal side: `tcgetattr`/`tcsetattr` run without an `isatty(STDIN_FILENO)` guard, so a GUI launch with no terminal runs with an uninitialised `termios`; and the restore sits only on `main`'s normal return path, so a crash or a signal leaves the launching shell in raw mode with no echo (the `stty sane` kind of surprise). `isatty` plus `atexit`/signal handlers would cover both.

PR #2947, opened 2026-05-31
Forth

luginf:
adding Forth langages to TIC-80 (using pforth)


luginf:
(made with the help of Claude code)

cmake/forth.cmake
```diff
@@ -0,0 +1,139 @@
...
+    else()
+        set(_PFORTH_DICDAT_BUNDLED "${CMAKE_SOURCE_DIR}/cmake/pfdicdat.h")
+    endif()
+
+    if(NOT EXISTS ${PFORTH_DICDAT})
+        if(EXISTS ${_PFORTH_DICDAT_BUNDLED})
+            message(STATUS "Forth: copying bundled ${_PFORTH_DICDAT_BUNDLED} to ${PFORTH_DICDAT}")
+            configure_file(${_PFORTH_DICDAT_BUNDLED} ${PFORTH_DICDAT} COPYONLY)
+        else()
+            message(STATUS "Forth: pfdicdat.h not found — bootstrapping pforth to generate it...")
```
MineRobber9000: If we can use pforth to bootstrap the `pfdicdat.h` generation, why are we bundling a version of this data anyways? What happens if pforth updates in a way that adds more stuff to the dictionary, and now our bundled data is out of date?

demos/forthdemo.fth
```diff
@@ -0,0 +1,148 @@
+\ title:  Hello Forth
+\ author: TIC-80
+\ desc:   Forth demo - change the word called in TIC to switch example
+\ license: MIT License
+\ script: forth
+\
+\ -- TIC-80 Forth API --------------------------------------------------
```
MineRobber9000: You don't need to leave this API reference in the code.

demos/forthdemo.fth
```diff
@@ -0,0 +1,148 @@
...
+  S" RECT"  10 68 at
+  S" RECTB" 90 68 at
+  S" CIRC"  36 118 at
+  S" ELLI"  145 118 at
+  S" TRI"   24 78 at
+  S" TRIB"  100 78 at
+;
+
+\ -- Demo 2 : sprite + arrow keys -------------------------------------
+: demo02
```
MineRobber9000: This is the only one you actually need to have. The others are all just clutter that a user would have to delete before starting their own game.

ddelemeny:
Note: OP had the courtesy to open a discussion (#2938) regarding the project's AI contribution policy before vibe-coding this payload.

Resolving the policy matter should probably be considered a prerequisite to reviewing this PR.

demos/forthdemo.fth
```diff
@@ -0,0 +1,148 @@
...
+
+\ ( n -- c-addr u )  integer to string
+: n>str  S>D <# #S #> ;
+
+\ ( c-addr u x y -- )  print white at position, scale 1
+: at  15 FALSE 1 FALSE PRINT DROP ;
+
+: BOOT
+  96 px !  24 py !
+  S" Forth BOOT" 12 TRACE
```
MineRobber9000: Remove this, no other language does this. (EDIT: to clarify, since GH shows a seemingly random collection of lines and not the one you explicitly select in the interface, this is referring to tracing "Forth BOOT" in the BOOT callback.)

src/api/forth_io.c
```diff
@@ -0,0 +1,129 @@
+// MIT License
```
MineRobber9000: This should also go in `forth.c`. In general, all languages should be one C file apiece; the only exception are the Lua languages (Lua, MoonScript, Yue, and Fennel), since they all share the same plumbing underneath and only differ in how they're initialized.

MineRobber9000 (changes requested):
I don't know enough about pforth to critique the actual implementation of the API, but there are definitely some spots in here that I am concerned about.

cmake/forth.cmake
luginf: @MineRobber9000 that's a good question. I've initially let it generate the ``pfdicdat.h`` and ``pfdicdat_32.h`` from pforth but there was a problem with some targets (nintendo3D and rpi) do I've included those headers but it's not ideal.

It should be fixed now. I've tested the nintendo 3DS build and it's working fine.

image

demos/forthdemo.fth
luginf: sure. I've removed all that, including the API references and additional examples. Do you think it should be better to include such informations in an additional file in the ``demos/`` folder under a new name, or leave this to the wiki, or even somewhere else? I see in the ``demos`` folder some lua samples like fire.lua or tetris.lua so I don't know

demos/forthdemo.fth
luginf: ok. I've just kept the BOOT procedure for initialising the variables, and kept the "demo01" word outside of the TIC loop, so people can just comment the demo and start creating new words.

src/api/forth_io.c
luginf: it has been merged into the forth.c

MineRobber9000:
How did you manage to `chmod +x` the whole repo?

luginf:
> How did you manage to `chmod +x` the whole repo?

oops, long story short I needed to edit some files from another (local) account (on my computer). I can try to revert the 755 to 644 if you wish, but that will trigger a new build for nothing.

MineRobber9000:
You definitely should undo it. I can't review the PR in this state since I'd have to scroll through basically every file in the repo just to see what you actually changed.

luginf:
> You definitely should undo it. I can't review the PR in this state since I'd have to scroll through basically every file in the repo just to see what you actually changed.

Ok, i've fixed it, I'm just working on another problem (wrong/misleading message during a stack underflow) and I'll commit that ASAP

demos/forthdemo.fth
```diff
@@ -0,0 +1,59 @@
...
+  2 BTN IF  px @ 1- px ! THEN
+  3 BTN IF  px @ 1+ px ! THEN
+  13 CLS
+  1 px @ py @ 14 3 0 0 2 2 SPR
+  S" Hello Forth!" 84 84 15 FALSE 1 FALSE PRINT DROP
+;
+
+: TIC
+demo01
+;
```
MineRobber9000:
```suggestion
: TIC
  0 BTN IF  py @ 1- py ! THEN
  1 BTN IF  py @ 1+ py ! THEN
  2 BTN IF  px @ 1- px ! THEN
  3 BTN IF  px @ 1+ px ! THEN
  13 CLS
  1 px @ py @ 14 3 0 0 2 2 SPR
  S" Hello Forth!" 84 84 15 FALSE 1 FALSE PRINT DROP
;
```
If the user wants to write their own code, they'll probably want to just delete the whole demo code, so having it be a separate word is just clutter.

src/api/forth.c
```diff
@@ -0,0 +1,1260 @@
...
+
+// =============================================================================
+// =============================================================================
+
+static const u8 DemoRom[] =
+{
+    #include "../build/assets/forthdemo.tic.dat"
+};
+
+static const u8 MarkRom[] = { 0 };
```
MineRobber9000: You should also provide a version of the bunny benchmark (see `demos/bunny`).

MineRobber9000 (changes requested):

MineRobber9000:
Also, you forgot to install `gcc-multilib` in the Android task, which is why that build is failing.

src/api/forth.c
luginf: it's working. Yet forth is not an object language so it has been limited to 6800 bunnies, which should be good enough for benchmarking. The absolute limite seems to be around 8000 bunnies, but it would need to swap with the tiles bank.

image

MineRobber9000: You need to follow the same steps as you did with your demo cart here (add it to `assets.bat`, `#include` it here, give the right size in `.mark`).
```suggestion
static const u8 MarkRom[] =
{
    #include "../build/assets/forthmark.tic.dat"
};
```

src/api/forth.c
```diff
@@ -0,0 +1,1260 @@
...
+    .blockEnd           = NULL,
+
+    .keywords         = ForthKeywords,
+    .keywordsCount    = COUNT_OF(ForthKeywords),
+
+    .api_keywords     = ForthAPIKeywords,
+    .api_keywordsCount = COUNT_OF(ForthAPIKeywords),
+
+    .demo = { DemoRom, sizeof DemoRom, "forthdemo.tic" },
+    .mark = { MarkRom, 0, "forthmark.tic" },
```
MineRobber9000:
```suggestion
    .mark = { MarkRom, sizeof MarkRom, "forthmark.tic" },
```

assets.bat
MineRobber9000, line 31:
```suggestion
build\bin\prj2cart demos\bunny\forthmark.fth build\forthmark.tic
build\bin\prj2cart demos\bunny\janetmark.janet build\janetmark.tic
```

assets.bat
```diff
@@ -59,6 +60,7 @@ build\bin\bin2txt build\schemedemo.tic build\assets\schemedemo.tic.dat -z
 build\bin\bin2txt build\sfx.tic build\assets\sfx.tic.dat -z
 build\bin\bin2txt build\squirreldemo.tic build\assets\squirreldemo.tic.dat -z
 build\bin\bin2txt build\tetris.tic build\assets\tetris.tic.dat -z
+build\bin\bin2txt build\forthdemo.tic build\assets\forthdemo.tic.dat -z
 build\bin\bin2txt build\wrendemo.tic build\assets\wrendemo.tic.dat -z
 build\bin\bin2txt build\janetmark.tic build\assets\janetmark.tic.dat -z
```
MineRobber9000, line 67:
```suggestion
build\bin\bin2txt build\forthmark.tic build\assets\forthmark.tic.dat -z
build\bin\bin2txt build\janetmark.tic build\assets\janetmark.tic.dat -z
```

MineRobber9000 (changes requested):

src/api/forth.c
luginf: @MineRobber9000 ok, I've included all this, thank you.

MineRobber9000 (approved):
LGTM!

nesbox:
Merged `main` in (MiniScript landed first). The conflicts were adjacency again: the language list in `script.c`, the sources in `core.cmake`, the includes in `CMakeLists` and the submodule list; both sides kept. My first pass nested the two `script.c` blocks inside one another and the compiler read that as an unterminated conditional — the file is `main`'s again with the Forth blocks beside MiniScript's in both the extern list and the script array, and a native build of the merge links.\n\nOn the CI side: the release lists are not touched here either, and that is ours to grow — `forth` and `miniscript` in one pass, before the next release, as the owner asked. Your two `build.yml` changes look right and unrelated to languages: `gcc-multilib` for the nintendo host toolchain, and `--parallel 3` on the macOS release job, which is that runner's core count.

nesbox (approved):
Read the binding and the wiring. The runtime arrives as a submodule of upstream pforth, `cmake/forth.cmake` names the kernel sources and takes the no-IO build with TIC-80's own shim, and the script registers as `forth` with `.fth` — the release lists are ours to grow, as with MiniScript. Worth saying plainly: the tick has no time bound, so a runaway Forth loop hangs the client exactly as a runaway Lua or JS loop does; that is the house behaviour and MiniScript is the exception, not Forth the deficiency. The two `build.yml` changes are unrelated to languages and both check out.

luginf:
@nesbox thank you very much, and also to @MineRobber9000 for the assistance!

PR #2948, opened 2026-06-01
Update SECURITY.md to remove specific versions

pixelbath:


pixelbath:
Rather than having to remember to update this file every new version, this change removes the versions table entirely.

nesbox (approved):
Reads right: the paragraph above already states the policy, while the table still called 1.0.x-dev alpha and 0.90.x stable — two releases stale, and nothing links to it. Thanks, merging.

Issue #1220, opened 2020-10-26
In HTML exported games, all unused buttons on Xbox One controller open/close the TIC menu


jf1x:
Is there a way to disable this behaviour, or maybe bind this function just to the Menu button on the controller? Right now it's very easy to accidentally hit one of these buttons while playing the game and restart or close it.

nesbox:
There is no way, we need to fix it.

nesbox:
will be fixed with #210

nesbox:
fixed here 9ec7eb7

Issue #2937, opened 2026-05-05
ESC key mapping and the Menu.


imsys:
I’d like to revisit this discussion.

I’m aware this was discussed and decided #1950 back in June 2022. (@blinry) While that solution made sense at the time, it’s starting to become a hindrance in some cases, especially outside of desktop environments.

The main problem today affects non-desktop platforms and controller-based usage.

Note: Edited a bit with AI for clarity, due to non-native English.

---

#### Android

On Android, the system back button is mapped to ESC.
Currently, this opens the console, which is not appropriate for a player-focused environment.

This behavior is confusing and disruptive for users who are simply trying to play.

Relevant code:
[https://github.com/nesbox/TIC-80/blob/9b79d496f0e3afcdf43bc5f97f59834114447b63/src/system/sdl/main.c#L1075-L1077](https://github.com/nesbox/TIC-80/blob/9b79d496f0e3afcdf43bc5f97f59834114447b63/src/system/sdl/main.c#L1075-L1077)

Possible approaches:

* Change ESC behavior globaly to open the menu instead of the console
* Decouple the Android back button (`SDLK_AC_BACK`) from ESC and map it directly to the menu. This probably has the least amount of side effects from future key-mapping changes.

Building without console/studio (`BUILD_EDITORS=OFF`) is not a complete solution, since Android can also be used as a desktop-like development environment. ([casualty](https://github.com/nesbox/TIC-80/pull/2889#issuecomment-4312035012) @sandio)

---

#### Linux handhelds (arm64) and controllers

On the [many](https://handhelds.wiki/Category:Linux) handheld Linux devices, the `Select` button (`SDL_CONTROLLER_BUTTON_BACK`) is mapped to ESC.

This leads to the same issue: pressing a controller button opens the console instead of a menu.

Relevant code:
[https://github.com/nesbox/TIC-80/blob/9b79d496f0e3afcdf43bc5f97f59834114447b63/src/system/sdl/main.c#L1010-L1017](https://github.com/nesbox/TIC-80/blob/9b79d496f0e3afcdf43bc5f97f59834114447b63/src/system/sdl/main.c#L1010-L1017)

This likely affects other devices as well (e.g. Steam Deck).

At minimum, mapping `SDL_CONTROLLER_BUTTON_BACK` to the menu instead of ESC would better match user expectations. Opening a console from a controller is unlikely wanted.

---

### Desktop behavior

[...]

PR #2952, opened 2026-06-09
Add configurable startup screen + improve gamepad menu access + Android build with studio

imsys:
### Problem

Some devices do not have convenient keyboard input. When using only a gamepad, starting directly in the Console can make navigation less intuitive and requires additional steps to reach the Menu.

### Proposal

This PR introduces a configurable startup screen and improves gamepad navigation.

#### Startup screen

Allow the default startup screen to be configured: (between Console and Menu)

* At build time (useful for Android and console ports).
* Through the Options menu (useful for Steam Deck and users who prefer starting in the Menu).

#### Gamepad menu access

Previously, the gamepad **Select** button was mapped to the same action as the **Esc** key, which opened the Console.

This PR decouples the **Select** button from the **Esc** key behavior and makes it open the **Menu** directly instead.

* **Select** opens the Menu from anywhere in the application.
* When already inside a submenu, **Select** returns to the previous menu level.

#### Minor changes

* Android now defaults to fullscreen mode.


imsys:
This decoupling of the Select button from the Esc should solve most of the dispute mentioned in #2937.

nesbox:
Requesting changes. The gamepad half is welcome as it is — it removes the `__SWITCH__` hack around the Back button and closes #1220.

The startup screen wants to be an experimental build flag rather than a runtime option:

- `DEFAULT_STARTUP_SCREEN`, **off by default** — a default build should behave exactly as it does today;
- the runtime side (the config entry and the menu item) behind the same flag — otherwise a default build shows a switch for a feature it does not have;
- out of CI: `build.yml` turns it on for the android build, and an experimental flag should not be switched on there. Whoever wants it builds with the flag.

And a rebase — the branch conflicts with main.

PR #2956, opened 2026-06-10
AndroidManifest -> Set to Game category, and fix `.tic` association on Android 12+

imsys:
Adds association to .tic files to newer Android versions. (12+)

The `.tic` association was working on my older phone, so this is probably a fix for just the newer versions.

The `.png` is still working fine.

And I have set the App to be in the Game category, so some launchers (like Daijisho) organize it in a Game section.

### DIFF
--- a/build/android/app/src/main/AndroidManifest.xml
+++ b/build/android/app/src/main/AndroidManifest.xml
@@ -66,6 +66,7 @@
         android:icon="@mipmap/ic_launcher"
         android:allowBackup="true"
         android:theme="@android:style/Theme.NoTitleBar.Fullscreen"
+        android:isGame="true"
         android:hardwareAccelerated="true" >
 
         <meta-data android:name="android.app.category" android:value="game" />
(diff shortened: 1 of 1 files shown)


joshgoebel:
I don't know about correctness, but looks reasonable.

nesbox:
Read through the manifest: `pathSuffix` is the form Android 12+ honours, and `exported="true"` was already there. Two notes, neither blocking: `intent.category.GAME` and the `android.app.category` meta-data are not read by AOSP (`android:isGame` is the attribute that is), and the branch was refreshed onto `main` so the run is current.

nesbox (approved):
Read the manifest against the platform rules: `pathSuffix` is the form Android 12+ honours, `exported="true"` is already on the activity, and `appCategory` is API 26+ while minSdk here is 24, so older devices ignore it. The `.tic` association is the fix the report was about; the two category extras are launcher folklore, harmless as written.

Issue #2940, opened 2026-05-11
Cursor position is miscalculated when tab size is greater than 1 and indentation is hidden by horizontal scroll


Kevadroz:
To reproduce:

1. Set tab size to at least 2
2. indent a line with some tabs
3. scroll horizontally until the indentation is no longer visible.
4. click on the line

https://github.com/user-attachments/assets/5819f72a-07a3-4116-9946-f9a1d11a6301

This is an oversigth in #2901 from fixing #2330.

Ricarinium:
for me this also happens, the more far away from the indentation mark the more misaligned

nesbox:
Fixed on `main` by e4c65572 and verified on the 1.3.3104-dev build, so I am closing this.

PR #2957, opened 2026-06-11
Fix code editor tab width after horizontal scroll

RDeckard:
## Why

Original issue: https://github.com/nesbox/TIC-80/issues/2940

Follow-up to https://github.com/nesbox/TIC-80/pull/2901

The previous tab-aware mouse mapping fix (#2901) handled visible tabs, but `#2940` shows a remaining mismatch after horizontal scrolling hides indentation tabs. In that case, hidden tabs could stop advancing the rendered x position by their full visual width, while mouse mapping still used visual tab columns.

## What

- Reuse the existing tab-column width calculation for code rendering.
- Compute a tab's x advance before the visibility check, so offscreen tabs still affect later character positions.
- Make `drawTab(...)` draw using the already computed visual width instead of recalculating from screen x.

## Impact

This should make rendering and mouse hit testing agree better when tabs are scrolled offscreen horizontally. I have not manually reproduced the exact video case, so reviewer confirmation on the `#2940` scenario would be useful.

### DIFF
--- a/src/studio/editors/code.c
+++ b/src/studio/editors/code.c
@@ -254,20 +254,23 @@ static inline void drawChar(tic_mem* tic, char symbol, s32 x, s32 y, u8 color, b
     tic_api_print(tic, (char[]){symbol, '\0'}, x, y, color, true, 1, alt);
 }
 
-static int drawTab(Code* code, s32 x, s32 y, u8 color)
+static s32 getTabColumnWidth(Code* code, s32 column)
+{
+    s32 tabSize = getConfig(code->studio)->options.tabSize;
+    if(tabSize <= 0) tabSize = 1;
+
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Checked the bases rather than the prose: the mouse column is `(mx - rect.x)/getFontWidth + scroll.x` (`code.c:3190-3194`) and the render's new column is `(x - xStart)/getFontWidth` with `xStart = rect.x - scroll.x*getFontWidth` — the same number, so both sides now size a tab with `getTabColumnWidth` on the same column. The old width came from the screen x (margin and scroll included), and a tab scrolled off the left edge advanced one cell, which is the #2940 shift. Thanks, merging.

PR #2958, opened 2026-06-11
Dynamic controller assignment

### LINKED ISSUE #2955: Player 1 is not necessarily the first controller in use
## Problem

Currently, TIC-80 appears to assign controller slots (Player 1, Player 2, Player 3, etc.) according to the order reported by SDL.

In practice, SDL device ordering is not always intuitive. Users may have disconnected controllers, virtual devices, adapters, or stale device entries that cause their active controller to be assigned as Player 3 or Player 4 instead of Player 1.

As a result, games may not respond to the controller the user expects to control Player 1.

## Proposed solution

Instead of permanently relying on SDL's controller order, assign player slots according to the order in which controllers are first used after TIC-80 starts.

Example:

1. Controller B presses a button first → assigned to Player 1.
2. Controller D presses a button second → assigned to Player 2.
3. Controller A presses a button third → assigned to Player 3.

The mapping would be reset each time TIC-80 is launched.

----

I ran into this issue because my controller only opens the menu, which is mapped globally and works regardless of the assigned player slot.

However, menu navigation and in-game controls do not work. After debugging, I found that the controller is being assigned as Player 3. In my case, SDL appears to be reporting two stale devices ahead of the controller I am actually using.

This could be one of the reasons some users reported that their controller is not detected, even though it is detected and simply assigned to a different player slot.

I can submit a PR for this.

imsys:
This was mainly coded by Gemini, I mostly reviewed and tested it.

This fixes #2955. I tested with [cart 893](https://tic80.com/play?cart=893), note that this test cart is positioned as:
| P1 | P3 |
|----|----|
| P2 | P4 |

Before this patch, when I try to play on my PC with a Xbox controller, it is assigned to P3. If I plug another controller, it gets assigned to P4.

With this patch I also ested that:
The first player to that moves or press any button is the 1st player. If I turn it off and on again, it also keeps being the 1st player.

This will probably have a merge conflict with #2952

Following is the comments by Gemini:

### Key Changes:
Increased Controller Support: Expanded the internal controller ports from 4 to 16. This ensures that even if you have multiple "ghost" devices or adapters at low SDL indices, your active controller will still be detected and opened.

Assignment by First Use: Introduced a mapping system that tracks the registration order of controllers. A controller is only assigned to a player slot (1-4) once it shows some activity (button press or stick movement).

Active Removal Handling: Corrected the hotplug logic to properly identify which controller is being disconnected using its unique SDL Instance ID, ensuring assigned slots are freed if a controller is unplugged.

Robust Mapping: Initialized all player slots as unassigned at startup. The first controller to send input becomes Player 1, the second becomes Player 2, and so on.

### Details:
- Added SDL_JoystickID mapping[TIC_GAMEPADS] to the platform.gamepad structure.
- Implemented isControllerActive() to detect button presses or axis movements exceeding a deadzone threshold.
- Updated processGamepad() to perform a two-phase check: first, mapping unassigned active controllers to available slots; then, reading input from the mapped controllers.
- Fixed SDL_CONTROLLERDEVICEREMOVED to use Instance IDs instead of indices, which is the correct behavior for this SDL event.

### DIFF
--- a/src/system/sdl/main.c
+++ b/src/system/sdl/main.c
@@ -119,7 +119,8 @@ static struct
 
     struct
     {
-        SDL_GameController* ports[TIC_GAMEPADS];
+        SDL_GameController* ports[16];
+        SDL_JoystickID mapping[TIC_GAMEPADS];
 
 #if defined(TOUCH_INPUT_SUPPORT)
         struct
@@ -947,11 +948,23 @@ static u8 getButton(SDL_GameController* controller, SDL_GameControllerButton but
         : 0;
 }
 
+static bool isControllerActive(SDL_GameController* controller)
+{
+    for(s32 i = 0; i < SDL_CONTROLLER_BUTTON_MAX; i++)
+        if(SDL_GameControllerGetButton(controller, i))
+            return true;
+
+    for(s32 i = 0; i < SDL_CONTROLLER_AXIS_MAX; i++)
+        if(abs(SDL_GameControllerGetAxis(controller, i)) > AXIS_THRESHOLD)
+            return true;
+
+    return false;
+}
+
 static void processGamepad()
 {
     {
         platform.gamepad.joystick.data = 0;
-        s32 index = 0;
 
         for(s32 i = 0; i < COUNT_OF(platform.gamepad.ports); i++)
         {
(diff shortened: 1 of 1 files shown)


nesbox:
Traced the assignment: `mapping[]` is cleared to `-1` in `start()`, a slot is claimed only by a controller that is actually active, and removal frees both the handle and the slot. Sound. About the #2952 conflict you mention: the refreshed branch merges cleanly for now.

nesbox (approved):
Traced the assignment: `mapping[]` is cleared to `-1` in `start()`, a slot is claimed only by a controller that is actually active, and removal frees both the handle and the slot that used it. Sound.

Issue #2768, opened 2025-03-31
keyboard input mapping broken on linux nightly


khimaros:
with the input tester i'm seeing:

left = right
right = caps
down = left
up = down
ctrl = shift
space = tab

these are pressed key = detected key.

oddly, the keys in the `menu` ui seem fine.

i tested both gcc12 and gcc14 variants.

Miguel-hrvs:
you're right, this is weird

PR #2962, opened 2026-06-27
Fix #2773: space key broken with KEYBOARD_LAYOUT_ES after tic_key_iso_extra removal

### LINKED ISSUE #2773: Spaces don't work in baremetal builds on pi4
## Reproduction
1. Build for Raspberry Pi 4
2. Load image on to Pi
3. Plug USB keyboard in
4. Power on Pi
5. Once terminal loads, hit space bar. No spaces are added to the terminal
6. Hold ALT+Space. Spaces are added

## Potential Fix
I was playing around with this, and it seems to be an issue with the `Symbols` array in the `studio.c` file.

When the space key comes through from I'm guessing circle, the value isn't the space in the symbol string:
```
" abcdefghijklmnopqrstuvwxyz0123456789-=[]\\;'`,./ "
```
I believe it is probably getting a null terminator at the end instead of the last space. I was able to fix this by adding an extra space to that string. Since SDL builds skip that part of the keyboard function, all SDL builds will work (I tried on linux and it does). 

Looking at the effected code, it seems like this has been an issue for awhile? Not sure if anyone else noticed it, or if maybe there was a recent update to Circle that wasn't checked fully? But if this fix can work, I can make a PR.

joshgoebel:
Closes #2773

Commit 1149e7cd (#2768) fixed baremetal space by removing `tic_key_iso_extra`, shifting `tic_key_space` to index 48. Non-ES builds were automatically fixed (the trailing `' '` was already at index 48). But `KEYBOARD_LAYOUT_ES` arrays still had `<`/`>` at position 48 (the old ISO extra key slot), so ES builds got `<` for space and `>` for shift+space. ALT+Space still worked (explaining the workaround described in the issue).

**Fix:** Remove `<` from `Symbols[48]` and `>` from `Shift[48]` in the ES layout so `tic_key_space (48)` correctly maps to `' '` across all layouts. Also trims `Alt` by one character for consistency and updates the reference comment.

### DIFF
--- a/src/studio/studio.c
+++ b/src/studio/studio.c
@@ -432,10 +432,10 @@ char getKeyboardText(Studio* studio)
         tic80_input* input = &tic->ram->input;
 
 #ifdef KEYBOARD_LAYOUT_ES
-        // US KEYS:                     " abcdefghijklmnopqrstuvwxyz0123456789-=[]\\;'`,./< ";
-        static const char Symbols[] =   " abcdefghijklmnopqrstuvwxyz0123456789'!`+cn'o,.-< ";
-        static const char Shift[] =     " ABCDEFGHIJKLMNOPQRSTUVWXYZ=!\" $%&/()??^*CN\"a;:_> ";
-        static const char Alt[] =       "                            |@#        []} {\\     ";
+        // US KEYS:                     " abcdefghijklmnopqrstuvwxyz0123456789-=[]\\;'`,./ ";
+        static const char Symbols[] =   " abcdefghijklmnopqrstuvwxyz0123456789'!`+cn'o,.- ";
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Verified against the enum rather than the description: `tic_key_space` is 48 (`src/tic.h`), and the ES arrays held `<`/`>` at that index while `Alt[48]` already held a space — which is exactly why ALT+Space worked in #2773. With this change all three ES arrays map 48 to a space, matching the non-ES layout; ES builds also stop inserting a space on TAB (index 49 was the trailing space in the longer string) and return NUL the way the US layout already does. The root cause in the body checks out (1149e7cd, #2768). Thanks — merging.

PR #2963, opened 2026-06-27
Fix #2920: Python sfx() ignores note/octave/speed from SFX editor

### LINKED ISSUE #2920: sfx()'s note property isn't defaulted to what is set in SFX tab in Python binding
Whenever you call sfx() in a python project without directly specifying the note you want to play it with, it defaualts to the lowest note (C-1), not the note this sfx should play at. For example, on my sfx tab the note of sfx 0 is supposed to be C-4

Image

But if I call it with sfx(0), it will play on note C-1. In order to play it on C-4, I have to call sfx(0, "C-4") explicitly. In all other interpreters calling without specifying the note seems to work fine, the problem is only on python as far as I can see (I tested js, lua, moon, wren). As far as binary versions I have this problem on the latest official version as well as the latest unofficial build by someone in the comunity (I think it's from August-ish 2025), I am using the Windows version, and I don't see any issues here that have to do with this problem so it probably still exists on the main branch.

joshgoebel:
Closes #2920

## What

The Python `sfx()` binding never read the stored note, octave, or speed from the SFX sample data. When called as `sfx(0)` (no note specified), it computed `note = -1 % 12 = -1` and `octave = -1 / 12 = 0`, causing playback at C-1 regardless of what's set in the SFX tab. All other language bindings (Lua, JS, Wren, etc.) work correctly because they read from the sample data first.

## Fix

Mirrors the Lua binding: read `effect->note`, `effect->octave`, and `effect->speed` from `tic->ram->sfx.samples.data` as defaults before applying any user-provided overrides.

Also changes `speed`'s default sentinel from `0` to `-1` (matching `note`/`duration`) so the effect's stored speed is used when not explicitly specified.

### DIFF
--- a/src/api/python.c
+++ b/src/api/python.c
@@ -824,7 +824,7 @@ static bool py_reset(int argc, py_Ref argv)
     return true;
 }
 
-// sfx(id: int, note=-1, duration=-1, channel=0, volume=15, speed=0)
+// sfx(id: int, note=-1, duration=-1, channel=0, volume=15, speed=-1)
 // void (*sfx)(tic_mem*, s32, s32, s32, s32, s32, s32, s32, s32)
 static bool py_sfx(int argc, py_Ref argv)
 {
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Checked this against the code rather than the description, and the old path is confirmed: the binding's default `note = -1` went through `-1 % 12` and `-1 / 12` to `channel->note = -1`, and `sfx()` in `src/core/sound.c` clamps that to index 0, the bottom of `NoteFreqs` — the C-1 that #2920 reports. Python was also the last binding left: js, luaapi, wren, mruby, janet and squirrel all read `effect->note` already, and this now follows the JS binding step for step. The speed change checks out as well: `setChannelData` uses a speed that fits the 3-bit field as is and falls back to the sample's own speed otherwise, so the old default of 0 was used directly and ignored the SFX tab, while -1 now resolves to `effect->speed`. Index checks running before the sample read is a good catch too. One follow-up, not a blocker for this one: the Scheme binding (`src/api/scheme.c`, `scheme_sfx`) has the same missing read, so `(sfx 0)` there still plays the lowest note. Thanks — merging.

PR #2964, opened 2026-06-27
Fix #1558: world map preview ignores map editor bpp setting

### LINKED ISSUE #1558: World map doesn't respect current bpp on map editor
World map still use 4bpp to get colors even if I currently editing in a different bpp, leading to completely inconsistent colors.
image

joshgoebel:
Closes #1558

## What

The world map overview (**Tab** in the map editor) always rendered using 4bpp regardless of the bpp mode selected in the map editor toolbar. This caused incorrect colors for cartridges using 2bpp or 1bpp tiles.

## Root cause

`initWorld()` used `tic_tool_peek4()` to read every tile's pixel data, hardcoding 4bpp interpretation:

```c
tic_tile* tile = &getBankTiles(world->studio)->data[index];
for(s32 p = 0; p < TIC_SPRITESIZE * TIC_SPRITESIZE; p++)
{
    u8 color = tic_tool_peek4(tile, p);  // always 4bpp
    ...
}
```

## Fix

Use `tic_tilesheet_get` with the segment derived from `map->sheet.blit` (the same blit state the map editor uses), then read pixels via `tic_tilesheet_gettilepix` which dispatches to the correct `peek1`/`peek2`/`peek4` function for the active bpp mode. This mirrors the approach already used by the sprite and map editors.

### DIFF
--- a/src/studio/editors/world.c
+++ b/src/studio/editors/world.c
@@ -125,6 +125,9 @@ void initWorld(World* world, Studio* studio, Map* map)
     memset(world->preview, 0, PREVIEW_SIZE);
     s32 colors[TIC_PALETTE_SIZE];
 
+    u8 segment = tic_blit_calc_segment(&map->sheet.blit);
+    tic_tilesheet sheet = tic_tilesheet_get(segment, (u8*)getBankTiles(world->studio));
+
     for(s32 i = 0; i < TIC80_WIDTH * TIC80_HEIGHT; i++)
     {
         u8 index = getBankMap(world->studio)->data[i];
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Checked this against the code rather than the description. The preview now asks its question with the same code the map itself is drawn with: `tic_tilesheet_gettile(&sheet, index, true)` is the call `drawMap` makes for a map cell (`src/core/draw.c:291`), and the segment comes from `map->sheet.blit` — the state `tic_blit_update_bpp` writes when the toolbar toggle moves (`src/studio/editors/map.c:270`) and the same one the map editor publishes to `vram.blit.segment` (`map.c:435`). Pixels come from `tic_tilesheet_gettilepix`, the accessor `drawTile` reads its own with, so the peek1/peek2/peek4 dispatch is the core's and not a copy of it. The sheet pointer is `getBankTiles(studio)`, the convention `initSprite` is called with (`studio.c:1552`), and the sheet's own bank walks into that bank's `sprites` block, which is the layout that makes segment bank 1 the fg sheet in the core path too. For 4bpp the new read is the same bytes the old one read (`&data[index]` and `x + 8y`), so the default case cannot regress. Nothing this patch depends on has moved since the branch was cut — the files it touches are untouched since its base, and the two commits that did land in studio.c since are the keyboard arrays and the run-origin rework. Thanks — merging.

PR #2966, opened 2026-07-06
Fix Memory Leak in TIC/BOOT/BDR .. Callback Functions When Variable i…

### LINKED ISSUE #2707: Memory Leak in TIC Function When Variable is at the End (Ruby)
There is a memory leak in the TIC function when a variable is placed at the end of the function in Ruby. This can be easily confirmed with large arrays, but it also occurs with simple expressions like $prev_mouse = mouse when placed at the end. The memory leak does not occur if the function explicitly returns nil.
```
# script:  ruby

def TIC
	(0..10000).to_a
end
```

hadamak:
Resolves #2707.

---

AI suggested the fix and @hadamak implemented and tested.

### DIFF
--- a/src/api/mruby.c
+++ b/src/api/mruby.c
@@ -1135,7 +1135,9 @@ static void callMRubyTick(tic_mem* tic)
     {
         if (mrb_respond_to(mrb, mrb_top_self(mrb), mrb_intern_cstr(mrb, TicFunc)))
         {
+            int ai = mrb_gc_arena_save(mrb);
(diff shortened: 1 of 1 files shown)


joshgoebel:
Does this mean it worked? :-)

hadamak:
That's right. As you suggested, the memory leak in question has been resolved, and everything seems to be working fine. Thank you very much.

joshgoebel (approved):

Issue #2792, opened 2025-06-28
Smart syntax highlight


duckafire:
### Down to earth

Tic80 could differentiate when a API function is used like a argument and then it could decide that it do not have highlight it.

It will allow that the users can use these words freely (as properties), no fear of to became the code confuse, desorganized and/or dirty.

![Image](https://github.com/user-attachments/assets/7921fa93-8cc0-4b2a-9eff-0a6e13964f86)

### Simple, but hard (or not)

It will be cool if Tic80 can highlight property, constant and other "group" of identifiers.

It do not know the Tic80 code responsible by syntax highlight, but I think that it is not very hard add a simple RegExp to highlight some patterns, like `\b[A-Z_][A-Z_0-9]*\b` (constants) and `\b0[bB][0-1]*\b` (binary notation).

I know that it looks like with thing that a IDE have to do and it do not a simple fantasy console, but with the right RegExp a lot of things can be to done with the syntax highlight.

Like a Vim user, I know that it is not so hard when it looks like, because it depends only of the code structure of the text redering and its support to RegExp.

> [!NOTE]
> Anyway, see the first topic as the main, the other is a "simples" suggestion. Sorry me by any ortographic error, I do not have a perfect English.

PR #2975, opened 2026-07-27
Implement draw cache for optimized rendering in TIC-80

Wang-Yue:
# TIC-80 Draw Cache System

## Rationale
In TIC-80, the main loop or editor tick function (`tic()`) is called 60 times per second, and by default, every single frame is fully redrawn from scratch, even when the screen contents are completely static (for example, when simply reading code, hovering a stationary mouse, or resting on a menu). 

Furthermore, many games do not change their state or visual representation on every frame, and some games may only need to update and redraw their screen every 5, 10, or more frames. 

Redrawing static frames or slow-running games continuously results in unnecessary and wasteful CPU consumption. The **Draw Cache** is an optimization layer designed to solve this by bypassing expensive, pixel-by-pixel rendering operations (like drawing text characters or rendering tilemaps) when the visual state of the application has not changed from the previous frame.

---

### 1. Interception (Hooking)
All standard drawing APIs called by the TIC-80 studio/editors (e.g. `tic_api_cls`, `tic_api_print`, `tic_api_rect`, etc.) are intercepted in [studio.h](file:///Users/wangyue/TIC-80/src/studio/studio.h#L41-L59) and routed through **caching wrappers** defined in [draw_cache.c](file:///Users/wangyue/TIC-80/src/core/draw_cache.c#L495-L550):

```c
// Example: Intercepting `rect` drawing
static void cache_api_rect(tic_mem* tic, s32 x, s32 y, s32 w, s32 h, u8 color)
{
    tic_core* core = (tic_core*)tic;
    DrawCallRect call = { .type = DRAW_CALL_RECT, .x = x, .y = y, .w = w, .h = h, .color = color };
    
    // If the draw cache says we can skip it, do not call the actual drawing function
    if (handle_draw_call(core, &call, sizeof(call)))
    {
        tic_api_rect(tic, x, y, w, h, color);
    }
}
```

---

### 2. State Validation (Start of Frame)
At the beginning of each frame (`tic_core_draw_cache_start`), the cache checks whether the environment is unchanged compared to the previous frame. It compares:
- **RAM Block A** (VRAM, tiles, sprites, map)
- **RAM Block B** (sprite flags, fonts, button mapping)
- **VBANK1** contents
- **Mouse position and button states**

If **any** of these inputs have changed, the draw cache is immediately **invalidated** for this frame, forcing a full redraw.

---

### 3. Record and Match (`handle_draw_call`)
During the update loop, as drawing commands are called, each command parameters are serialized into a `DrawCall` struct and passed to `handle_draw_call`:

1. **Recording**: The command is appended to the current frame's draw call buffer (`curr_calls_buf`).
2. **Matching**: If the cache is still valid, the new command is compared byte-by-byte with the command at the corresponding position in the *previous* frame's draw call buffer (`prev_calls_buf`).
   - **Match**: `handle_draw_call` returns `false` (skip execution). The drawing function is bypassed entirely.
   - **Mismatch**: The cache is instantly invalidated. `handle_draw_call` returns `true` (execute drawing). All subsequent drawing commands for the rest of the frame will be executed.

---

### 4. Finalization (End of Frame)
At the end of the frame (`tic_core_draw_cache_end`):
- **If the cache remained valid** (all draw calls matched and inputs were identical):
  - No new commands were executed.
  - The final output screen pixels are guaranteed to be identical to the previous frame.
  - The renderer bypasses pushing a new frame to the GPU, saving both CPU and GPU rendering cycles.
- **If the cache was invalidated**:
  - The new drawing commands were executed and recorded into `curr_calls_buf`.
  - The contents of `curr_calls_buf` are swapped to become `prev_calls_buf` for the next frame.
  - The current RAM and input states are saved to act as the reference for the next frame's start-of-frame validation.

---

## Performance Benefits
- **Static Screens**: In terminal or editor (code, music, etc) when no code is typed, CPU/GPU rendering overhead drops to nearly **0%** active time.
- **Games**: Tested 8 bit panda and Cauliflower Power. When the game frame is not changing rapidly (when the actor stand still), we only see low single digit (3-5%) CPU consumption, compared  to ~20% previously using the new Apple backend. 
-  **Highly Dynamic Screens**: Verified rendering. The CPU consumption does not see measurable increase as the software renderer (`draw.c`) is the main bottleneck and compare to that the cache operation is very cheap.

### DIFF
--- a/cmake/core.cmake
+++ b/cmake/core.cmake
@@ -18,6 +18,7 @@ set(TIC80CORE_SRC
     ${TIC80CORE_DIR}/fftdata.c
     ${TIC80CORE_DIR}/core/core.c
     ${TIC80CORE_DIR}/core/draw.c
+    ${TIC80CORE_DIR}/core/draw_cache.c
     ${TIC80CORE_DIR}/core/io.c
     ${TIC80CORE_DIR}/core/sound.c
     ${TIC80CORE_DIR}/tic.c
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -25,6 +25,7 @@
 
 #include "api.h"
 #include "core.h"
+#include "draw_cache.h"
 #include "tilesheet.h"
 
 #include <assert.h>
@@ -484,6 +485,18 @@ void tic_core_tick(tic_mem* tic, tic_tick_data* data)
             config->boot(tic);
             core->state.tick = config->tick;
             core->state.callback = config->callback;
+            
+            if (strlen(tic->cart.code.data) > 0)
+            {
+                core->state.has_scn = strstr(tic->cart.code.data, "SCN") != NULL;
+                core->state.has_bdr = strstr(tic->cart.code.data, "BDR") != NULL;
+            }
+            else
+            {
+                core->state.has_scn = true;
+                core->state.has_bdr = true;
+            }
+
             core->state.initialized = true;
         }
         else return;
@@ -541,12 +554,15 @@ void tic_core_close(tic_mem* memory)
     free(memory->product.screen);
 #endif
     free(memory->product.samples.buffer);
+    tic_core_draw_cache_free(core);
+
     free(core);
 }
 
 void tic_core_tick_start(tic_mem* memory)
 {
     tic_core* core = (tic_core*)memory;
+    tic_core_draw_cache_start(core);
     tic_core_sound_tick_start(memory);
     tic_core_tick_io(memory);
 
@@ -575,6 +591,7 @@ void tic_core_tick_end(tic_mem* memory)
     core->state.gamepads.previous.data = core->state.gamepads.now.data;
 
     tic_core_sound_tick_end(memory);
+    tic_core_draw_cache_end(core);
 }
 
 // copied from SDL2
@@ -643,9 +660,22 @@ static inline u32 blitpix(tic_mem* tic, s32 offset0, s32 offset1, const tic_blit
         : pal0->data[tic_tool_peek4(vbank0(core)->screen.data, offset0)];
 }
 
+static inline void scanline(tic_mem* memory, s32 row, void* data);
+static inline void border(tic_mem* memory, s32 row, void* data);
+
 void tic_core_blit_ex(tic_mem* tic, tic_blit_callback clb)
 {
     tic_core* core = (tic_core*)tic;
+
+    if (core->draw_cache && !tic_core_draw_cache_has_invalidated(core))
+    {
+        bool has_scanline = (clb.scanline == scanline) ? (core->state.initialized && core->state.has_scn) : (clb.scanline != NULL);
+        bool has_border   = (clb.border   == border)   ? (core->state.initialized && core->state.has_bdr) : (clb.border != NULL);
+        if (!has_scanline && !has_border)
+        {
+            return;
+        }
+    }
 
     tic_blitpal pal0, pal1;
     updpal(tic, &pal0, &pal1);
@@ -743,6 +773,8 @@ tic_mem* tic_core_create(s32 samplerate, tic80_pixel_color_format format)
     blip_set_rates(core->blip.left, CLOCKRATE, samplerate);
     blip_set_rates(core->blip.right, CLOCKRATE, samplerate);
 
+    tic_core_draw_cache_init(core);
+
     {
 #define API_FUNC_DEF(name, ...) core->api.name = tic_api_ ## name;
         TIC_API_LIST(API_FUNC_DEF)
@@ -752,6 +784,8 @@ tic_mem* tic_core_create(s32 samplerate, tic80_pixel_color_format format)
         void tic_api_textri(tic_mem* tic, float x1, float y1, float x2, float y2, float x3, float y3, float u1, float v1, float u2, float v2, float u3, float v3, bool use_map, u8* colors, s32 count);
         core->api.textri = tic_api_textri;
 #endif
+
+        tic_core_draw_cache_hook_api(core);
     }
 
     tic_api_reset(&core->memory);
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -165,12 +165,17 @@ typedef struct
         s32 l, t, r, b;
     } clip;
 
+    bool has_scn;
+    bool has_bdr;
     bool initialized;
 } tic_core_state_data;
+
+struct tic_draw_cache;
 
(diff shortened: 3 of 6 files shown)


joshgoebel:
> Record and Match (handle_draw_call)

This (at first blush) sounds overengineered... IF we're already comparing ALL the VRAM, VBANK1, sprites, fonts, etc... shouldn't we have reached certainty about whether the screen could be different or not?  Why do we also need to walk a history of draw commands?  Sounds like a lot of extra complexity.  

_It's also possible the draw calls could have changed while the display didn't change at all._. Yes, edge cases for sure... but... 

Curious what this part of the system is trying to account for?

Wang-Yue:
> IF we're already comparing ALL the VRAM, VBANK1, sprites, fonts, etc... shouldn't we have reached certainty about whether the screen could be different or not? Why do we also need to walk a history of draw commands? Sounds like a lot of extra complexity.

If you profile the application, in the current system, the software renderer `draw.c` is the main bottleneck (accounts for > 70% of the total run time). The GPU rendering part, especially after moving to Metal, is fairly cheap, as it's just moving the software rendered pixel buffers to GPU. Even user program running in byte code accounts for a much smaller part of the run time for most games I tested. Comparing the ram and buffers in every cycle could help you skip the GPU rendering for that frame, but CPU still does all the software rendering at 60fps and you can't skip it unless comparing the command history. 

Moving the software renderer to GPU is infeasible, due to retro games peek and poke all the time. I actually tried to port the entire software renderer to Metal but due to every time it peek/poke I have to copy the buffer back from GPU to CPU, the metal based renderer runs even slower than software one, for almost all games I tested. 

So draw cache aim to solve this bottleneck in the software renderer.
 
>Curious what this part of the system is trying to account for?

the major customer that will enjoy this optimization is all built in editors (code, music, sprite, etc). you will see CPU utilization drop to around 1/25. There are also games don't need 60fps rendering. for instance, my favorite game Cauliflower Power. When you are tackling the puzzle game and the character is standing still, you will see it only redraws 1 frame per second. The various slow games such as puzzle games, large portions of cartridges in [Tech](https://tic80.com/play?cat=1) [Tools](https://tic80.com/play?cat=2) [Music](https://tic80.com/play?cat=3) sections from the official site, will all benefit from this tremendously.

joshgoebel:
> but CPU still does all the software rendering at 60fps

Why does it do this if no VRAM or registers have changed?  Could we just fix the software rendering to be a bit smarter in terms of changes from frame to frame?  If there were truly no changes to display then I'd think the software renderer could essentially be a NOOP, no?

Wang-Yue:
> Why does it do this if no VRAM or registers have changed?

because  VRAM is the final state of the software rendering. In retro console, user function (`TIC()`) will call draw function to redraw every frame. Without draw cache, you have to do the redraw from start to finish in order to find if VRAM changes or not. With draw cache you can compare the drawing command history and tell if we need to redraw or not.

Once at the stage of comparing VRAMS, the software rendering is already finished --- You cannot rewind the clock to skip that step when you find this frame's rendering is not needed.

> Could we just fix the software rendering to be a bit smarter in terms of changes from frame to frame? If there were truly no changes to display then I'd think the software renderer could essentially be a NOOP, no?

The draw cache is fixing exactly this! When nothing needs to be redrawn, all drawing instructions becomes noops.

Wang-Yue:
I recorded a few performance graph for reviewers to understand the benefit of draw cache for quite games and code editing in this comment: https://github.com/nesbox/TIC-80/pull/2973#issuecomment-5099626500 please take a look

I want to post one additional performance graph here. It's recorded when playing stele, a action game where each single frame is different.


Screenshot 2026-07-28 at 12 50 13

so the user program calls `map()` on the lua side, which consumes 20% of the run time. the lua callback calls our `cache_api_map` function, which quickly cause cache miss, then, the cached system invalidated the cache and calls `tic_api_map` for software rendering. 

On the graph you can find `tic_api_map` and `cache_api_map` takes almost exactly same time --- maybe there is only 0.1% difference but it's so hard to see. This proofs that even when every single frame needs to be redrawn, our cache has almost zero overhead, compared to the software rendering cycle.  

the conclusion is: For dynamic screens, the system has non-measurable overhead. For close to static screens, the system saves 10-25X amount of CPU cycles.

nesbox:
Read the mechanism closely: it is sound, and my hesitation is cost against benefit rather than the idea itself. Three things hold it back — `strstr(code, "SCN"/"BDR")` in the core decides whether user callbacks may be skipped, where the bindings already know that fact at call time and the core reads the source text for it; a skipped SCN/BDR can drop side effects, not only pixels; and the cache has no off switch. Its editor-mode win also needs the #2972 half, which is not in `main`. Leaving this open as a direction question. If it comes back, the shape I would argue for first is smaller: skip the blit when the cart produced no new screen content, without the call list and without the replay.

Wang-Yue:
I have addressed all three concerns and rebased this PR on top of pr/skip-clean-screen-blit:                                                                                                                                  
1. No strstr in Core: SCN/BDR scanline detection is now performed directly by language bindings during boot (`vm->has_scn_or_bdr`), completely removing text scanning from the core.                                            
2. Preserves Side Effects: Any cartridge with SCN or BDR defined is flagged and always executes its callbacks without caching to prevent side-effect loss.                                                                    
3. Runtime Off Switch & Zero Memory Overhead:                                                                                                                                                                                 
    • Added a runtime toggle in the Studio Options Menu (DRAW CACHE [ON / OFF]) and persisted it in options.json.                                                                                                             
    • The 4MB command buffers are allocated lazily on demand only when enabled, and freed immediately when toggled off (0 byte RAM overhead when disabled).                                                                   
4. Includes [#2792](https://github.com/nesbox/TIC-80/pull/2972) Fix: The stationary cursor fix is included (via the base branch), ensuring editor modes achieve maximum CPU reduction.

NOTE: https://github.com/nesbox/TIC-80/pull/2974 is not merged in this as that is an independent change. To test performance please also merge that which will benefit CPU utilization in code mode.

nesbox:
Superseded by #2973, which now contains this branch. All three points raised here are addressed in that tree: the source-text `strstr` is gone in favour of `has_scn`/`has_bdr`, and the cache has an off switch — `drawCache` in `config.c`, a menu entry, and `tic_core_draw_cache_set_enabled` in the API. Closing rather than leaving two threads for one mechanism; further discussion belongs on #2973.

PR #2983, opened 2026-08-21
Fix WASM libretro core build failing at runtime

Ferk:
I had to make this change to get the TIC-80 wasm build for libretro to load games successfully, otherwise it fails at runtime.

TIC-80’s API expects callbacks shaped like `u64 callback(void*)`, but the libretro wrapper defined those callbacks as `u64 callback()`.

Wasm indirect calls are type-checked at runtime, so the mismatch traps with indirect call signature mismatch.

### DIFF
--- a/src/system/libretro/tic80_libretro.c
+++ b/src/system/libretro/tic80_libretro.c
@@ -84,8 +84,10 @@ static struct tic80_state* state = NULL;
 /**
  * TIC-80 callback; Request counter.
  */
-static u64 tic80_libretro_counter()
+static u64 tic80_libretro_counter(void *opaque)
 {
+	TIC_UNUSED(opaque);
+
 	if (state == NULL) {
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


joshgoebel:
Did this break somewhere along the way or has it neverr worked?

Ferk:
It probably never worked for this specific edge case, it's only a problem for the WASM build of the libretro core but I expect most people that might be using the Retroarch core would use a native build (which works fine), not the [Web Player](https://web.libretro.com/) (where the error is happening right now), since TIC-80 already has a decent web player itself there's not much reason to use Retroarch's ...and the error only happened at runtime so you have to actually test it to notice it...

I stumbled into it because I'm experimenting with a web player for libretro myself and I wanted to include the TIC-80 core.

nesbox (approved):
Merged main into your branch to clear the conflict: those two callbacks carry the `void*` signature in main now too, because #2981 fixed the same declaration for the C23 build — your wasm trap is the runtime face of the same bug. The merge keeps main's signature and the `TIC_UNUSED` marks from this branch. Thanks, merging.

PR #2988, opened 2026-09-08
Fix attempting to out-of-bounds index Squirrel arrays

### LINKED ISSUE #2987: Problems with Squirrel ttri and others
I've noticed a couple of issues with Squirrel's `ttri` implementation:

1. Reading an array of chroma colours has an out-of-bounds due to (I assume) a copy paste from the lua implementation which is 1-based.  There's no crash when it tries to access index -1, however this means that in the below example `ttri` will still receive two colours - the first is the second element of the passed array argument, and the second will be whatever was in index 1 of `static u8 colors[TIC_PALETTE_SIZE];` the last time `squirrel_ttri` was called.  Edit: this also occurs in `map` and `spr`.
https://github.com/nesbox/TIC-80/blob/4aba09c98f1e5028b82765be1647677b08d35942/src/api/squirrel.c#L520

2. Depth values are read into the `pt` array rather than `z`, which means it overwrites the first 3 values if they're passed.
https://github.com/nesbox/TIC-80/blob/4aba09c98f1e5028b82765be1647677b08d35942/src/api/squirrel.c#L543

Thankfully these are both easy fixes, and I'll submit a PR.

---

Edit: Further to this, I've discovered that attempting to retrieve an out-of-bounds squirrel array index via `sq_rawget` will throw a Squirrel error via `sq_throwerror` that the application is meant to handle.  This means that in cases where we don't verify the size of the incoming Squirrel array, reading the out-of-bounds index will return an error that TIC-80 is completely ignoring.

For example, reading the chromakey array will loop from 0 to `TIC_PALETTE_SIZE`-1 even if the Squirrel array is not that large.  Since it is out of bounds, the call to `sq_rawget` pops the integer index off the stack and returns `sq_throwerror`, which TIC-80 is discarding.  TIC-80 then attempts to assign the top value of the stack into the `colors` array, which will likely be the last argument passed to the actual function rather than any valid colour.  Finally, TIC-80 will `sq_poptop` what it assumes is the result of `sq_rawget`, which means that later arguments to the function call are actually discarded.

Ideally we would just ensure we're not attempting to read too many values, which can be easily checked with `sq_getsize` on the array.

---

Lua version showing the desired functionality (slightly modified version of one of the `ttri` examples):
```lua
ttri(0,0,64,0,0,64,
     0,0,usize,0,0,vsize,
     true,
     {14,10})
ttri(64,0,0,64,64,64,
     usize,0,0,vsize,usize,vsize,
     true,
     {14,10},
     1,2,3)
```

Image

---

Squirrel version showing the broken functionality:
```squirrel
ttri(0,0,64,0,0,64,
     0,0,usize,0,0,vsize,
     true,
     [14,10])
ttri(64,0,0,64,64,64,
     usize,0,0,vsize,usize,vsize,
     true,
     [14,10],
     1,2,3)
```

Image

swoolcock:
Resolves #2987

### DIFF
--- a/src/api/squirrel.c
+++ b/src/api/squirrel.c
@@ -511,13 +511,14 @@ static SQInteger squirrel_ttri(HSQUIRRELVM vm)
         //  check for chroma
         if(OT_ARRAY == sq_gettype(vm, 15))
         {
-            for(s32 i = 0; i < TIC_PALETTE_SIZE; i++)
+            s32 arr_len = (s32)sq_getsize(vm, 15);
+            for(s32 i = 0; i < TIC_PALETTE_SIZE && i < arr_len; i++)
             {
                 sq_pushinteger(vm, (SQInteger)i);
                 sq_rawget(vm, 15);
                 if(sq_gettype(vm, -1) & (OT_FLOAT|OT_INTEGER))
                 {
-                    colors[i-1] = getSquirrelNumber(vm, -1);
+                    colors[i] = getSquirrelNumber(vm, -1);
                     count++;
                     sq_poptop(vm);
                 }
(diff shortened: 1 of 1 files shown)


nesbox (approved):
Confirmed all three, and the off-by-one is exactly the suspected copy: `luaapi.c:489` runs that loop from 1, `squirrel.c:514` ran the same body from 0, so the first colour went to `colors[-1]`, one byte before a `static` array. The `sq_getsize` bound is what makes the rest safe, since a rawget past the end throws instead of returning null and the stack desynchronizes. All four `sq_rawget` sites in the file are the four you bounded. Thanks, merging.

PR #2992, opened 2026-09-10
Add MiniScript support

MineRobber9000:
Adds support for Joe Strout's MiniScript (specifically MiniScript 2).

One notable feature of MiniScript is that infinite loops in user code won't cause TIC-80 to hang; MiniScript specifically supports running code for a limited amount of time each tick (currently set to 0.1 seconds).


nesbox:
Read through the binding and the wiring. The property this PR is bought for is real: `RunUntilDone(0.1, true)` gives a cart 0.1s per tick, so an infinite loop costs frame rate (10 FPS worst case) and never the client — nothing else in the tree can say that. The runtime arriving as a submodule of the author's repository is how the others arrive too, and the C++14/C++20 split is contained in `cmake/miniscript.cmake`.\n\nOne gap, and it is ours rather than yours: the release matrix does not know the language yet — `build.yml`'s `langs=(...)` list, which builds the per-language export stubs, still stops at the old twelve, so a MiniScript cart would have no stub to export against. You covered `assets.bat` and the baremetal Makefile. I will take the CI list as a follow-up before the next release.

nesbox (approved):
Read the binding and the wiring. The property this is bought for is real: `RunUntilDone(0.1, true)` bounds a cart to 0.1s of work per tick, so a runaway loop costs frame rate and cannot hang the client — nothing else in the tree can say that. The runtime arriving as a submodule of the language author's repository matches how the others arrive, and the C++14/C++20 split is contained in `cmake/miniscript.cmake`. The release-matrix lists are ours to grow and are deliberately waiting for the Forth PR, so this merges as the language itself.

PR #2993, opened 2026-09-11
Fix bookmark panel crashing if there is a bookmark on a blank line

swoolcock:
Self explanatory.


nesbox (approved):
Traced the crash rather than taking "self explanatory" for it: `getLineSize()` gives a blank line a size of 0, so once the trim walked past its newline, `ptr->size - (trimmed - ptr->pos)` went negative in a `size_t` — a near-`SIZE_MAX` length handed to the drawing call. Stopping at the newline keeps `trimmed` on the line and the size at 0; lines with content trim as before. Thanks, merging.

PR #2998, opened 2026-09-16
Change in the index.html file for html exports

bettywolfy:
File made by `tricoefi` on Discord


CarbonatedFennec:
I'm tricoefi, this one is also based on @matstc 's virtual gamepad HTML but I modified it to support multi-touch and the `= CLICK TO PLAY =` being able to be hidden.

CarbonatedFennec:
Also, if I can have two questions answered here, could I add a `buttons.png` file so I can work on a virtual gamepad that resembles the one in mobile? And could I add PWA-specific files similar to [this](https://github.com/CarbonatedFennec/TIC-80-PWA-Export), or does that need to be it's own target?

nesbox:
Reworked onto `main` and kept the two layout changes, because they are what the page wants: the `100dvh` height, and the portrait placement of the touch controls. The rest came out, each for a concrete reason: a hardcoded `['cart.tic']` is not the empty list the export path patches, `#add-modal` is how a player adds a cartridge from disk, the Safari numpad remap and the canvas scroll guard are platform fixes, and the `preRun` hook changed `integerScale` for the site's player and for every exported game. The commit message has the full list; the `diy-console` div and the fullscreen-button id can come back as their own change.

nesbox (approved):
Reworked onto `main`: kept the `100dvh` sizing and the portrait placement of the touch controls, and restored the empty argument list the export path patches, the add-modal the player looks up, the Safari remap and the canvas scroll guard. The `preRun` hook that wrote `integerScale` for every player and export came out, with the unused div and id.

PR #3000, opened 2026-09-17
Add raw FFT magnitude APIs fftr and fftrs

aliceisjustplaying:
`fft()` and `ffts()` apply automatic gain, which makes them unsuitable when a cart needs the captured signal's raw magnitude. This adds `fftr(start_freq, end_freq=-1)` and `fftrs(start_freq, end_freq=-1)` for raw and smoothed raw magnitudes. Both preserve the existing FFT bin and inclusive range-sum behavior. The raw values are recorded before amplification; smoothing uses the existing 0.6 history weight. See [the FFT implementation](https://github.com/aliceisjustplaying/TIC-80/blob/upstream-fftr/src/ext/fft.c).

This ports the raw FFT feature from [alice-feature-cqt](https://github.com/aliceisjustplaying/TIC-80/tree/alice-feature-cqt) onto current upstream main. It includes Lua, JavaScript, Python, Ruby, Janet, Scheme, Squirrel and Wren bindings, built-in API help and resetting the raw buffers when a cart starts. [VQT is in the separate PR #3001](https://github.com/nesbox/TIC-80/pull/3001).

Validation on macOS ARM64:

- Built the SDL application and all eight affected language runtimes. Runtime checks verified single-bin and range calls in each language.
- Ran [the offline FFT regression test](https://github.com/aliceisjustplaying/TIC-80/blob/upstream-fftr/tests/fft.c) with AddressSanitizer and UndefinedBehaviorSanitizer: amplitude scaling, smoothing, range behavior and disabled access passed.
- Compiled and ran the same test with `TIC80_FFT_UNSUPPORTED`.
- Built and tested the combined FFT/VQT changes after resolving their shared registration edits locally.

Physical capture devices and other operating systems have not been tested locally.


glastonbridge:
I am a Field-FX organiser and the people on our Monday night livecoding jams have been using a version of this (the branch mentioned above) for over a year now. It is well tested, and very useful for writing reactive effects.

nesbox:
Read it through: the raw values are recorded before the amplification multiply, the smoothed variant keeps the existing 0.6 weight, and the public `fft`/`ffts` are untouched, so the new pair is the only new behaviour. One thing worth naming: `tests/fft.c` is the first test in this tree, and it is not wired into CI yet — that is a follow-up, not a blocker.

nesbox (approved):
Read it: the raw values are recorded before the amplification multiply, the smoothed variant keeps the 0.6 weight, and the public `fft`/`ffts` are untouched. `tests/fft.c` is the first test in this tree; wiring it into CI is a follow-up.

PR #3001, opened 2026-09-17
Add VQT audio analysis and spectral whitening APIs

aliceisjustplaying:
Add Variable-Q Transform (VQT) analysis for carts that need musical-note frequency bins. The [implementation](https://github.com/aliceisjustplaying/TIC-80/blob/upstream-vqt/src/ext/vqt.c) provides 120 semitone-spaced bins using an 8192-sample capture window, enabled by the existing `--fft` option. Bin frequency is `19.445 * 2^(bin/12)` Hz, starting at D#0/Eb0. See [the bin configuration](https://github.com/aliceisjustplaying/TIC-80/blob/upstream-vqt/src/vqtdata.h).

The [eight APIs](https://github.com/aliceisjustplaying/TIC-80/blob/upstream-vqt/src/api.h) are `vqt`, `vqts`, `vqtr`, `vqtrs`, `vqtw`, `vqtsw`, `vqtrw` and `vqtrsw`. They expose normalized, raw and smoothed outputs, plus spectral whitening that reduces the broad spectral envelope to emphasize notes. Bindings cover Lua, JavaScript, Python, Ruby, Janet, Scheme, Squirrel and Wren.

This ports [alice-feature-cqt](https://github.com/aliceisjustplaying/TIC-80/tree/alice-feature-cqt) onto current upstream main. The port removes unconditional benchmarking/profiling output, completes unsupported-platform stubs, makes failed allocation cleanup safe and keeps silence at zero after whitening. [Capture integration](https://github.com/aliceisjustplaying/TIC-80/blob/upstream-vqt/src/ext/fft.c) synchronizes buffer access and keeps the existing 2048-sample FFT on the newest samples. The [raw FFT APIs are in #3000](https://github.com/nesbox/TIC-80/pull/3000); this branch does not depend on them.

Validation on macOS ARM64:

- Built the SDL application and all eight affected language runtimes. Runtime checks exercised all eight VQT functions in each language.
- Ran [the offline VQT regression test](https://github.com/aliceisjustplaying/TIC-80/blob/upstream-vqt/tests/vqt.c) with AddressSanitizer and UndefinedBehaviorSanitizer: note-bin peaks, gain scaling, smoothing, silence, finite whitening outputs, capture windows and repeated open/close passed.
- Compiled and ran the same test with `TIC80_FFT_UNSUPPORTED`.
- Injected 18 allocation failures and verified cleanup plus successful reinitialization under the sanitizers.
- Built and tested the combined FFT/VQT changes after resolving their shared registration edits locally.

Physical capture devices and other operating systems have not been tested locally.


glastonbridge:
To add the same comment to this PR too (sorry for spam): I am a Field-FX organiser and the people on our Monday night livecoding jams have been using a version of this (the branch mentioned above) for over a year now. It is well tested, and very useful for writing reactive effects.

nesbox:
Read it through. The bin base is right (D#0 is C0 plus three semitones), the FFT window keeps its place now that the buffer is `max(VQT_FFT_SIZE, FFT_SIZE*2)`, and the spinlock is taken on both sides — the capture callback and the `memcpy` that feeds the analysis — which also closes a race that was there before this PR. `tests/vqt.c` is a nice second in the style of #3000's.

nesbox (approved):
Read it: the bin base is right, the FFT window keeps its place in the enlarged buffer, and the spinlock on both sides also closes a race that predates this PR. `tests/vqt.c` is in the same style as #3000's.

nesbox:
Merged `main` in, because #3000 landed first and both PRs add an API family in the same places. Every conflict was adjacency: the macro list in `api.h`, the binding tables in janet, python and wren, and the end of `fft.c`, where the raw pair joins `FFT_CopyAudio`. Both sides kept everywhere, the macro chain closed back up; a native build of the merge links, so the two families coexist.

PR #3002, opened 2026-09-17
templates: add Go starter project template using TinyGo

akhenakh:
This is an AI assisted port of the WASM4 for Tinygo template.

I've tested some advanced features and hit some new bugs, that's why there are some more examples.

Compile Go to WebAssembly with TinyGo (target.json) and run it with TIC-80's wasm runtime.  Includes the cart/tic80 API wrapper library, the demo game and wiki-inspired examples (pix, BDR glitch, per-scanline palette gradient, ttri texture quad, sprite rotation), plus Makefile targets and run.sh for each.

- Memory: reserve TIC-80's first 96 KiB of linear memory via a stack-first layout (96 KiB + 8 KiB stack) inside the 4 wasm pages TIC-80 provides; free RAM starts at 0x18000.
- tic80.Init() must be called first in BOOT: the TIC-80 wasm runtime never calls the module's _initialize entry point, so TinyGo's runtime (heap limits, GC state) only gets set up when Init runs - without it the first heap allocation kills the game ('[trap] unreachable executed').
- The wrapper masks wasm3's byte-typed returns (pix, peek*, vbank) which only fill the low byte of the i32 return slot.
- Map remap callbacks work through the existing wasm function-table mechanism without host changes; Print/Font/Trace use a shared static NUL-terminated buffer so the leaking-GC setup stays allocation-free.
- The tilesheet is stored tile-major (tile (x/8,y/8) = 32 consecutive bytes) - the examples' setSheetPix documents the correct addressing.
- ttri2 rotates the 2x2-tile sprite from the cartridge (the TIC guy with cart.wasmp) with per-vertex z for perspective-correct texture mapping; chroma key 14.


nesbox:
Read it: `templates/` is not new, and its README asks for exactly this work, so the change is a Go entry in that list plus the `templates/go/` tree and the `.gitignore` line its Makefile needs. Nothing outside `templates/` is touched. The green run does not test it, since the template builds a cart by hand — that part rests on your testing.

nesbox (approved):
Read it: `templates/` and its README already ask for this work, and the diff adds the Go entry, the `templates/go/` tree and the `.gitignore` line its Makefile needs — nothing outside `templates/` is touched.

PR #3023, opened 2026-09-19
python: outline the code editor's defs and classes

zayfod:
The code editor's outline sidebar answers `(empty)` for every Python cart. Python was the one script table of twelve with `getOutline` left `NULL`, and `code.c:1870` skips the whole sidebar when it is, so there was never anything to list.

| | `getOutline` |
| --- | --- |
| fennel, forth, janet, js, lua, moonscript, mruby, scheme, squirrel, wasm, wren | implemented |
| python | `NULL` |

`git log -S getOutline -- src/api/python.c` returns one commit, the original pocketpy integration, so this was never written rather than lost.

## The scanner

Follows the others — find the keyword, take the name, hand back positions into the source rather than copies. `"def "` and `"class "` both introduce one: a def's name ends at its parameter list, a class's at its bases or, since those are optional, at the colon. So `class Enemy(Player):` and `class Player:` both give the class.

Two details worth calling out:

- **The keyword has to start a word.** Lua can `strstr` for `"function "` and stop thinking, but `def ` is short enough to sit at the end of an identifier — without the check, `undef = 1` and `subclass = 1` would each define something. Covered by the test.
- **The two passes emit items grouped by keyword, not in source order.** The editor sorts (`qsort` with `funcCompare`, code.c:1929), so this does not matter, but it is deliberate rather than overlooked.

Methods are listed flat, alongside their class, which matches how every other scanner here treats nested definitions.

## Test

image

`tests/python_outline.c`, in the shape of the two files already in `tests/` — standalone, `assert`-based, build line in the header comment.

The scanner is `static` inside `python.c`, which cannot be linked without pocketpy and the entire API binding, so the test goes through the `tic_script` the built module exports — ie exactly the pointer the editor calls:

```
cc -Isrc -Iinclude tests/python_outline.c -ldl -o python-outline-test
./python-outline-test build/bin/python.so
```

16 cases: the shapes carts actually contain (`def TIC`, `scanline`, plain and derived classes, a method inside a class, several defs); the identifier-tail cases (`undef`, `subclass`, `redefine_all`); extra space after the keyword; names that never terminate (`def \n`, `class\n`, `def broken\n`); and empty input.

I checked the test fails when it should — mutating one expected count makes it abort with a diagnostic rather than pass quietly.

The last case pins something that is a division of labour rather than a bug: a `def` written inside a comment **is** reported by the scanner, for Python as for every other language, because the editor drops items landing on comments before drawing them (`code.c:1877`). The test records that so a change to it has to be deliberate.

## Verified against a real cart

`demos/pythondemo.py` has exactly one definition, `def TIC():` on line 15, and the scanner returns exactly that.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nesbox (approved):

nesbox:
`main` moved under this one: #3019 landed `eval_pkpy_v2` in the same place after the keywords table, so `src/api/python.c` conflicts. Please rebase — both blocks are wanted, the two functions just land at the same anchor. The scanner itself is welcome as it is; built the branch and ran `tests/python_outline.c` against the compiled module, all tests passed.

zayfod:
Rebased onto main; kept both `eval_pkpy_v2` and `getPythonOutline`, and wired up both `.eval` and `.getOutline`. Rebuilt and re-ran `tests/python_outline.c` — all pass.

